PR #33, opened 2019-10-16
Fixed issue for OS X 10.15 Catalina (may close #25)

### LINKED ISSUE #25: Does not work when opened from .app
Switching works fine when I open it from terminal `/Applications/alt-tab-macos.app/Contents/MacOS/alt-tab-macos` but it doesn't show the switching popup when I open the app from Finder or `open /Applications/alt-tab-macos.app/`
It runs in status bar and binds to ctlr + tab shortcut because I can not longer switch between Firefox tabs but nothing happens. I have given the app Accessibility permission.
Tested on latest macOS Catalina version.

shaqed:
Problem was that for Catalina, the app needed the Screen Recording permissions to get `windowName` so added a check at the beginning.

Not much of a Swift guy, so feel free to refactor my mess.

Cool idea that I've been waiting for a long time for my Mac, so wanted to contribute :)

### DIFF
--- a/alt-tab-macos/ui/Application.swift
+++ b/alt-tab-macos/ui/Application.swift
@@ -31,7 +31,10 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate, NSCol
     }
 
     func applicationDidFinishLaunching(_ aNotification: Notification) {
+        ensureScreenRecordingCheckboxIsChecked() /* if removed, in Catalina screen shots will be replaced with program icons, and titles would be the owner app name rather than the actual window name */
         ensureAccessibilityCheckboxIsChecked()
+        
+        
         updateThumbnailMaxSize()
         makeStatusBarItem()
         makeWindow()
@@ -106,6 +109,25 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate, NSCol
             NSApp.terminate(self)
         }
     }
+    
+    func ensureScreenRecordingCheckboxIsChecked() {
+        
+        // get first window ID
+        let window = cgWindows()[0]
+        let windowOwnerPID = window[kCGWindowNumber] // window[kCGWindowOwnerPID]
+        
+        // Try getting the image
+        let windowImage = CGWindowListCreateImage(.null, .optionIncludingWindow, windowOwnerPID as! CGWindowID, [.boundsIgnoreFraming, .bestResolution])
+        
+        // If nil, permission denied (this will also prompt for permission)
+        // Should this fail, just exit
+        if windowImage == nil{
+            debugPrint("Cannot get screen shot of another window - This app does not have Screen Recording permissions")
+            NSApp.terminate(self)
+        }
+        
+        
+    }
 
     func computeOpenWindows() {
         openWindows.removeAll()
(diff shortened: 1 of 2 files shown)


lwouis:
Hi! Thanks a lot for digging into the problem, and even sending a PR!

This approach seems a bit fragile. I read more on the topic on [this stackoverflow question](https://stackoverflow.com/questions/56597221/detecting-screen-recording-settings-on-macos-catalina), and I would like to take a moment to read more and see if maybe there is a more robust solution, before merging this PR.

shaqed:
Of course, I actually saw that SO question but it didn't seem that elegant as your other Accessibility permission check and I thought I'd just write a workaround. I actually couldn't test it on other OS X versions (other than Catalina)

Before implementing the check I've added these changes locally:
Used `kCGWindowOwnerName` for Catalina users and  `kCGWindowName` for lower versions as the first doesn't require Screen Recording permissions. And also change the images from screen shots to application logos 

It's also written in the commit itself.

jfdhuiz:
Confirming this fixed #25 for me

lwouis:
I reviewed your PR in more details. Here is some feedback and questions

You explained:

> Used kCGWindowOwnerName for Catalina users and kCGWindowName for lower versions as the first doesn't require Screen Recording permissions.

However, according to the docs, `kCGWindowOwnerName` is the app name and `kCGWindowName` is the window title. That would mean that on Catalina, users would just see text like "Preview" or "Mail". This happens all the time I assume, not just if they haven't checked the new Screen Recording checkbox (I would love a screenshot of this btw!). That sounds like it's an issue. I think showing the app name should depend on the permission not being granted, not the OS version, no?

You also explained:

> And also change the images from screen shots to application logos

Do you mean we would need to write that code, or that by default Catalina replaces screenshots with icons of the app if permission wasn't granted?

OranShuster:
according to the macos docs 
```
The key that identifies the name of the window, as configured in Quartz. The value for this key is a CFStringRef type. (*Note that few applications set the Quartz window name.*)
```

seems like something you don't want to use on any osx version

lwouis:
You quoted the [docs from Apple website](https://developer.apple.com/documentation/coregraphics/kcgwindowname). Here are the docs from the actual SDK

```
/* If present, the name of the window. The value of this key is a
   CFString. */

@available(OSX 10.5, *)
public let kCGWindowName: CFString
```

Not sure which one is up-to-date. Anyway I think you worry about the name being empty most of the time is no problem at all. It's there 90% of the time from my tests. I changed recently to fallback on the app name when the window name is empty because it was happening for Line.

alt-tab-macos/ui/Application.swift
```diff
@@ -106,14 +109,41 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate, NSCol
...
+    func ensureScreenRecordingCheckboxIsChecked() {
+        
+        // get first window ID
+        let window = cgWindows()[0]
+        let windowOwnerPID = window[kCGWindowNumber] // window[kCGWindowOwnerPID]
+        
+        // Try getting the image
+        let windowImage = CGWindowListCreateImage(.null, .optionIncludingWindow, windowOwnerPID as! CGWindowID, [.boundsIgnoreFraming, .bestResolution])
+        
+        // If nil, permission denied (this will also prompt for permission)
```
lwouis: If calling `CGWindowListCreateImage` prompts for permission, then why are people saying "the app doesn't work on Catalina". They should be getting the prompt if they haven't checked the checkbox, no?

lwouis: I tested with the accessiblity checkbox which I assume to work exactly the same as the screen recording checkbox. Indeed the best approach is to stop the app after prompting. On restart by the user if will either ask again, or if the user checked the box, let it proceed.

lwouis:
I reworked this PR by removing the code about windowTitle vs appName as stopping the app after the permission checks means the situation is a non-issue. Also I changed in a recent commit the logic so that when windowTitle is empty, it falls back to appName which is always there. (the issue was with the Line app).

I pushed the change in cbfa5869d98d94f2d115fdd895c89204a0d316fb. Of course I committed under your name @shaqed to keep credits! Thanks again for the PR! I think a bunch of people on Catalina will be thankful to you!

image

PR #57, opened 2019-10-26
fix: adds missing return statements

gingerr:
so the project builds without any compile errors on macOS high sierra (10.13.x)

### DIFF
--- a/alt-tab-macos/logic/Preferences.swift
+++ b/alt-tab-macos/logic/Preferences.swift
@@ -101,11 +101,11 @@ class Preferences {
     }
 
     private static func preferencesVersion(_ url: URL) throws -> Int {
-        try Int(loadFromDisk(url)["version"] ?? "0").orThrow()
+        return try Int(loadFromDisk(url)["version"] ?? "0").orThrow()
     }
 
     private static func loadFromDisk(_ url: URL) throws -> [String: String] {
-        try JSONDecoder().decode([String: String].self, from: Data(contentsOf: url))
+        return try JSONDecoder().decode([String: String].self, from: Data(contentsOf: url))
     }
 
     private static func handleNoFileOrOldFile(_ userFile: URL) throws {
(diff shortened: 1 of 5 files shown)


gingerr:
My home computer system is a Hackintosh, which I currently can't update to Mojave or Catalina cause of missing NVIDIA drivers.

Strangely the project does not compile here but it does on my macBook which runs macOS Catalina (10.15.x). Maybe some Swift language update / configuration thing.

This commit fixes that.

lwouis:
Adding back the implicit `return`s is not a big deal so I'll merge it, but overall it would be nice to understand what went wrong better on High Sierra. Ideally I would love for this project to be portable from the dev environment perspective. I did not make use of Interface Builder to reduce dependency on (different versions of) XCode for instance.

Implicit returns are a Swift feature. If you are compiling against Swift 4.2, then it should work. I'm guessing that somehow on your High Sierra machine, the project was not configured to use the proper Swift version, or maybe some OS library? That's the kind of things that may be improved by adding some `.xcconfig` file or something. I'm not an XCode expert though, so maybe somebody else can help here

gingerr:
I see. 

It seems like to get the newest Xcode one has to use the latest macOS (apple has an min macOS Version check in the updater/installer). So my Xcode on the high Sierra hackintosh is stuck on an older version, whondering if that means that swift is also stuck on an older version. Gonna take a look today evening.

Thanks for merging the explicit returns into the codebase!

gingerr:
@lwouis Alright I've looked into it.

High Sierra is stuck on Xcode 10.1 which bundles Swift 4.2.1 which probably does not support closures. 

If I would have to I could attempt to upgrade my Hackintosh to Catalina or get with a workaround the latest Xcode running on my system so I get the latest Swift library.

lwouis:
Apple developper ecosystem is one of the worst I've experienced so far. People complain about JavaScript fatigue but this is way worse 🙉

PR #58, opened 2019-10-26
refactor: usage of apple keycodes

gingerr:
When trying to get AltTab to work on a regular pc keyboard (wrong keycodes, hopefully upcoming feature pull request) I had to get into the keycode differences and where to get them from.

I stumbled upon an stack overflow issue and found there the [apple provided "enum"](https://stackoverflow.com/questions/3202629/where-can-i-find-a-list-of-mac-virtual-key-codes/16125341).

The file is located (as mentioned in one stack overflow answer) here: `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Carbon.framework/Versions/A/Frameworks/HIToolbox.framework/Versions/A/Headers/Events.h`

This commit uses an import of that definitions (bunch of Ints) and deletes the KeyCode enum. 

As said yesterday, I don't have Swift experience so this is a implementation proposal. Ive made only the required minimal adjustments and tested on High Sierra and Catalina.

This commit bases on my earlier commit which adds the missing return statements (so the application can be compiled on High Sierra).

Cheers

### DIFF
--- a/alt-tab-macos/logic/Preferences.swift
+++ b/alt-tab-macos/logic/Preferences.swift
@@ -101,11 +101,11 @@ class Preferences {
     }
 
     private static func preferencesVersion(_ url: URL) throws -> Int {
-        try Int(loadFromDisk(url)["version"] ?? "0").orThrow()
+        return try Int(loadFromDisk(url)["version"] ?? "0").orThrow()
     }
 
     private static func loadFromDisk(_ url: URL) throws -> [String: String] {
-        try JSONDecoder().decode([String: String].self, from: Data(contentsOf: url))
+        return try JSONDecoder().decode([String: String].self, from: Data(contentsOf: url))
     }
 
     private static func handleNoFileOrOldFile(_ userFile: URL) throws {
@@ -134,7 +134,7 @@ class Preferences {
     }
 
     private static func fileFromPreferencesFolder(_ fileName: String) -> URL {
-        FileManager.default
+        return FileManager.default
                 .urls(for: .libraryDirectory, in: .userDomainMask)
                 .first!
                 .appendingPathComponent("Preferences", isDirectory: true)
--- a/alt-tab-macos/logic/Preferences.swift
+++ b/alt-tab-macos/logic/Preferences.swift
@@ -1,5 +1,6 @@
 import Foundation
 import Cocoa
+import Carbon.HIToolbox
 
 class Preferences {
     static var defaults: [String: String] = [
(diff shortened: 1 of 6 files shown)


alt-tab-macos/logic/Preferences.swift
```diff
@@ -79,7 +80,7 @@ class Preferences {
             tabKeyCode = try UInt16(value).orThrow()
         case "metaKey":
             let p = try metaKeyMacro.labelToMacro[value].orThrow()
-            metaKeyCode = p.preferences.0
+            metaKeyCode = UInt16(p.preferences.0)
```
gingerr: Ive kept metaKeyCode and tabKeyCode to be UInt16 so we don't need to cast during the often occurring listen events in Keyboard.keyboardHandler() which would be then from NSEvent.keyCode (UInt16) to HIToolbox.Events Symbols (Int).

lwouis: This is casting the values declared as `Int` from line 44 above to `UInt16`. I think just use `UInt16` everywhere and you will not need to cast here, no?

gingerr: Yeah, somewhere the cast has to happen, either at the top in this file where the Mapping occurs or at this code line. The apple provided symbols are old and are just declared as Int.

lwouis:
This is nice! I had seen mentions of this `HIToolbox` library previously, but I thought it would add weight to the project. For instance, this is the current breakdown on this app:

image

Look how the actual app is 4% of the size, and Swift bindings are 96% of the size.

Anyway I think here it should be fine. I'm gonna test locally to see if there is any change.

But anyway like I mentioned in another thread, I think all this code is gonna be replaced with a hotkey library anyway to make it more robust. I expect then that the size of the app will increase quite a bit unfortunately, but I think it's worth it if it means nice UI and reliable functionality.

gingerr:
Thanks at pointing at this aspect, did not come into my mind to check the bundle for size increase. 

Did my import cause all of this? I was expecting it would just import the little file i used.

gingerr:
Maybe the import can be changed to carbon.hitoolbox.events? I was expecting it would take only what it needs and do some tree shaking of sorts as in angular cli.

lwouis:
> Did my import cause all of this? I was expecting it would just import the little file i used.

No, I meant to show that I learned about this `hitoolbox` thing but because I had previously noticed the impact of Swift on the bundle size, I decided to code a small enum myself in case this pull a lot of dependency somehow. I haven't tested your change yet locally

lwouis:
I pushed these commits (479778662de085cbf4a293c3479ca7299923eb9b, 7e21974ebfa0acbd86eaf9e6b9d6680c49c88a54), so no need for the PR anymore. Btw I found a regression on master (4976267a821c61fbbc43948898e4ed7f2b0f4a60) testing this PR, so thanks for that! It was a bad regression too :o

PR #60, opened 2019-10-27
feat: multiple meta keys

gingerr:
Hi Louis!

Ive made an little change and am interested in your opinion and if you accept the pull request or improve upon.

`metaKeyCodes` is now an array and is provided with both keycodes for `option`, `control` and `command`. 

This fixes two issues:
- UI stayed open when the user did open it with the secondary `metaKey` keyCode (the UI got toggled by the `modifierFlags` check) and the user had to click the proper `metaKeyCode` to close it again
- PC keyboards (all of them? i observed this on a Logitech K740) send the secondary keyCode for the command and option keys so the application did not work for this setup

```
MacBook Pro 2015 modifier keycodes
left / right
 
Command: 55 / 54
Option: 58 / 61
Control: 59
Function: 63
```

```
PC Keyboard (Logitech K740) on macOS
left / right

Command: 54 / 55
Option: 61
Control: 59 / 62
Function: no keycode event
```

Ive tested the application on macOS High Sierra (mentioned pc keyboard) and macOS Catalina (MacBook Pro keyboard).

The pc keyboard I use has an issue with the option metakey event handling but this not related to this commit.

Cheers

### DIFF
--- a/alt-tab-macos/logic/Preferences.swift
+++ b/alt-tab-macos/logic/Preferences.swift
@@ -32,7 +32,7 @@ class Preferences {
     static var tabKeyCode: UInt16?
     static var highlightBorderColor: NSColor?
     static var highlightBackgroundColor: NSColor?
-    static var metaKeyCode: UInt16?
+    static var metaKeyCodes: [UInt16] = []
     static var metaModifierFlag: NSEvent.ModifierFlags?
     static var windowDisplayDelay: DispatchTimeInterval?
     static var windowCornerRadius: CGFloat?
@@ -41,12 +41,12 @@ class Preferences {
         MacroPreference(" macOS", (0, 5, 20, .clear, NSColor(red: 0, green: 0, blue: 0, alpha: 0.3))),
         MacroPreference("❖ Windows 10", (2, 0, 0, .white, .clear))
     ])
-    static var metaKeyMacro = MacroPreferenceHelper<(Int, NSEvent.ModifierFlags)>([
-        MacroPreference("⌥ option", (kVK_Option, .option)),
-        MacroPreference("⌃ control", (kVK_Control, .control)),
-        MacroPreference("⌘ command", (kVK_Command, .command)),
-        MacroPreference("⇪ caps lock", (kVK_CapsLock, .capsLock)),
-        MacroPreference("fn", (kVK_Function, .function))
+    static var metaKeyMacro = MacroPreferenceHelper<([Int], NSEvent.ModifierFlags)>([
+        MacroPreference("⌥ option", ([kVK_Option, kVK_RightOption], .option)),
+        MacroPreference("⌃ control", ([kVK_Control, kVK_RightControl], .control)),
+        MacroPreference("⌘ command", ([kVK_Command, kVK_RightCommand], .command)),
+        MacroPreference("⇪ caps lock", ([kVK_CapsLock], .capsLock)),
+        MacroPreference("fn", ([kVK_Function], .function))
     ])
 
     private static let defaultsFile = fileFromPreferencesFolder("alt-tab-macos-defaults.json")
@@ -80,7 +80,7 @@ class Preferences {
             tabKeyCode = try UInt16(value).orThrow()
         case "metaKey":
             let p = try metaKeyMacro.labelToMacro[value].orThrow()
-            metaKeyCode = UInt16(p.preferences.0)
+            metaKeyCodes = p.preferences.0.map { UInt16($0) }
             metaModifierFlag = p.preferences.1
         case "theme":
             let p = try themeMacro.labelToMacro[value].orThrow()
--- a/alt-tab-macos/logic/Preferences.swift
+++ b/alt-tab-macos/logic/Preferences.swift
@@ -101,11 +101,11 @@ class Preferences {
     }
 
     private static func preferencesVersion(_ url: URL) throws -> Int {
-        try Int(loadFromDisk(url)["version"] ?? "0").orThrow()
+        return try Int(loadFromDisk(url)["version"] ?? "0").orThrow()
     }
 
     private static func loadFromDisk(_ url: URL) throws -> [String: String] {
-        try JSONDecoder().decode([String: String].self, from: Data(contentsOf: url))
+        return try JSONDecoder().decode([String: String].self, from: Data(contentsOf: url))
     }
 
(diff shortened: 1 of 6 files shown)


gingerr:
We could also let the User to choose between e.g. command left/right explicitly instead of using an array (for now, until the library mentioned is evaluated/integrated).

The native macOS alt tab Switcher also reacts to both key variants of I remember correctly.

lwouis:
I tried on my macbook keyboard and see this behaviour:

* pressing left_control+tab triggers the built-in app-switcher
* there is no right_control so can't test that
* pressing left_command+tab triggers AltTab
* pressing right_command+tab triggers the built-in app-switcher (strange)

When I made the current implementation, I noticed left/right have different codes. At that time I thought "alt-tab is a shortcut you are always going to press with your left hand alone. It's a one-hand shortcut, so right side of the keyboard shouldn't matter". This is why the code ignores the other codes.

I think while this PR attempts to make it more flexible for the user, it may break shortcuts for some users. What would you say we try to integrate [ShortcutRecorder](https://github.com/Kentzo/ShortcutRecorder) and not waste our energy on fiddling with these low-level concerns? ;p

lwouis:
I'm testing this locally right now, and every shortcut seems to work with this PR

alt-tab-macos/logic/Preferences.swift
```diff
@@ -79,7 +80,7 @@ class Preferences {
             tabKeyCode = try UInt16(value).orThrow()
         case "metaKey":
             let p = try metaKeyMacro.labelToMacro[value].orThrow()
-            metaKeyCode = p.preferences.0
+            metaKeyCodes = p.preferences.0.map { UInt16($0) }
```
lwouis: Just FYI I was afraid that this `map` would not guarantee to respect the input array order. It [seems it does](https://stackoverflow.com/a/31358962/2249756) :)

alt-tab-macos/logic/Preferences.swift
```diff
@@ -31,7 +32,7 @@ class Preferences {
     static var tabKeyCode: UInt16?
     static var highlightBorderColor: NSColor?
     static var highlightBackgroundColor: NSColor?
-    static var metaKeyCode: KeyCode?
+    static var metaKeyCodes: [UInt16] = []
```
lwouis: Could you replace this with `[UInt16]?`. I would like to avoid ambiguity of where things are initialized. I used to have all members of this class have some dummy value like this, but then it when changing the default values I would get confused on where the default values are.

Here we have a class that has no useful values until the JSON file is read or created on disk, so I think optional values represent that correctly. We know that after this file is read the optional are filled (otherwise the app is crashed and stopped), so we can use `!` everywhere in the app to query these preferences.

This is not ideal, but for some reason, Swift compiler doesn't notice that reading the JSON file happens in the initializer and that it has to happen or it throws. If the compiler was smarter, we could avoid the `?` since we know the object has to be initialized with actual values. But like I said between satisfying the compiler with random init values like `[]` or `.white` which could confuse the reader as default, and using `?` and later `!`, I would prefer the later approach

lwouis:
I'm going to rebase and push this PR myself. Of course the 2 commits will keep your name as the author. My goal is just to make the PR merge faster since I've played with it locally and I have it already rebased

lwouis:
I pushed these commits (479778662de085cbf4a293c3479ca7299923eb9b, 7e21974ebfa0acbd86eaf9e6b9d6680c49c88a54), so no need for the PR anymore. Btw I found a regression on master (4976267a821c61fbbc43948898e4ed7f2b0f4a60) testing this PR, so thanks for that! It was a bad regression too :o

alt-tab-macos/logic/Preferences.swift
gingerr: In this case its not critical, it e.g. contains `[kVK_Control, kVK_RightControl]` and the usage is `let isMeta = Preferences.metaKeyCodes!.contains(event.keyCode)`

Thanks for pointing at this aspect. Good to know!

alt-tab-macos/logic/Preferences.swift
gingerr: Roger that. I was thinking about if it should be initialised or not but am not familiar enough with the Optional situation in Swift and chose the easy way out since there are other variables already initialised. Your point of view on that clears that up, thanks!

gingerr:
Great, thanks for merging this into your codebase.

gingerr:
> I tried on my macbook keyboard and see this behaviour:
> 
> * pressing `left_command`+`tab` triggers the built-in app-switcher
> * pressing `right_command` + `tab` triggers the built-in app-switcher
>   AltTab set to `option`:
> * pressing `left_option`+`tab` triggers AltTab
> * pressing `right_option`+`tab` triggers AltTab **but doesn't close it correctly**
>   AltTab set to `control`:
> * pressing `left_control`+`tab` triggers AltTab
> * there is no `right_control` so can't test that
> 
> I also own a Logitech keyboard (K380) and saw this:
> 
> * pressing `left_command`+`tab` triggers the built-in app-switcher
> * pressing `right_command` + `tab` triggers the built-in app-switcher
>   AltTab set to `option`:
> * pressing `left_option`+`tab` triggers AltTab
> * pressing `right_option`+`tab` doesn't trigger anything
>   AltTab set to `control`:
> * pressing `left_control`+`tab` triggers AltTab
> * there is no `right_control` so can't test that
> 
> When I made the current implementation, I noticed left/right have different codes. At that time I thought "alt-tab is a shortcut you are always going to press with your left hand alone. It's a one-hand shortcut, so right side of the keyboard shouldn't matter". This is why the code ignores the other codes.
> 
> I think while this PR attempts to make it more flexible for the user, it may break shortcuts for some users. What would you say we try to integrate [ShortcutRecorder](https://github.com/Kentzo/ShortcutRecorder) and not waste our energy on fiddling with these low-level concerns? ;p

Yeah usually this should be a one hand job and the application would need only to support the left keycodes but my pc keyboard sends for the physically left placed command (windows alt) and option (windows win key) buttons the **right** side keycodes 😕  

Wondering if thats my keyboard doing that or if its normal for windows keyboards. How about your K380? Did it send the left option keycode when pressing the left option button?

I stumbled upon this the `right_option` key issue (not working at all) and fixed it also yesterday (upcoming pull request which explains the issue and my solution).

PR #63, opened 2019-10-28
fix: kVK_RightOption issue

gingerr:
by implementing the trigger without usage of modifierFlags.

We observed both, that triggering the application with kVK_RightOption is not possible (native alt tab comes up instead on my system) on some? keyboards. Ive looked into it yesterday when trying out all possible combinations. Ive looked into it since kVK_RightOption is actually located physically left on my pc keyboard (Logitech K740).

This is what happens if we trigger the main feature with kVK_RightCommand (54) + kVK_Tab (48) on a pc keyboard (Logitech K740):

```
"pressed" 54
"isMeta"
"metaModifierFlag"
"-------"
"pressed" 48
"metaModifierFlag"
"keyDown"
"isTab"
"showUiOrSelectNext"
"showUiOrCycleSelection"
"pressed" 48
"metaModifierFlag"
"-------"
"pressed" 54
"isMeta"
"else isMeta and not keyDown"
"focusTarget"
"hideUi"
"-------"
```

This is what happens if we trigger the main feature with kVK_RightOption (61) + kVK_Tab (48) the same keyboard: 

```
"pressed" 61
"isMeta"
"metaModifierFlag"
"-------"
"pressed" 48
"no metaModifierFlag" <<<<<<<<<<<<<<
"-------"
"pressed" 48
"no metaModifierFlag" <<<<<<<<<<<<<<
"-------"
"pressed" 55
"metaModifierFlag"
"keyDown"
"-------"
"pressed" 55
"metaModifierFlag"
"-------"
"pressed" 61
"isMeta"
"else isMeta and not keyDown"
"focusTarget"
"-------"
"pressed" 0
"no metaModifierFlag"
"-------"
```

Tab (48) gets pressed without the proper NSEvent.ModifierFlags `.option` being present. The Tab event will contain `.command` flag instead 💩 

What happens is fascinating, I wonder if macOS adds all the additional keycodes or if it the hardware (keyboard) itself does that (command key 55 and keycode 0 get added). The 0 keycode gets also added on a macBook keyboard if I switch the cmd and option keys via the macOS native "Modifier Keys..." menu. But it does not add the command keycode when pressing option. 

Iam using the "Modifier Keys" menu for my PC keyboard so that the physical windows key is option and ALT is command to mimic the macOS keyboard layout.

Ive tried to come up with different solutions and ended up with not honouring the NSEvent.ModifierFlags at all.

Before commit: `if event.modifierFlags.contains(Preferences.metaModifierFlag!) {`
After commit: `if delegate.isMetaKeyPressed {`

The application watches and stores (Bool toggle), if the metaKey (whatever the user has choosen) is being pressed. The NSEvent.ModifierFlags are removed from the Preferences.

Do you want to support this combination? Do you see any issues with that implementation? What do you think?

I know, this ignores the given flags (but cgEvent.type == .flagsChanged is still required) to solve one currently known problematic meta key combination so I fully understand if you reject this change.

Tested on MacBook Pro 2015 keyboard and PC keyboard (Logitech K740).

### DIFF
--- a/alt-tab-macos/logic/Keyboard.swift
+++ b/alt-tab-macos/logic/Keyboard.swift
@@ -38,7 +38,10 @@ func keyboardHandler(_ cgEvent: CGEvent, _ delegate: Application) -> Unmanaged<C
             let isRightArrow = event.keyCode == kVK_RightArrow
             let isLeftArrow = event.keyCode == kVK_LeftArrow
             let isEscape = event.keyCode == kVK_Escape
-            if event.modifierFlags.contains(Preferences.metaModifierFlag!) {
+            if isMeta {
+                delegate.toggleMetaKey()
+            }
+            if delegate.isMetaKeyPressed {
                 if keyDown {
                     if isTab && event.modifierFlags.contains(.shift) {
                         delegate.showUiOrSelectPrevious()
--- a/alt-tab-macos/logic/Preferences.swift
+++ b/alt-tab-macos/logic/Preferences.swift
@@ -33,7 +33,6 @@ class Preferences {
     static var highlightBorderColor: NSColor?
     static var highlightBackgroundColor: NSColor?
     static var metaKeyCodes: [UInt16]?
-    static var metaModifierFlag: NSEvent.ModifierFlags?
     static var windowDisplayDelay: DispatchTimeInterval?
     static var windowCornerRadius: CGFloat?
     static var font: NSFont?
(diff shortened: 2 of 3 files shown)


gingerr:
@lwouis mention cause not sure if you get notifications from pull request.

gingerr:
@lwouis never mind, the additional keycodes are interferences that has been caused by another tool iam using (BetterTouchTool) ....... 😞 

Iam using it to get simulated trackpad swipes via mouse4/mouse5 buttons and window snapping & moving.

What a bummer (time investment) but was still a good exercise. Glad I tested it again with all other third party tools turned off.

lwouis:
I had the  same thing happen to me yesterday. Spent a while testing and seeing weird behavior before realizing i had Karabiner Elements still open and messing with my keys :p

Just to reiterate, hardware diversity is a very difficult and tedious thing to support which is why i want to switch to an established hotkey library. Also the old Carbon APIs that these guys use have way more control that the one i'm using. You can chose stuff like "can another app override my shortcut or not? Should i take the shortcut if it is already taken? Etc"

Issue #13, opened 2019-08-29
Sign .app to avoid the "app from an unidentified developer" prompt


lwouis:
The `.app` is unsigned currently. When the user opens the app for the first time, they get a scary prompt, and must jump through hoops to allow the app to run:

![image](https://user-images.githubusercontent.com/106195/63993812-29659200-cb2d-11e9-8458-598f74079b5a.png)

Getting an apple certificate and signing the app would remove that friction.

Here is a tutorial: https://successfulsoftware.net/2012/08/30/how-to-sign-your-mac-os-x-app-for-gatekeeper/

**An issue** is that this means buying a 100$/year apple dev account. Beyond the money, it makes this less of an open-source project as i would need my personal paid account to sign. Forks would need their own accounts to sign. I don't think there is any way around this though. This is an Apple ecosystem limitation.

lwouis:
I researched code-signing in the OSS community. There is no way around paying Apple [礼金](https://en.wikipedia.org/wiki/Key_money). I bit the bullet and just bought the certificate for ￥12,980. I don't want people to have a bad experience when launching the app for the first time. The certificate should be available for 5 years from now if I don't renew the developer account.

Adding this into CI was a bit tricky as Apple makes us jump through endless hoops with [notarization](https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution). I hope it works on v3 release. As usual with CI, it's pretty hard to test changes. I tested as much as possible locally, but only the real travis build will tell if the new code is correct. Same with my work on #75.

PR #65, opened 2019-10-29
docs: improve readme formatting

Calinou:
This makes the README follow a more conventional structure, while also adding Homebrew installation instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,32 +1,37 @@
-# [Download the latest release](https://github.com/lwouis/alt-tab-macos/releases/latest)
-
-# Purpose
+# alt-tab-macos
 
 [![Screenshot](docs/img/alt-tab-macos/frontpage.jpg)](docs/img/alt-tab-macos/frontpage.jpg)
 
-**alt-tab-macos** brings the power of Windows "alt-tab" window switcher to macOS.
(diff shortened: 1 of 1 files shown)


lwouis:
Thanks for contributing! I didn't know someone put this app on homebrew. If the readme is going to mention homebrew as an install method, then I want to make sure it is properly maintained. I just checked and saw for instance:

> Requires: macOS >= 10.14

Which is incorrect. The app runs on 10.12-10.15.

Also from looking at [this file history on the official Homebrew cask repo](https://github.com/Homebrew/homebrew-cask/commits/master/Casks/alt-tab.rb), it seems the versions are currently manually bumped which is something I want to avoid. I will check if there is a way to link releases to github releases to automate that process.

When these 2 issues are fixed with Homebrew, I will merge this PR

Calinou:
@lwouis Done :slightly_smiling_face:

lwouis:
> @lwouis Done 🙂 I also added instructions to bypass Gatekeeper (I had to do this here, as it seems the app isn't signed).

Oh very nice, thank you! There is a ticket to add signing #13 but someone needs to be paying for an Apple Developer account and kind of "own" the app to enable signing, which goes a bit against the OSS spirit. That being said I may just do it for users convenience. It's low on my priority list right now though tbh

Issue #28, opened 2019-10-15
show window on display with the mouse, not where an active window is


Somberland:
I'm working with an 3-display setup, sometimes it happens that I select the wrong window. when I open the switcher again the overlay is on the monitor with the opened app, even if my mouse is on an other display.

see also #8

lwouis:
To illustrate your reference to #8, this is what the preference looks like in HyperSwitch:

![image](https://user-images.githubusercontent.com/106195/67002112-268f2280-f116-11e9-8baf-34df5d43cd05.png)

lwouis:
Please check out [release 1.12.0](https://github.com/lwouis/alt-tab-macos/releases) which implements this feature!

Issue #59, opened 2019-10-27
Input field for window apparition delay not wide enough to handle more than 2 digits without wrapping.


adrianbj:
![image](https://user-images.githubusercontent.com/871211/67629548-94350000-f834-11e9-850d-a46196222174.png)

lwouis:
I'm aware that the preferences are pretty rudimentary for now. I added this to provide an easier option than editing the JSON file directly.

I'll close this ticket as plans to improve and overall roadmap for better UI are detailed in #49.

PR #66, opened 2019-10-30
feat: windows on mouse screen, implements #28

gingerr:
@lwouis Ive implemented the Preference and required logic that allows to choose between NSScreen.main (keyboard focus) and the NSScreen which houses the mouse pointer.

This is issue #28 and the commit has also an minimal adjustment for #59 (since I implicitly made the PreferencesPanel be even wider by us having now a new DropDown with a longer text compared to the others).

Curious if you find something that should be improved in this pull request or if its fine as it is and of course if you accept this feature addition 😎 

Btw. I found out how to **not** get prompted with the security & privacy allowance dialogs after each build. I noticed the builds done by Xcode didn't ask for enabling the permissions on each new build and it got me thinking. Ive added AppCode then to both permissions and I get no longer asked to allow them for new builds of AltTab done with AppCode 🚀 

Commit message:
- adds a new Preference to use the main screen (keyboard focus) or mouse location screen
- honors that preference for thumbnail generation and display of all panels
- increases the Preferences version
- widens PreferencesPanel width and input sizing to allow 3 characters, issue #59 minimal adjustment

### DIFF
--- a/alt-tab-macos/logic/Preferences.swift
+++ b/alt-tab-macos/logic/Preferences.swift
@@ -2,9 +2,14 @@ import Foundation
 import Cocoa
 import Carbon.HIToolbox.Events
 
+enum ShowOnScreenPreference {
+    case MAIN
+    case MOUSE
+}
+
 class Preferences {
     static var defaults: [String: String] = [
-        "version": "2", // bump this anytime the dictionary is changed
+        "version": "3", // bump this anytime the dictionary is changed
         "maxScreenUsage": "80",
         "maxThumbnailsPerRow": "4",
         "iconSize": "32",
@@ -12,11 +17,10 @@ class Preferences {
         "tabKeyCode": String(kVK_Tab),
         "metaKey": metaKeyMacro.macros[0].label,
         "windowDisplayDelay": "0",
-        "theme": themeMacro.macros[0].label
+        "theme": themeMacro.macros[0].label,
+        "showOnScreen": showOnScreenMacro.macros[0].label
     ]
     static var rawValues = [String: String]()
-    static var thumbnailMaxWidth: CGFloat = 200
-    static var thumbnailMaxHeight: CGFloat = 200
     static var minimumWindowSize: CGFloat = 200
     static var fontColor: NSColor = .white
     static var windowMaterial: NSVisualEffectView.Material = .dark
@@ -37,6 +41,7 @@ class Preferences {
     static var windowDisplayDelay: DispatchTimeInterval?
     static var windowCornerRadius: CGFloat?
     static var font: NSFont?
+    static var showOnScreen: ShowOnScreenPreference?
     static var themeMacro = MacroPreferenceHelper<(CGFloat, CGFloat, CGFloat, NSColor, NSColor)>([
         MacroPreference(" macOS", (0, 5, 20, .clear, NSColor(red: 0, green: 0, blue: 0, alpha: 0.3))),
         MacroPreference("❖ Windows 10", (2, 0, 0, .white, .clear))
@@ -45,6 +50,10 @@ class Preferences {
         MacroPreference("⌥ option", ([kVK_Option, kVK_RightOption], .option)),
         MacroPreference("⌃ control", ([kVK_Control, kVK_RightControl], .control)),
         MacroPreference("⌘ command", ([kVK_Command, kVK_RightCommand], .command))
+    ])
+    static var showOnScreenMacro = MacroPreferenceHelper<ShowOnScreenPreference>([
+        MacroPreference("with keyboard focus", ShowOnScreenPreference.MAIN),
+        MacroPreference("with mouse pointer", ShowOnScreenPreference.MOUSE),
     ])
 
     private static let defaultsFile = fileFromPreferencesFolder("alt-tab-macos-defaults.json")
@@ -89,6 +98,9 @@ class Preferences {
             highlightBackgroundColor = p.preferences.4
         case "windowDisplayDelay":
             windowDisplayDelay = DispatchTimeInterval.milliseconds(try Int(value).orThrow())
+        case "showOnScreen":
+            let p = try showOnScreenMacro.labelToMacro[value].orThrow()
+            showOnScreen = p.preferences
         default:
             throw "Tried to update an unknown preference: '\(valueName)' = '\(value)'"
         }
--- a/alt-tab-macos/logic/Preferences.swift
+++ b/alt-tab-macos/logic/Preferences.swift
@@ -52,8 +52,8 @@ class Preferences {
         MacroPreference("⌘ command", ([kVK_Command, kVK_RightCommand], .command))
     ])
     static var showOnScreenMacro = MacroPreferenceHelper<ShowOnScreenPreference>([
-        MacroPreference("with keyboard focus", ShowOnScreenPreference.MAIN),
-        MacroPreference("with mouse pointer", ShowOnScreenPreference.MOUSE),
+        MacroPreference("Main screen", ShowOnScreenPreference.MAIN),
+        MacroPreference("Screen including mouse", ShowOnScreenPreference.MOUSE),
     ])
 
     private static let defaultsFile = fileFromPreferencesFolder("alt-tab-macos-defaults.json")
--- a/alt-tab-macos/logic/Screen.swift
+++ b/alt-tab-macos/logic/Screen.swift
@@ -2,55 +2,44 @@ import Foundation
 import Cocoa
 
 class Screen {
-    // currently not in use (but kept for reference & future use), earlier registered by Application.applicationDidFinishLaunching()
-    static func listenToChanges() {
-        NotificationCenter.default.addObserver(
-                forName: NSApplication.didChangeScreenParametersNotification,
-                object: NSApplication.shared,
-                queue: OperationQueue.main
-        ) { notification -> Void in
-            // do something
-        }
-    }
-
-    static func calcThumbnailMaxSize(_ screen: NSScreen) -> NSSize {
-        let width: CGFloat = (screen.frame.size.width * Preferences.maxScreenUsage! - Preferences.windowPadding * 2) / Preferences.maxThumbnailsPerRow! - Preferences.interItemPadding
-        let height: CGFloat = width * (screen.frame.height / screen.frame.width)
-        return NSSize(width: width, height: height)
-    }
-
-    static func calcFrameMaxSize(_ screen: NSScreen) -> NSSize {
-        return NSSize(width: screen.frame.width * Preferences.maxScreenUsage!, height: screen.frame.height * Preferences.maxScreenUsage!)
-    }
-
-    // TODO: currently unknown and unhandled error use-case possible: NSScreen.main being nil
-    static func getPreferredScreen() -> NSScreen {
+    static func preferredScreen() -> NSScreen {
         switch Preferences.showOnScreen! {
         case .MOUSE:
-            return getScreenWithMouse() ?? NSScreen.main!; // .main as fall-back
+            return screenWithMouse() ?? NSScreen.main!; // .main as fall-back
         case .MAIN:
             return NSScreen.main!;
         }
     }
 
-    static func getScreenWithMouse() -> NSScreen? {
+    private static func screenWithMouse() -> NSScreen? {
         return NSScreen.screens.first { NSMouseInRect(NSEvent.mouseLocation, $0.frame, false) }
     }
 
-    /*
-    usage notes:
-    - useAppleVerticalOffset: applies a vertical offset (higher positioning) attempting to approximate the NSView.center() results but only if we are not dealing with screen filling frames (height)
-    */
-    static func showCenteredFrontPanel(_ panel: NSPanel, _ screen: NSScreen, _ useAppleVerticalOffset: Bool = true) {
-        var verticalOffset: CGFloat = 0
+    static func thumbnailMaxSize(_ screen: NSScreen) -> NSSize {
+        let frame = screen.visibleFrame
+        let width = (frame.width * Preferences.maxScreenUsage! - Preferences.windowPadding * 2) / Preferences.maxThumbnailsPerRow! - Preferences.interItemPadding
+        let height = width * (frame.height / frame.width)
+        return NSSize(width: width, height: height)
+    }
 
-        if useAppleVerticalOffset && panel.frame.height < screen.visibleFrame.height / 1.5 {
-            verticalOffset = CGFloat(screen.visibleFrame.height * 0.175)
-        }
+    static func thumbnailPanelMaxSize(_ screen: NSScreen) -> NSSize {
+        let frame = screen.visibleFrame
+        return NSSize(width: frame.width * Preferences.maxScreenUsage!, height: frame.height * Preferences.maxScreenUsage!)
+    }
 
-        let centerPosition = NSPoint(x: screen.visibleFrame.midX - panel.frame.width / 2, y: screen.visibleFrame.midY - panel.frame.height / 2 + verticalOffset)
-        panel.setFrameOrigin(centerPosition)
+    static func showPanel(_ panel: NSPanel, _ screen: NSScreen, _ alignment: VerticalAlignment) {
+        let screenFrame = screen.visibleFrame
+        let panelFrame = panel.frame
+        let x = screenFrame.minX + max(screenFrame.width - panelFrame.width, 0) * 0.5
+        let y = screenFrame.minY + max(screenFrame.height - panelFrame.height, 0) * alignment.rawValue
+        panel.setFrameOrigin(NSPoint(x: x, y: y))
         panel.makeKeyAndOrderFront(nil)
         Application.shared.arrangeInFront(nil)
     }
+}
+
+enum VerticalAlignment: CGFloat {
+    case centered = 0.5
+    // vertically centered but with an upward offset, similar to a book title; mimics NSView.center()
+    case appleCentered = 0.75
 }
--- a/alt-tab-macos/logic/Screen.swift
+++ b/alt-tab-macos/logic/Screen.swift
@@ -2,21 +2,55 @@ import Foundation
 import Cocoa
 
 class Screen {
+    // currently not in use (but kept for reference & future use), earlier registered by Application.applicationDidFinishLaunching()
     static func listenToChanges() {
         NotificationCenter.default.addObserver(
                 forName: NSApplication.didChangeScreenParametersNotification,
                 object: NSApplication.shared,
                 queue: OperationQueue.main
         ) { notification -> Void in
-            updateThumbnailMaxSize()
+            // do something
         }
     }
 
-    static func updateThumbnailMaxSize() -> Void {
-        if Preferences.thumbnailMaxWidth == 200 && Preferences.thumbnailMaxHeight == 200 {
-            let main = NSScreen.main!.frame
-            Preferences.thumbnailMaxWidth = (NSScreen.main!.frame.size.width * Preferences.maxScreenUsage! - Preferences.windowPadding * 2) / Preferences.maxThumbnailsPerRow! - Preferences.interItemPadding
-            Preferences.thumbnailMaxHeight = Preferences.thumbnailMaxWidth * (main.height / main.width)
+    static func calcThumbnailMaxSize(_ screen: NSScreen) -> NSSize {
+        let width: CGFloat = (screen.frame.size.width * Preferences.maxScreenUsage! - Preferences.windowPadding * 2) / Preferences.maxThumbnailsPerRow! - Preferences.interItemPadding
+        let height: CGFloat = width * (screen.frame.height / screen.frame.width)
+        return NSSize(width: width, height: height)
+    }
+
+    static func calcFrameMaxSize(_ screen: NSScreen) -> NSSize {
+        return NSSize(width: screen.frame.width * Preferences.maxScreenUsage!, height: screen.frame.height * Preferences.maxScreenUsage!)
+    }
+
+    // TODO: currently unknown and unhandled error use-case possible: NSScreen.main being nil
+    static func getPreferredScreen() -> NSScreen {
+        switch Preferences.showOnScreen! {
+        case .MOUSE:
+            return getScreenWithMouse() ?? NSScreen.main!; // .main as fall-back
+        case .MAIN:
+            return NSScreen.main!;
         }
+    }
+
+    static func getScreenWithMouse() -> NSScreen? {
+        return NSScreen.screens.first { NSMouseInRect(NSEvent.mouseLocation, $0.frame, false) }
+    }
+
+    /*
+    usage notes:
+    - useAppleVerticalOffset: applies a vertical offset (higher positioning) attempting to approximate the NSView.center() results but only if we are not dealing with screen filling frames (height)
+    */
+    static func showCenteredFrontPanel(_ panel: NSPanel, _ screen: NSScreen, _ useAppleVerticalOffset: Bool = true) {
+        var verticalOffset: CGFloat = 0
+
+        if useAppleVerticalOffset && panel.frame.height < screen.visibleFrame.height / 1.5 {
+            verticalOffset = CGFloat(screen.visibleFrame.height * 0.175)
+        }
+
+        let centerPosition = NSPoint(x: screen.visibleFrame.midX - panel.frame.width / 2, y: screen.visibleFrame.midY - panel.frame.height / 2 + verticalOffset)
+        panel.setFrameOrigin(centerPosition)
+        panel.makeKeyAndOrderFront(nil)
+        Application.shared.arrangeInFront(nil)
     }
 }
--- a/alt-tab-macos/ui/Application.swift
+++ b/alt-tab-macos/ui/Application.swift
@@ -28,7 +28,7 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
         SystemPermissions.ensureAccessibilityCheckboxIsChecked()
         Preferences.loadFromDiskAndUpdateValues()
         statusItem = StatusItem.make(self)
-        thumbnailsPanel = ThumbnailsPanel(self, Screen.getPreferredScreen())
+        thumbnailsPanel = ThumbnailsPanel(self)
         preferencesPanel = PreferencesPanel()
         Keyboard.listenToGlobalEvents(self)
     }
@@ -58,8 +58,8 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
         }
     }
 
-    @objc func showCenteredPreferencesPanel() {
-        Screen.showCenteredFrontPanel(preferencesPanel!, Screen.getPreferredScreen())
+    @objc func showPreferencesPanel() {
+        Screen.showPanel(preferencesPanel!, Screen.preferredScreen(), .appleCentered)
     }
 
     func computeOpenWindows() {
@@ -101,12 +101,12 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
                 return
             }
             selectedOpenWindow = cellWithStep(step)
-            let currentScreen = Screen.getPreferredScreen() // we want all computations and renderings to use the same screen for this summon (in case mouse movements switching screens etc.)
             var workItem: DispatchWorkItem!
             workItem = DispatchWorkItem {
+                let currentScreen = Screen.preferredScreen() // fix screen between steps since it could change (e.g. mouse moved to another screen)
                 if !workItem.isCancelled { self.thumbnailsPanel!.computeThumbnails(currentScreen) }
                 if !workItem.isCancelled { self.thumbnailsPanel!.highlightCellAt(step) }
-                if !workItem.isCancelled { Screen.showCenteredFrontPanel(self.thumbnailsPanel!, currentScreen, true) }
+                if !workItem.isCancelled { Screen.showPanel(self.thumbnailsPanel!, currentScreen, .appleCentered) }
             }
             workItems.append(workItem)
             DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + Preferences.windowDisplayDelay!, execute: workItem)
--- a/alt-tab-macos/ui/Application.swift
+++ b/alt-tab-macos/ui/Application.swift
@@ -27,12 +27,10 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
         SystemPermissions.ensureScreenRecordingCheckboxIsChecked()
         SystemPermissions.ensureAccessibilityCheckboxIsChecked()
         Preferences.loadFromDiskAndUpdateValues()
-        Screen.updateThumbnailMaxSize()
         statusItem = StatusItem.make(self)
-        thumbnailsPanel = ThumbnailsPanel(self)
+        thumbnailsPanel = ThumbnailsPanel(self, Screen.getPreferredScreen())
         preferencesPanel = PreferencesPanel()
         Keyboard.listenToGlobalEvents(self)
-        Screen.listenToChanges()
     }
 
     func showUiOrSelectNext() {
@@ -61,13 +59,7 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
     }
 
     @objc func showCenteredPreferencesPanel() {
-        showCenteredPanel(preferencesPanel!)
-    }
-
-    func showCenteredPanel(_ panel: NSPanel) {
-        panel.center()
-        panel.makeKeyAndOrderFront(nil)
-        Application.shared.arrangeInFront(nil)
+        Screen.showCenteredFrontPanel(preferencesPanel!, Screen.getPreferredScreen())
     }
 
     func computeOpenWindows() {
@@ -109,11 +101,12 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
                 return
             }
             selectedOpenWindow = cellWithStep(step)
+            let currentScreen = Screen.getPreferredScreen() // we want all computations and renderings to use the same screen for this summon (in case mouse movements switching screens etc.)
             var workItem: DispatchWorkItem!
             workItem = DispatchWorkItem {
-                if !workItem.isCancelled { self.thumbnailsPanel!.computeThumbnails() }
+                if !workItem.isCancelled { self.thumbnailsPanel!.computeThumbnails(currentScreen) }
                 if !workItem.isCancelled { self.thumbnailsPanel!.highlightCellAt(step) }
-                if !workItem.isCancelled { self.showCenteredPanel(self.thumbnailsPanel!) }
+                if !workItem.isCancelled { Screen.showCenteredFrontPanel(self.thumbnailsPanel!, currentScreen, true) }
             }
             workItems.append(workItem)
             DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + Preferences.windowDisplayDelay!, execute: workItem)
--- a/alt-tab-macos/ui/ThumbnailsPanel.swift
+++ b/alt-tab-macos/ui/ThumbnailsPanel.swift
@@ -5,14 +5,16 @@ class ThumbnailsPanel: NSPanel, NSCollectionViewDataSource, NSCollectionViewDele
     var collectionView_: NSCollectionView!
     var application: Application?
     let cellId = NSUserInterfaceItemIdentifier("Cell")
+    var currentScreen: NSScreen?
 
     override init(contentRect: NSRect, styleMask style: StyleMask, backing backingStoreType: BackingStoreType, defer flag: Bool) {
         super.init(contentRect: contentRect, styleMask: style, backing: backingStoreType, defer: flag)
     }
 
-    convenience init(_ application: Application) {
+    convenience init(_ application: Application, _ currentScreen: NSScreen) {
         self.init()
         self.application = application
+        self.currentScreen = currentScreen
         isFloatingPanel = true
         animationBehavior = .none
         hidesOnDeactivate = false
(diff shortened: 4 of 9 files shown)


gingerr:
Tested on my High Sierra Hackintosh (single monitor) and Catalina MacBook (integrated and pc monitor).

alt-tab-macos/ui/Application.swift
```diff
@@ -61,11 +61,13 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
     }
 
     @objc func showCenteredPreferencesPanel() {
-        showCenteredPanel(preferencesPanel!)
+        showCenteredFrontPanel(preferencesPanel!)
     }
 
-    func showCenteredPanel(_ panel: NSPanel) {
-        panel.center()
```
gingerr: btw. we can't use .center() any longer since that seems to set the screen to NSScreen.main

alt-tab-macos/ui/Application.swift
```diff
@@ -109,11 +111,12 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
...
             }
             selectedOpenWindow = cellWithStep(step)
+            let preferredScreen = Screen.getPreferredScreen() // we want the thumbnail computation and panel display use the same screen (in case mouse movements switching screens etc.)
             var workItem: DispatchWorkItem!
             workItem = DispatchWorkItem {
-                if !workItem.isCancelled { self.thumbnailsPanel!.computeThumbnails() }
+                if !workItem.isCancelled { self.thumbnailsPanel!.computeThumbnails(preferredScreen) }
                 if !workItem.isCancelled { self.thumbnailsPanel!.highlightCellAt(step) }
-                if !workItem.isCancelled { self.showCenteredPanel(self.thumbnailsPanel!) }
+                if !workItem.isCancelled { self.showCenteredFrontPanel(self.thumbnailsPanel!, preferredScreen) }
```
lwouis: It took me a while to understand what was happening here. That's really cool that you covered that use-case!

alt-tab-macos/ui/ThumbnailsPanel.swift
```diff
@@ -91,8 +91,9 @@ class ThumbnailsPanel: NSPanel, NSCollectionViewDataSource, NSCollectionViewDele
         }
     }
 
-    func computeThumbnails() {
-        let maxSize = NSSize(width: NSScreen.main!.frame.width * Preferences.maxScreenUsage!, height: NSScreen.main!.frame.height * Preferences.maxScreenUsage!)
+    func computeThumbnails(_ screenOverride: NSScreen? = nil) {
+        let preferredScreen = screenOverride ?? Screen.getPreferredScreen() // use optional provided screen or determine it just in time
+        let maxSize = NSSize(width: preferredScreen.frame.width * Preferences.maxScreenUsage!, height: preferredScreen.frame.height * Preferences.maxScreenUsage!)
```
lwouis: I think the `maxSize` for the main panel, as well as the main panel origin code inside `showCenteredFrontPanel` would be good candidates to be extracted in the `Screen` class. I'm trying to keep the `Application` and `ThumbnailsPanel` from becoming god classes since most everything happens there. As an anecdote, I started with only `Application` and it quickly became clear that I had to split stuff :p

alt-tab-macos/ui/ThumbnailsPanel.swift
```diff
@@ -91,8 +91,9 @@ class ThumbnailsPanel: NSPanel, NSCollectionViewDataSource, NSCollectionViewDele
         }
     }
 
-    func computeThumbnails() {
-        let maxSize = NSSize(width: NSScreen.main!.frame.width * Preferences.maxScreenUsage!, height: NSScreen.main!.frame.height * Preferences.maxScreenUsage!)
+    func computeThumbnails(_ screenOverride: NSScreen? = nil) {
+        let preferredScreen = screenOverride ?? Screen.getPreferredScreen() // use optional provided screen or determine it just in time
```
lwouis: `computeThumbnails` is only called from one context, in which the parameter is provided. Thus I think you could just remove `? = nil` and line 95, to simplify

alt-tab-macos/logic/Screen.swift
```diff
@@ -14,9 +14,23 @@ class Screen {
 
     static func updateThumbnailMaxSize() -> Void {
         if Preferences.thumbnailMaxWidth == 200 && Preferences.thumbnailMaxHeight == 200 {
-            let main = NSScreen.main!.frame
+            let screenFrame = getPreferredScreen().frame
             Preferences.thumbnailMaxWidth = (NSScreen.main!.frame.size.width * Preferences.maxScreenUsage! - Preferences.windowPadding * 2) / Preferences.maxThumbnailsPerRow! - Preferences.interItemPadding
-            Preferences.thumbnailMaxHeight = Preferences.thumbnailMaxWidth * (main.height / main.width)
+            Preferences.thumbnailMaxHeight = Preferences.thumbnailMaxWidth * (screenFrame.height / screenFrame.width)
         }
     }
```
lwouis: I think there is a bug here. This method is called when `NSApplication.didChangeScreenParametersNotification` happens. The use-case is the user changes their display, so we should recalculate the size of thumbnails based on `Preferences.maxThumbnailsPerRow`. Already when I decided to remove the thumbnails size from the Preferences panel and make is a derived value from `maxThumbnailsPerRow`, I thought I may want to remove the thumbnail width/height from the `Preferences`. I didn't though.

Now that you add this PR, it becomes clearer that the width/height of thumbnail is a dynamic value and not at all a preferences.

I would suggest these changes:

* Remove `Preferences.thumbnailMaxWidth` and `Preferences.thumbnailMaxHeight` from being `Preferences`
* Rename `updateThumbnailMaxSize` and make it return `(width, height)` instead of storing in `Preferences`
* Call `updateThumbnailMaxSize` in place which use `Preferences.`width/height today: `layout.estimatedItemSize`, `layoutAttributesForElements` and `computeDownscaledSize`

alt-tab-macos/logic/Preferences.swift
```diff
@@ -46,6 +53,10 @@ class Preferences {
         MacroPreference("⌃ control", ([kVK_Control, kVK_RightControl], .control)),
         MacroPreference("⌘ command", ([kVK_Command, kVK_RightCommand], .command))
     ])
+    static var showOnScreenMacro = MacroPreferenceHelper<(ShowOnScreenPreference)>([
+        MacroPreference("with keyboard focus", (ShowOnScreenPreference.MAIN)),
+        MacroPreference("with mouse pointer", (ShowOnScreenPreference.MOUSE)),
+    ])
```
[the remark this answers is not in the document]
gingerr: So thats what that is. Didn't come into my mind the Generic <> and additional () notation was holding a tuple and took me a while to actually half-way understand what MacroPreferenceHelper and its short code and the Swift notation usage does. Was not sure if this was Swift way of writing an array, dictionary, set or I don't know what I expected. Still trying to get familiar with all the language features with its different parentheses use cases.

Thanks!

alt-tab-macos/logic/Screen.swift
```diff
@@ -14,9 +14,23 @@ class Screen {
...
+
+    static func getPreferredScreen() -> NSScreen {
+        switch Preferences.showOnScreen! {
+        case .MOUSE:
+            return getScreenWithMouse() ?? NSScreen.main!; // .main as fail-safe
```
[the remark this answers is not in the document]
gingerr: Yeah I assume NSScreen.main is always filled during normal system operations and wanted also to preserve the pre-commit NSScreen logic so this commit does not introduce changes out of its scope. 

But now that you mention it, yeah its a funny/hopeful fail-safe 🤔

The only case I can currently guess where NSScreen.main is nil would be if all displays are disconnected/sleeping or on the OS lock-screen (didn't test if the application listeners/events get executed there).

Thanks for mentioning that!

alt-tab-macos/logic/Screen.swift
```diff
@@ -14,9 +14,23 @@ class Screen {
...
+        default: // no case for .MAIN since its currently the same as default case
```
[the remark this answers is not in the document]
gingerr: Yeah thats what happened here.

It does not allow a DRY combined case/default. But now I worked on it a little bit the IDE said rightfully the default case is never called (since we said Preferences.showOnScreen! is always filled).

We could also throw here incase NSScreen.main is empty but then we will need to handle it somewhere and decide which methods in the call hierarchy rethrow the exception.

I gonna commit a proposal. Tell me at re-viewing the changes the way you would like to go here (or you can also commit -> you have as a maintainer write permissions to this feature branch).

alt-tab-macos/logic/Screen.swift
```diff
@@ -14,9 +14,23 @@ class Screen {
...
+            return NSScreen.main!;
+        }
+    }
+
+    // https://stackoverflow.com/a/49624487/4767781
```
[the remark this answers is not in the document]
gingerr: Yeah, originally I added a comment there but then it just said more or less the same as the method name already expressed so I removed it but mentioned the original stack overflow author of that code line.

Gonna remove it. Thanks for your POV!

alt-tab-macos/ui/Application.swift
```diff
@@ -61,11 +61,13 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
...
-        showCenteredPanel(preferencesPanel!)
+        showCenteredFrontPanel(preferencesPanel!)
     }
 
-    func showCenteredPanel(_ panel: NSPanel) {
-        panel.center()
+    func showCenteredFrontPanel(_ panel: NSPanel, _ screenOverride: NSScreen? = nil) {
+        let preferredScreen = screenOverride ?? Screen.getPreferredScreen() // use optional provided screen or determine it just in time
+        let centerPosition = NSPoint(x: preferredScreen.visibleFrame.midX - panel.frame.width / 2, y: preferredScreen.visibleFrame.midY - panel.frame.height / 2)
+        panel.setFrameOrigin(centerPosition)
```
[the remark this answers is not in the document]
gingerr: Nice catch. 😎 

Yeah I've noticed that too. Ive changed it on purpose to be vertically centered cause the native app switcher is not using the offset. But we need the offset for other windows e.g. the PreferencesPanel.

I will make this an argument of this method (if one would like to apply the appleVerticalOffset), make it true by default and apply an visibleFrame based ratio. The PreferencesPanel would like to apply it and ThumbnailsPanel would like to be centered without any offset. Agreed?

alt-tab-macos/ui/Application.swift
gingerr: Thanks! 

I found it cause if was monitoring how often getPreferredScreen() was called and during the AltTab trigger it got called twice. There could be a time difference between two calls and it would cause problems if the thumbnails are computed for an huge screen (PC monitor) and the panel is then shown an a MacBook monitor (when the user is using the mouse screen and moved the mouse pointer between both method executions).

gingerr:
Gonna go to work now and continue working on this pull request later today. Should I rebase and force push this feature branch on my fork if there other commits on your upstream in the meantime or will you do it when merging?

@lwouis Thanks for the deep and detailed review! Appreciate it 👍

lwouis:
> Gonna go to work now and continue working on this pull request later today. Should I rebase and force push this feature branch on my fork if there other commits on your upstream in the meantime or will you do it when merging?

Rebasing on your side is always best. Otherwise if there is a conflict i won't be able to press the merge button on your PR

> @lwouis Thanks for the deep and detailed review! Appreciate it 👍

No, thank you for collaborating on this project with me! It's really a pleasant experience!

alt-tab-macos/logic/Screen.swift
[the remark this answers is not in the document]
lwouis: I think the current code is fine. I was curious about not listing the 2 cases directly, but yeah I guess swift compiler isn't smart enough. I remember doing lots of pattern matching in Scala a few years ago and I remember it being always able to check if it was exhaustive or not

alt-tab-macos/ui/Application.swift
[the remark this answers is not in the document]
lwouis: I prefer `ThumbnailsPanel` being higher-than-centered. Like I said it is a thing I encountered in typography. Book titles and even pages are usually higher than centered. However since I don't know, and couldn't Google, an explanation on *how* that is supposed to be more ergonomic, I guess realizing my position is weak. I'll let you decide.

Note that HyperSwitch does centered but WindowSwitcher does higher-than-centered.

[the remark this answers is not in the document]
gingerr: Iam fully ok with that (apple vertical offset mimic). Either way looks great. I didn't knew if you placed `ThumbnailPanel` higher on purpose (by using .center() for both panels).

alt-tab-macos/logic/Screen.swift
gingerr: Independent of that I see Ive overlooked one usage of `NSScreen.main` here at line 18.

gingerr: My bad, this needs to be done in this commit/pull request or the feature will be released bugged. Gonna do the required changes here.

alt-tab-macos/ui/ThumbnailsPanel.swift
gingerr: **Moved to Screen**
- maxSize -> calcFrameMaxSize()
- showCenteredFrontPanel

alt-tab-macos/logic/Screen.swift
gingerr: Ok did a break but are now back on it. I profiled the application, it are indeed cheap calls. 

I refactored as you suggested with additional required changes:
We need to make sure the same NSScreen is used for all computations during one possibly long ongoing UI summon process. 

The call to `getPrefferedScreen()` was called many times during different computations at different places. I am not familiar enough with the call hierarchy yet and don't wanted to let them all to drag the `screenOverride` argument with them (through all method invocations inside the whole UI summon process) so I added `Screen.preferredScreenOverride` which  is set by `Application.showUiOrCycleSelection()` and unset by `Application.hideUi()` and then used by default by `getPreferredScreen()` as long as the value is not `nil` and the method caller does not say otherwise.

Tell me if its ok for you for now.

gingerr:
> > Gonna go to work now and continue working on this pull request later today. Should I rebase and force push this feature branch on my fork if there other commits on your upstream in the meantime or will you do it when merging?
> 
> Rebasing on your side is always best. Otherwise if there is a conflict i won't be able to press the merge button on your PR
> 
> > @lwouis Thanks for the deep and detailed review! Appreciate it 👍
> 
> No, thank you for collaborating on this project with me! It's really a pleasant experience!

Yeah, for me too! 😎 

Ok, I had to figure that fork/upstream sync out and think its rebased correctly now. Its a learning experience for me with feature branches on forks that need to be synced with upstream.

Pull request is now open again for re-review.

alt-tab-macos/logic/Screen.swift
gingerr: But I must say I dont like that `preferredScreenOverride` as a resolution of this. 

It feels like the structure has to be reviewed and changed but I need more time for that to get familiar with all of your code for rendering. 

`Screen` class and all of its methods is static, so its more like a helper class. The current solution with `preferredScreenOverride` limits the application to one view at a time or lets say one screen. I think the callers of `Screen` (`calcThumbnailMaxSize`, `calcFrameMaxSize`, `showCenteredFrontPanel`) would need to provide which screen this methods should use, possibly as an optional argument.

gingerr: Sorry for another commit to that. Ive redone and tested it again in a dual monitor situation with both monitors having hugely different resolutions.

Corresponding part of commit message:
```
- Screen.preferredScreenOverride: removed
- instead currentScreen gets provided or updated now at all required places before triggering UI summon calculations
  - ThumbnailsPanel.computeThumbnails(): re-creates collectionViewLayout instead of invalidating it
```

Eventually we should reinstantiate ThumbnailsPanel on each AltTab invocation. I'm curious about your judgement.

Cheers.

alt-tab-macos/ui/Application.swift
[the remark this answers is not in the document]
gingerr: In the current commit i had to tweak this behaviour slightly. If the panel to be rendered is huge in its height the vertical offset does not look very good. So I've added a check: If the panel is bigger than 2/3 of the screen, the offset will be not applied.

Is that alright for you?
Thanks!

gingerr:
Ive read on hackernews you are on a single monitor setup? One single monitor in addition to a macbook monitor or one monitor total? I'm asking cause this issue requires two monitors (with different effective resolutions) to verify the correct implementation. Let me know if you want me to run some additional tests scenarios or to provide some additional screenshots.

lwouis:
> Ive read on hackernews you are on a single monitor setup?

I'm on 1 monitor total. For the anecdote, I used to feel handicapped when I switched from my 2 monitor setup to the laptop in meetings or other scenarios. Also I noticed I was turning my head slightly left or right for long periods of time, so I started putting one monitor really in front, and one clearly on the side. With that setup I realize it's not that convenient to switch. At the same time I colleague told me he switched back to 1 monitor + a complete window-manager shortcut coverage, and he's been happy. Since then I've experimented with having only 1, even a 4k monitor at home. Overall I think it is indeed a better experience for me. My flow is not as static as some other people though. For people using the same software the same way for years, then I clearly multi-screen provides value.

Anyway, I can reproduce the 2-monitor setup at work or at home no problem :)

lwouis:
> Pull request is now open again for re-review.

I just noticed that your latest commit is with a different email address. You may want to change that ;p

lwouis:
Got that exception:

```
"focusTarget"
"hideUi"
"showUiOrSelectNext"
2019-11-01 09:36:06.262 alt-tab-macos[92330:5981006] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArrayM objectAtIndex:]: index 5 beyond bounds [0 .. 4]'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff4e206a7d __exceptionPreprocess + 256
	1   libobjc.A.dylib                     0x00007fff788fda17 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff4e245a74 -[CFPrefsConfigurationFileSource initWithConfigurationPropertyList:containingPreferences:] + 0
	3   CoreFoundation                      0x00007fff4e1421de -[CFPrefsSearchListSource addNamedVolatileSourceForIdentifier:] + 0
	4   UIFoundation                        0x00007fff74ecce79 -[_NSFlowLayoutSection frameForItemAtIndexPath:] + 91
	5   UIFoundation                        0x00007fff74f4fe18 -[NSCollectionViewFlowLayout _frameForItem:inSection:usingData:] + 149
	6   UIFoundation                        0x00007fff74f4c36e -[NSCollectionViewFlowLayout layoutAttributesForItemAtIndexPath:usingData:] + 334
	7   UIFoundation                        0x00007fff74f4c640 -[NSCollectionViewFlowLayout layoutAttributesForItemAtIndexPath:] + 465
	8   UIFoundation                        0x00007fff74e55d62 -[_NSCollectionViewCore _setCollectionViewLayout:animated:isInteractive:completion:] + 3906
	9   AppKit                              0x00007fff4b9a1a5c -[NSCollectionView _setCollectionViewLayout:animate:] + 346
	10  alt-tab-macos                       0x0000000100147e09 $s13alt_tab_macos15ThumbnailsPanelC07computeD0yySo8NSScreenCF + 697
	11  alt-tab-macos                       0x000000010012fe30 $s13alt_tab_macos11ApplicationC22showUiOrCycleSelectionyySiFyycfU_ + 704
	12  alt-tab-macos                       0x00000001001219bd $sIeg_IeyB_TR + 45
	13  libdispatch.dylib                   0x00007fff7a089cea _dispatch_block_async_invoke2 + 83
	14  libdispatch.dylib                   0x00007fff7a07e63d _dispatch_client_callout + 8
	15  libdispatch.dylib                   0x00007fff7a08968d _dispatch_main_queue_callback_4CF + 1135
	16  CoreFoundation                      0x00007fff4e150a37 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
	17  CoreFoundation                      0x00007fff4e150161 __CFRunLoopRun + 2289
	18  CoreFoundation                      0x00007fff4e14f61e CFRunLoopRunSpecific + 455
	19  HIToolbox                           0x00007fff4d3ae1ab RunCurrentEventLoopInMode + 292
	20  HIToolbox                           0x00007fff4d3adee5 ReceiveNextEventCommon + 603
	21  HIToolbox                           0x00007fff4d3adc76 _BlockUntilNextEventMatchingListInModeWithFilter + 64
	22  AppKit                              0x00007fff4b74677d _DPSNextEvent + 1135
	23  AppKit                              0x00007fff4b74546b -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 1361
	24  AppKit                              0x00007fff4b73f588 -[NSApplication run] + 699
	25  alt-tab-macos                       0x0000000100117d47 $s13alt_tab_macosyyXEfU_ + 71
	26  alt-tab-macos                       0x0000000100117d6f $ss5Error_pIgzo_ytsAA_pIegrzo_TR + 15
	27  alt-tab-macos                       0x0000000100117dd4 $ss5Error_pIgzo_ytsAA_pIegrzo_TRTA + 20
	28  libswiftObjectiveC.dylib            0x00007fff79ddaf4f $s10ObjectiveC15autoreleasepool8invokingxxyKXE_tKlF + 47
	29  alt-tab-macos                       0x0000000100117ce3 main + 83
	30  libdyld.dylib                       0x00007fff7a0cb3d5 start + 1
)
libc++abi.dylib: terminating with uncaught exception of type NSException

Process finished with exit code 134 (interrupted by signal 6: SIGABRT)
```

I think my flow was: alt-tab on one screen, release, then move mouse to the other screen and alt-tab -> crash.

**1min later:** I got another exception, almost the same. This time I went into the preferences and switched from keyboard to mouse focus, then moved mouse to the other screen, alt-tab -> crash.

**5min later:** now it works without crashing anymore even when I try to mix every interaction. I'll try to reproduce more, and check the code in the stacktrace to see which index is going out of bounds to imagine the use-case that produces that

gingerr:
> I'm on 1 monitor total. For the anecdote, I used to feel handicapped when I switched from my 2 monitor setup to the laptop in meetings or other scenarios. Also I noticed I was turning my head slightly left or right for long periods of time, so I started putting one monitor really in front, and one clearly on the side. With that setup I realize it's not that convenient to switch. At the same time I colleague told me he switched back to 1 monitor + a complete window-manager shortcut coverage, and he's been happy. Since then I've experimented with having only 1, even a 4k monitor at home. Overall I think it is indeed a better experience for me. My flow is not as static as some other people though. For people using the same software the same way for years, then I clearly multi-screen provides value.

Noticed the same when working in front of two monitors (slim bezels but one monitor slightly to the left and one to the right), its not that optimal. Ive tried a single monitor 4k but that felt too big. I ended up on one 34 inches ultra wide (3440x1440) monitor which I can only recommend.

> Anyway, I can reproduce the 2-monitor setup at work or at home no problem :)
Perfect 👍 

> I just noticed that your latest commit is with a different email address. You may want to change that ;p

Yeah I noticed that too in the commit log in the IDE. My company MacBook has setup a different global git email address. Its alright, both my private and company email addresses are provided in this GitHub account. Thanks for the hint!

Btw. the code comments/conversations here have more replies now. This is my first time contributing to an GitHub project. Were you spammed by GitHub every time I commented something? Or only when there are named mentions? 

There are two new additional commits/iterations since the first pull request review. With the latest one I improved upon the `Screen.preferredScreen()` usage.

gingerr:
> Got that exception:
> 
> ```
> "focusTarget"
> "hideUi"
> "showUiOrSelectNext"
> 2019-11-01 09:36:06.262 alt-tab-macos[92330:5981006] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArrayM objectAtIndex:]: index 5 beyond bounds [0 .. 4]'
> *** First throw call stack:
> (
> 	0   CoreFoundation                      0x00007fff4e206a7d __exceptionPreprocess + 256
> 	1   libobjc.A.dylib                     0x00007fff788fda17 objc_exception_throw + 48
> 	2   CoreFoundation                      0x00007fff4e245a74 -[CFPrefsConfigurationFileSource initWithConfigurationPropertyList:containingPreferences:] + 0
> 	3   CoreFoundation                      0x00007fff4e1421de -[CFPrefsSearchListSource addNamedVolatileSourceForIdentifier:] + 0
> 	4   UIFoundation                        0x00007fff74ecce79 -[_NSFlowLayoutSection frameForItemAtIndexPath:] + 91
> 	5   UIFoundation                        0x00007fff74f4fe18 -[NSCollectionViewFlowLayout _frameForItem:inSection:usingData:] + 149
> 	6   UIFoundation                        0x00007fff74f4c36e -[NSCollectionViewFlowLayout layoutAttributesForItemAtIndexPath:usingData:] + 334
> 	7   UIFoundation                        0x00007fff74f4c640 -[NSCollectionViewFlowLayout layoutAttributesForItemAtIndexPath:] + 465
> 	8   UIFoundation                        0x00007fff74e55d62 -[_NSCollectionViewCore _setCollectionViewLayout:animated:isInteractive:completion:] + 3906
> 	9   AppKit                              0x00007fff4b9a1a5c -[NSCollectionView _setCollectionViewLayout:animate:] + 346
> 	10  alt-tab-macos                       0x0000000100147e09 $s13alt_tab_macos15ThumbnailsPanelC07computeD0yySo8NSScreenCF + 697
> 	11  alt-tab-macos                       0x000000010012fe30 $s13alt_tab_macos11ApplicationC22showUiOrCycleSelectionyySiFyycfU_ + 704
> 	12  alt-tab-macos                       0x00000001001219bd $sIeg_IeyB_TR + 45
> 	13  libdispatch.dylib                   0x00007fff7a089cea _dispatch_block_async_invoke2 + 83
> 	14  libdispatch.dylib                   0x00007fff7a07e63d _dispatch_client_callout + 8
> 	15  libdispatch.dylib                   0x00007fff7a08968d _dispatch_main_queue_callback_4CF + 1135
> 	16  CoreFoundation                      0x00007fff4e150a37 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
> 	17  CoreFoundation                      0x00007fff4e150161 __CFRunLoopRun + 2289
> 	18  CoreFoundation                      0x00007fff4e14f61e CFRunLoopRunSpecific + 455
> 	19  HIToolbox                           0x00007fff4d3ae1ab RunCurrentEventLoopInMode + 292
> 	20  HIToolbox                           0x00007fff4d3adee5 ReceiveNextEventCommon + 603
> 	21  HIToolbox                           0x00007fff4d3adc76 _BlockUntilNextEventMatchingListInModeWithFilter + 64
> 	22  AppKit                              0x00007fff4b74677d _DPSNextEvent + 1135
> 	23  AppKit                              0x00007fff4b74546b -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 1361
> 	24  AppKit                              0x00007fff4b73f588 -[NSApplication run] + 699
> 	25  alt-tab-macos                       0x0000000100117d47 $s13alt_tab_macosyyXEfU_ + 71
> 	26  alt-tab-macos                       0x0000000100117d6f $ss5Error_pIgzo_ytsAA_pIegrzo_TR + 15
> 	27  alt-tab-macos                       0x0000000100117dd4 $ss5Error_pIgzo_ytsAA_pIegrzo_TRTA + 20
> 	28  libswiftObjectiveC.dylib            0x00007fff79ddaf4f $s10ObjectiveC15autoreleasepool8invokingxxyKXE_tKlF + 47
> 	29  alt-tab-macos                       0x0000000100117ce3 main + 83
> 	30  libdyld.dylib                       0x00007fff7a0cb3d5 start + 1
> )
> libc++abi.dylib: terminating with uncaught exception of type NSException
> 
> Process finished with exit code 134 (interrupted by signal 6: SIGABRT)
> ```
> 
> I think my flow was: alt-tab on one screen, release, then move mouse to the other screen and alt-tab -> crash.
> 
> **1min later:** I got another exception, almost the same. This time I went into the preferences and switched from keyboard to mouse focus, then moved mouse to the other screen, alt-tab -> crash.
> 
> **5min later:** now it works without crashing anymore even when I try to mix every interaction. I'll try to reproduce more, and check the code in the stacktrace to see which index is going out of bounds to imagine the use-case that produces that

Oh, I didn't had any exceptions, just UI issues when the code was using the wrong screen but that should all be handed now. I was switching the screens as-well to see if the computations/renderings use different effective screen resolutions. Gonna try to reproduce an exception somehow. Thanks!

gingerr:
Ok got it, happens after opening the Preferences and then AltTabbing. Gonna hunt down the cause.

lwouis:
> Ok got it, happens after opening the Preferences and then AltTabbing. Gonna hunt down the cause.

I can reproduce consistently. It's not about the pref panel. It's about having 1 more or 1 less window, and it turns out the pref panel is a window so opening it triggers the crash

To reproduce: alt-tab > works > release > open a window > alt-tab > crashes because of some index OOB now that we have an extra window

gingerr:
Your right, Thank you. 

The change that is causing this is this one:

`ThumbnailsPanel.computeThumbnails()`
from
`collectionView_!.collectionViewLayout!.invalidateLayout()`
to
`collectionView_!.collectionViewLayout = makeLayout()`

I changed it cause of a wrong rendering but when retesting the application never had different amounts of windows 😞 Lesson learned.

lwouis:
I'm making some changes to your PR. I think in that case it's faster for me to show you my changes than explain what they would look like. You suggested I pushed on your fork directly, but I don't want to impact your work, and your space, so I pushed it on this fork here on a [branch with the same name](https://github.com/lwouis/alt-tab-macos/commits/feature/windows-on-mouse-screen). Right now I have changed the "apple centering". I'm gonna look at the other elements of your PR now

gingerr:
Alright. Thank you for re-reviewing it again with your corrections and improvements.

gingerr:
Once the array out of bounds issue was fixed in my branch i had sometimes an other exception but then one of your branch commits pointed me at a async issue.

New commit/push:
> - Application: thumbnailsPanel gets instantiated now for each UI summon and inside the same thread as the other thumbnailsPanel operations
> - ThumbnailsPanel: reverted computeThumbnails() to its old behaviour (since my improper solution caused an array out of bounds exception)
> - Application: cleanup: removed unused backgroundView

alt-tab-macos/ui/Application.swift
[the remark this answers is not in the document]
lwouis: > So I've added a check: If the panel is bigger than 2/3 of the screen, the offset will be not applied.

In my branch I have mimic'ed Apple implementation. They simply position at 75% instead of 50% (0 is bottom, not top). Could you check my implem? I think it's nicer because in your implem there is a big jump in high after you reach a certain size

alt-tab-macos/ui/ThumbnailsPanel.swift
lwouis: > Btw. does the maxSize work as designed?

One of the most complex part of the app is the layout of the thumbnails. This is handled by `ThumbnailsPanel` that has a `NSCollectionView` member, but implements its `NSCollectionViewDataSource, NSCollectionViewDelegate, NSCollectionViewDelegateFlowLayout` protocols itself. There is also a custom class for `CollectionViewCenterFlowLayout`.

All these moving pieces have implementation hooks you're supposed to use. It is very OOP so you're supposed to subclass or implements protocols in order to customize your `NSCollectionView`. Documentation from Apple as usual is terrible, so I had to jump through many loops to get it exactly the way I wanted.

Indeed, when the panel can become bigger than the screen, instead of stopping at the bottom and providing scroll maybe like you see in the Photos app, or another app from Apple. My point is that the documentation is terrible and I did the current implementation. When we go forward to implementing #69, we can change a lot if there is a better way to approach this.

I'm really struggling to understand how one is supposed to use some parts of Cocoa sometimes

lwouis:
Hey @gingerr, so I tested my branch with has some commits and top of your branch. I think it works fine.

Could you check it out? Ideally I would like to merge just 1 commit from you which called `feat: windows on mouse screen, implements #28` so the git history is nice is readable.

So because of that I would suggest you either take inspiration from my commits and integrate that feedback into your branch, or if you're fine with the exact commits, just add them to your branch, and squash every commit into 1 (or more if you think it is useful in the git history), and re-submit the PR. I think at this point we are ready to merge.

Btw this is anecdotal but I felt it took like 1s to display the thumbnails on my macbook compared to the external monitor. Not sure why but anytime I see the app being slow to display I'm sad :/

alt-tab-macos/ui/Application.swift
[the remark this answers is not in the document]
gingerr: @lwouis I see, your implementation is perfect. Mine was slightly off at different frame sizes. Thanks for that!

alt-tab-macos/ui/ThumbnailsPanel.swift
gingerr: @lwouis 

> One of the most complex part of the app is the layout of the thumbnails. 

And thats a part where my lacking Swift/Cocoa knowledge shows itself nicely and i could not explain the flow fully right now. Btw. AppCode does not support Call Hierarchy yet for Swift (bummer!). Had to switch to Xcode to get a grasp at the execution flow with the Call Hierarchy there and where to make the required changes to have the same preferredScreen everywhere.

Regarding the Documentation from Apple: Do you know if one can find the source code of their implementation or is this closed source? E.g. for NSCollectionViewFlowLayout.

I thought this feature would be not that of a big change but it send me all over the codebase (which is good).

> Indeed, when the panel can become bigger than the screen, instead of stopping at the bottom and providing scroll maybe like you see in the Photos app, or another app from Apple. My point is that the documentation is terrible and I did the current implementation. When we go forward to implementing #69, we can change a lot if there is a better way to approach this.
> I'm really struggling to understand how one is supposed to use some parts of Cocoa sometimes. Especially because every tutorial on Earth show how to build UI with InterfaceBuilder, and I'm actively wanting to not use that tool. Maybe it's a mistake though?

For complex / dynamic windows like ThumbnailsPanel i think its right to go the programmatic way but my knowledge here if greatly limited, who knows maybe it is nicely interoperable.

The PreferencesPanel could be probably done easily with InterfaceBuilder (static) but you said there is some dependency on Xcode/Framework versions which would up the min required macOS version for AltTab right?

lwouis: > Do you know if one can find the source code of their implementation or is this closed source

I believe macOS libraries such as Cocoa are all closed-source, but I may be wrong, or some people might have disassembled lots of it by now. Not sure

> there is some dependency on Xcode/Framework versions which would up the min required macOS version for AltTab right?

The reason I avoided InterfaceBuilder is that it's a GUI tool unique to XCode. I want to work in AppCode, and also I want for other people to be apple to contribute even without having XCode installed. The way the project is right now you can use any text editor and edit Swift code. Interface Builder produces either binary or XML files which are blackboxes from a version control point of view, and also when you don't have XCode handy. Like imagine a complex changes to a `.xib` (Interface Builder output) to review in a PR. You now have to checkout the code in XCode in order to have any idea what was done. Then in XCode you need to check all the UI checkboxes, inspectors, etc, and guess what was changed.

No constraints on min macOS version though. That only depends on which OS APIs are called.

gingerr: @lwouis I see now where you are coming from, thanks! Going through every UI objects and its setting panels (there are plenty of them there) manually without VCS would a big no.

gingerr:
@lwouis Man, that right now was a practice hour in git for me. I wanted to squash them into two commits and keeping the authors. One commit should be my implementation attempt for pull request review and one commit your corrections and improvements to it. Its done now but Ive lost one commit (that i made after your commits to the upstream feature branch) but thats alright. Is it ok with you being two commits?

This pull request is ready for a merge now. Its files are identical with your upstream feature branch.

> Btw this is anecdotal but I felt it took like 1s to display the thumbnails on my macbook compared to the external monitor. Not sure why but anytime I see the app being slow to display I'm sad :/

Yeah, this should be one of our top priorities. My Hackintosh PC has an Intel i7-7700k and the performance of AltTab is nice but on the MacBook Pro 2015 with an I7-4870HQ when my usual development and office applications are open the delay is noticeable but still ok! 👍

lwouis:
> One commit should be my implementation attempt for pull request review and one commit your corrections and improvements to it

Haha man you're being way too humble! You should've just squashed with your name on it. Anyway, I appreciate the gesture :)

> Yeah, this should be one of our top priorities

I already profiled though, and my diagnostic is that the bottleneck is the CollectionView. I think we may have to scrap all this and make a similar view from scratch. Before starting that, I would need to profile on a slow macbook somehow to see if I get the same diagnostic. I don't wanna invest lots of time remaking the layout logic and not solving the actual performance issue 😅

Anyway that's for another day ;p

Thanks again for that PR!

gingerr:
@lwouis Hehe, thanks! But i also need this git practice so its an welcome challenge and was the right thing to do anyway 😄 

> I already profiled though, and my diagnostic is that the bottleneck is the CollectionView. I think we may have to scrap all this and make a similar view from scratch. Before starting that, I would need to profile on a slow macbook somehow to see if I get the same diagnostic. I don't wanna invest lots of time remaking the layout logic and not solving the actual performance issue 😅

Interesting, i thought by reading the corresponding issue it was the wakeup of other windows and screenshot capture of them. But thats as you said for another day and ticket. Maybe some macOS stress test utility could help here to slow down the device by some percentage (usually maxing it out to 100%).

Will you notify Somberlald in his issue #28 ?

lwouis:
> Will you notify Somberlald in his issue #28 ?

I was so surprised to see the ticket still open. You wrote "implements #28" in the commit message. I thought it would close it like "closes #28" does. I guess it's not in the [keywords list](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords#about-issue-references). Good to know; next time use one of these ;p

gingerr:
@lwouis Other than that I don’t have permissions (which is good) on your upstream, so if it would have been the right keyword it should not close it I think. Thanks for the keywords list!

PR #68, opened 2019-10-30
Add ShortcutRecorder lib to handle shortcuts (closes #17, closes #72)

### LINKED ISSUE #17: Global shortcut fails when focused field is a password field
image
Pressing alt-tab here doesn't trigger the app. Somehow the tab event is not propagated to the app in that case, potentially for security purpose.

There may or may not be a way to work around this behaviour.

### LINKED ISSUE #72: Add preferences: more flexible controls
From rich discussion in #61 

Right now users can change the modifier key. It would be nice to **be able to set other keys** such as next, previous, cancel.

Supporting **multiple shortcuts** for each action would also be more flexible

lwouis:


lwouis:
I'm kind of stuck because of https://github.com/Kentzo/ShortcutRecorder/issues/114. I think next step would be to either wait, contribute the feature to SR repo, use another lib (here though, I think that SR is really the better lib after reviewing options), or just use SR input validation, and listening facilities, without using the capture UI.

lwouis:
Closing this PR as this was completed in another PR.

Issue #49, opened 2019-10-23
Better UX to set user preferences


lwouis:
Current UI to set user preferences is raw:

![image](https://user-images.githubusercontent.com/106195/67402534-b25bef80-f5eb-11e9-9217-fa778197eb1c.png)

**Goal 1**

Fewer options and better validation. In order of priority:

* tab key -> dropdown or "press the key" button
* meta key + meta key modifier -> dropdown or "press the key" button
* thumbnail quality -> dropdown
* maximum size of main window -> clarify unit; maybe switch to a slider
* padding in the main window -> clarify unit; maybe switch to a slider
* padding in each cell -> clarify unit; maybe switch to a slider
* border width of each cell -> clarify unit; maybe switch to a slider
* max number of thumbnails -> maybe switch to a slider
* maximum width of thumbnails -> clarify unit; maybe switch to a slider
* maximum height of thumbnails -> clarify unit; maybe switch to a slider
* app icon size -> clarify unit; maybe switch to a slider
* font height -> clarify unit; maybe switch to a slider
* padding  between cells -> clarify unit; maybe switch to a slider
* delay in ms -> clarify unit; maybe switch to a slider

Here is HyperSwitch more refined UI as reference:

![image](https://user-images.githubusercontent.com/106195/67402468-9ce6c580-f5eb-11e9-8fc0-bd2d17825a64.png)

**Goal 2**

Better documentation. Replace current text with shorter explanations, and add either a hover tooltip or some other variation of more-info-on-demand UI, so that users who found the short text unclear can get more explanations. We can think about adding illustrations/images/videos here too to help get the point across.

**Goal 3**

Make categories like dimensions, keyboard, performance, etc. This can be tabs as is traditional on macOS, or any other approach.

**Goal 4**

Make all preferences live. Currently some of them are live, and some require a restart to apply. An easy fix would be to kill and recreate the main window on preferences change. Another brute-force approach would be to add a "restart the app" button at the bottom.

gingerr:
I am working on some parts of this. A primer into macOS programmatic UI.

gingerr:
Regarding **Goal 4**: Ive implemented the brute force approach -> Restart app button.

What do you mean by recreating the main window. Just a new instantiation of `ThumbnainPanel` in `Application`?

lwouis:
> What do you mean by recreating the main window. Just a new instantiation of `ThumbnainPanel` in `Application`?

[...]

Issue #73, opened 2019-11-01
Better preferences migration


lwouis:
From discussion in #71:

**If the json parsing fails, we can:**

* show the user there was a problem
* possibly send that to some stats server for telemetry but this is a controversial topic in the OSS community
* remove the existing file and use the defaut from the current app

**Better handle when parsing works**

* remove the version key/value and instead do some smart merging of old prefs and new defaults

mfn:
> show the user there was a problem

To would probably the easiest for a start and, until the software is deemed stable, probably as the most economical way. IMHO :)

Issue #51, opened 2019-10-24
Error loading preferences from JSON file on disk


xakraz:
## Error

When installing the latest version on my workstation, the app does not start.

By CLI, it seems to be an issue with the config file parsing.

```
"Error loading preferences from JSON file on disk" "Optional contained nil"
```

## Data

```
➜  ~ cat ~/Library/Preferences/alt-tab-macos.json
{
  "windowDisplayDelay" : "200",
  "cellPadding" : "6",
  "thumbnailMaxHeight" : "0",
  "metaModifierFlag" : "262144",
  "maxScreenUsage" : "0.8",
  "cellBorderWidth" : "2",
  "thumbnailMaxWidth" : "0",
  "highlightColor" : "white",
  "maxThumbnailsPerRow" : "4",
  "fontHeight" : "15",
  "thumbnailQuality" : "4",
  "interItemPadding" : "4",
  "metaKey" : "59",
  "tabKey" : "48",
  "iconSize" : "32",
  "windowPadding" : "20"
}%
```

```
➜  ~ cat ~/Library/Preferences/alt-tab-macos-defaults.json
{
  "windowDisplayDelay" : "200",
  "cellPadding" : "6",
  "thumbnailMaxHeight" : "0",
  "metaModifierFlag" : "262144",
  "maxScreenUsage" : "0.8",
  "cellBorderWidth" : "2",
  "thumbnailMaxWidth" : "0",
  "highlightColor" : "white",
  "maxThumbnailsPerRow" : "4",
  "fontHeight" : "15",
  "thumbnailQuality" : "4",
  "interItemPadding" : "4",
  "metaKey" : "59",
  "tabKey" : "48",
  "iconSize" : "32",
  "windowPadding" : "20"
}%
```
## Info

```
➜  ~ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G103
```

```
25 › <key>CFBundleVersion</key>
 26 › <string>1.5.1</string>
```

lwouis:
Hey @xakraz, sorry about this issue.

I had some very primitive logic for handling migration. That is: you download a new version of the app, and it tries to detect that your preferences.json file is not an existing file, but a previous version of an existing file. I was just counting the lines and assuming the file will only get bigger (i.e. retrocompatible forever).

[...]

PR #76, opened 2019-11-02
feat: improves PreferencesPanel UX, partially implements #49

gingerr:
- ux: first pass on UI
- ux: groups available preferences by keys-, visual-, behaviour-preferences (vertical separation without labels)
- ux: adds Restart action
- code: removes unused variables
- code: PreferencesPanel is no longer a Delegate of an view child

Hey @lwouis 

This is my first improvement iteration of the PreferencesPanel. My target was to improve the visuals. This is ready for review and integration into upstream/master if you like.

Mariusz

Issue reference: #49 

Screen Shot 2019-11-02 at 18 17 54


gingerr:
@lwouis Next iteration is committed/pushed and was quite fun. The sliders are now in but there is an UX thing missing for them. They need to show the currently selected value somehow:
- below the tick marks every {x} positions
- or some other input field or text that changes on the fly
- or some tooltip while holding the slider control

Apple talks about tickMark labels in their [UI guidelines](https://developer.apple.com/design/human-interface-guidelines/macos/selectors/sliders/) but at the first look the [NSSlider](https://developer.apple.com/documentation/appkit/nsslider) class does not offer anything for that. I will look into this next (probably tomorrow).

As usual, I am Interested in your feedback and critique.

Screen Shot 2019-11-02 at 23 07 44


**Commit message/changelog**

> - ux/code: adds sliders (maxScreenUsage, iconSize, fontHeight)
> - ux: defined sensible min/max/step values for sliders
> - code: fixes panelWidth (panel and its content did not grow correctly)
> - code: removes inputsMap (NSControl.identifier is now used for that)
> - code: adds makeLabelWithProvidedControl() as a helper to create labeled controls with a optional suffix
> - code: removes dropdownDidChange() & textDidEndEditing() -> combined in controlDidEndEditing()
> - code: adds controlDidEndEditing() as an universal .action for NSControl elements
>   - adds a restore of values in UI and Preferences if there is an issue while saving a Preference
> - code: adds another little code clarity improvements

lwouis:
Hey @gingerr! Thanks for the work, it looks so much better than the current UI!

* "Max window size": i think that label is really not clear. Something should be done here to clarify what it does
* Sliders having to numbers bellow them: i'm so surprised that adding label is not a flag on the slider object or some `addTick` method.
* i can't wait to see how you added padding around the `nsgridview`. I was struggling so much for some reason. I think Cocoa design doesn't sit well in my mind.
* i'll try to find time to review the code and test locally tonight. Should i merge if it's ok, or do you plan on iterating one last time on this PR for the labels for instance, and i should wait for that?

gingerr:
@lwouis Thanks!

Naming: Yeah, this and the others can also get greatly improved once we got tooltips or something as you said in the ticket goals.

Padding: I spent hours and hours trying to understand positioning with different views and nested views and still am not sure if I got it right. The naming, the methods and the system is not obvious at all. I had trouble to control the first row or following rows padding and positioning (auto layout was messing with me). I’ve changed all views to be nested Stackviews (horizontal and vertical) since in the current layout grids are not needed but we could change it back to grids anytime. I wanted each row to implement its layout itself, to give more power to them. The view is now a vertical stack.

Code review and This pull requests: it’s alright, leave this waiting until you find time, you don’t have to work on this everyday 🍻. I would like to implement the values visualization for the sliders somehow. But you can merge this at anytime you like. If this is merged when in finish the next iteration I will open another branch and pull request. 😅

Mariusz

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
-    var fontHeight: NSTextView?
-    var interItemPadding: NSTextView?
-    var tabKeyCode: NSTextView?
-    var windowDisplayDelay: NSTextView?
+class PreferencesPanel: NSPanel {
+    // ui: base layout
+    let panelWidth = CGFloat(400)
+    let panelHeight = CGFloat(400) // gets auto adjusted to content height
+    let panelPadding = CGFloat(40)
+    let panelWidthToLabelRatio = CGFloat(0.5)
```
lwouis: Not the clearest concept 😅Does this mean 50% of the width of the panel is reserved for labels?

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
-    var inputsMap = [NSTextView: String]()
+
+    var invisibleTextField: NSTextField? // default firstResponder and used for triggering of focus loose
 
     override init(contentRect: NSRect, styleMask style: StyleMask, backing backingStoreType: BackingStoreType, defer flag: Bool) {
-        super.init(contentRect: contentRect, styleMask: style, backing: backingStoreType, defer: flag)
+        let initialRect = NSRect(x: 0, y: 0, width: panelWidth, height: panelHeight)
+        super.init(contentRect: initialRect, styleMask: style, backing: backingStoreType, defer: flag)
         title = Application.name + " Preferences"
+        titlebarAppearsTransparent = true
```
lwouis: What's the motivation for that change? I was curious and check Apple first-party apps to see if it's the new trend. I found these preferences panels:

**Notes:**
image

**Voice Memos:**
image

Calendar:
image

**Safari:**
image

**Terminal:**
image

So not very conclusive. I would say most of the apps on my system are non-Apple, and Apple themselves seem to care less and less about consistency, so to be honest I don't mind this change or the previous version. I was just curious to hear your rationale for changing it

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
         hidesOnDeactivate = false
-        contentView = makeGridView(makeLabelsAndInputs(), makeWarningLabel())
+        contentView = makeContentView()
+
+        // setup hidden element
+        invisibleTextField = NSTextField()
+        invisibleTextField?.isHidden = true
+        contentView?.subviews.append(invisibleTextField!)
+        self.initialFirstResponder = invisibleTextField
```
lwouis: The inline comment is a tell that this would be made clearer be being in a sub-method, the same way right above there is a `makeContentView` method and not a block of instructions with an inline comment `// make content view` ;p

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
+        contentView.addSubview(wrappingView)
+
+        // visual setup
+        wrappingView.orientation = .vertical
+        wrappingView.alignment = .left
+        wrappingView.spacing = panelPadding * 0.3
+        wrappingView.topAnchor.constraint(equalTo: contentView.topAnchor, constant: panelPadding * 0.5).isActive = true
+        wrappingView.bottomAnchor.constraint(equalTo: contentView.bottomAnchor, constant: panelPadding * -0.5).isActive = true
+        wrappingView.leftAnchor.constraint(equalTo: contentView.leftAnchor, constant: panelPadding * 0.5).isActive = true
+        wrappingView.rightAnchor.constraint(equalTo: contentView.rightAnchor, constant: panelPadding * -0.5).isActive = true
```
lwouis: I started having way too many of these lines in the `Application` class. Then I extracted them in sub-methods to make it more readable at a high level. Then I realize that the people who made this UI framework are really thinking OOP all the way, and you are expected to have subclasses to customize your visual elements.

Also you can make the parallel with Angular/React and think that this `wrappingView`/`NSStackView` may be reuse in some other place one day, so storing it in its own file in the `ui` folder kind of make sense.

Once I realized the framework designers intent, I extracted a bunch of files/classes in the `ui` folder. I didn't do it for every single 2 liner UI element though. I think there is a balance between having too many files in that folder, and having too many blocks like this one here in bigger classes.

I think that one should get its own class

Note that I think I will do it when I have some time for `makeBackgroundView` and `makeCollectionView` in `ThumbnailsPanel`. They are similar candidates to should get their own subclasses I think

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
+
+        return makeLabelWithProvidedControl(labelText, rawName, slider, suffixText)
+    }
+
+    private func makeLabelWithProvidedControl(_ labelText: String, _ rawName: String, _ control: NSControl, _ suffixText: String? = nil) -> NSStackView {
+        let label = NSTextField(wrappingLabelWithString: labelText + ": ")
+        label.alignment = .right
+        label.widthAnchor.constraint(equalToConstant: calcLabelWidth()).isActive = true
+
+        control.identifier = NSUserInterfaceItemIdentifier(rawName)
```
lwouis: Aaah, I knew there was a way to embed some ID in these controls so that later they can be identified in the callback... I remember going through the different attributes without finding one that was a `String`. I see you store it in `NSUserInterfaceItemIdentifier` then later grab the `rawValue`. Very happy to see the horrible `inputsMap` go away 🙏

Now that you found how to ID the inputs, you should remove all attributes of this class! You can remove everything from line 10 to 19, as well as simplify all methods on lines 57 to 67. These members only existed because I needed to check which input changed in the callbacks but now it's useless. The parent view holds a reference so the object can be a local variable here that goes out of scope without having the object deallocated.

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
+    @objc func controlDidEndEditing(senderControl: NSControl) {
+        self.makeFirstResponder(invisibleTextField) // deselects any possibly selected NSTextField (so slider & popUp changes deselect them)
+
+        let key: String? = senderControl.identifier?.rawValue
+        var newValue: String?
+
+        if senderControl is NSPopUpButton {
+            newValue = (senderControl as! NSPopUpButton).titleOfSelectedItem!
+        } else if senderControl is NSSlider {
+            newValue = String(format: "%.0f", Double(senderControl.stringValue)!) // we are only interested in decimals of the provided double
```
lwouis: That part is confusing. Is it some kind of trick where the value is like `0.80` and you want to extract `80`. I hope you can either clarify what the trick is about and why it's necessary, and simply do without this when you iterate on the sliders to give them text marks

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
-        if let textView = notification.object as? NSTextView {
-            let key = inputsMap[textView]!
+    /*
+    usage notes:
+    - NSSlider: supports on purpose currently only decimal values
+    */
+    @objc func controlDidEndEditing(senderControl: NSControl) {
+        self.makeFirstResponder(invisibleTextField) // deselects any possibly selected NSTextField (so slider & popUp changes deselect them)
+
+        let key: String? = senderControl.identifier?.rawValue
```
lwouis: You always set `identifier` on line 146 so you can use use `let key = senderControl.identifier!.rawValue` here

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
+            newValue = (senderControl as! NSPopUpButton).titleOfSelectedItem!
+        } else if senderControl is NSSlider {
+            newValue = String(format: "%.0f", Double(senderControl.stringValue)!) // we are only interested in decimals of the provided double
+        } else {
+            newValue = senderControl.stringValue
+        }
+
+//        debugPrint("PreferencesPanel: save: change", key!, newValue!)
+
+        if key != nil && newValue != nil {
```
lwouis: Like I said above, `key` is never `nil`. For `newValue`, I think it's also impossible to get `nil` from an input field, a dropdown, or a slider, given the code above. It think this whole check can be removed

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
+                }
             }
+        } else {
+            debugPrint("PreferencesPanel: save: error: key||newValue = nil", key!, newValue!)
         }
     }
+
+    private func calcLabelWidth() -> CGFloat {
+        return (panelWidth - panelPadding) * panelWidthToLabelRatio
+    }
```
lwouis: These 3 values are static I think, so you can have this derived value as an attribute in the class instead of a dynamic method. You can either have it after the 3 attributes as a simple attribute, or use swift computed attribute language feature.

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
+            newValue = String(format: "%.0f", Double(senderControl.stringValue)!) // we are only interested in decimals of the provided double
+        } else {
+            newValue = senderControl.stringValue
+        }
+
+//        debugPrint("PreferencesPanel: save: change", key!, newValue!)
+
+        if key != nil && newValue != nil {
+            if newValue == Preferences.rawValues[key!] {
+                debugPrint("PreferencesPanel: save: abort: value was not changed")
```
lwouis: I think there is a bug here. I keep seeing "PreferencesPanel: save: abort: value was not changed" in the logs when I use the max width slider and move to any position

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: I wanted to have this derived convenience value somewhere and initially had it setup as a class variable which then I had to set up in the init() of the class and it felt wrong for such a simple thing. Thats why I switched to a helper method for that value. Thanks for the hint about that language feature! 😎

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
+                return
+            }
+
+            let previousValue = Preferences.rawValues[key!]!
```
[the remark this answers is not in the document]
gingerr: Gotcha, missed that one.

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: Yes, currently its that much. This is used for the derived value to make the labels reserve the same amount of horizontal space. 

At the moment the label wording has to be long to get the message across. We can shorten them once we get tooltips done to explain settings in more details. This will allow us to reserve less space for the labels.

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: My bad, I didn't check apple apps but was thinking they were using transparent title bars (recalled that one incorrectly). I think it looks better, a more modern and flat look. I don't also mind to change it back so we are in line with apple native look. You decide.

In my opinion the titles which include icons for sub-sections of the Preferences you have shown do look kinda dated. 😮 Which of course is understandable since it are just the Preferences of an application so I get the priorities here.

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: You are absolutely right. I was in my 'think about all possible errors' mode and went overboard.

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: Gotcha. I think I was writing defensive code there which is not needed nor useful in this scope.

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: NSSlider works internally with Doubles. So if the user selects a value it could be either 78.875 or 80 (depending on if NSSlider.allowsTickMarkValuesOnly is set). Our Preferences implementation expects decimals (since the internalisation thing with 0.8 and 0,8 I reported in #51 and you changed the maxScreenUsage from 0.8 to be 80). This code snippet was one of a few proposals on stackoverflow to get a decimal from a double value in Swift. Thats the one I picked.

The NSSlider Text Marks will probably be just some representation of the internal Double value and not be used as its actual value.

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
-                debugPrint(key, error)
-                textView.string = Preferences.rawValues[key]!
+                debugPrint("PreferencesPanel: save: error", key!, error, "previousValue", previousValue, " | newValue", newValue!)
+
+                // restores the previous value in Preferences and senderControl
+                try! Preferences.updateAndValidateFromString(key!, previousValue)
+
+                if senderControl is NSPopUpButton {
+                    (senderControl as! NSPopUpButton).selectItem(withTitle: previousValue)
+                } else {
```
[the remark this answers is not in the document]
gingerr: This handles a runtime use case: to reset the corresponding `Preferences` memory value and `PreferencesPanel` view value to its `previousValue` when there was an issue with storing `newValue` into the Preferences (saveOnDisk). 

Since we currently do not inform the user about such things (notification/alert/modal missing) I think it made sense to recover from that not only in the Preferences memory values but also visually. The user will see there something strange going on and not expect his `newValue` to be actually set. In my opinion we need something like this. You could simulate this use-case by making the .json file to be read-only (some disk or permissions issue). 

Once we implement #73 we will have a way to show alerts of some kind which could then be used here as well to show a message for the user in addition to the revert of visual & memory values.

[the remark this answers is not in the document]
gingerr: You were previously changing the views value in a catch block. I was thinking about your motivation behind that code line 👍 and interpreted that use case.

```
            do {
                try Preferences.updateAndValidateFromString(key, textView.string)
                try Preferences.saveRawToDisk()
            } catch {
                debugPrint(key, error)
                textView.string = Preferences.rawValues[key]!
            }
```

gingerr:
@lwouis thank you for your in-depth review, appreciate it! Ive made a few of the requested or proposed changes and will continue to work on that tomorrow evening. Nothing new is committed yet.

> I think I would like to merge it only after you've found a way to add text marks on the sliders as otherwise I imagine it's not very user friendly.

Fully agree, gonna implement something here.

> I've had similar issues with the rest of the codebase, and thought I would share my experience, but even for me it's still very open discussion since I have no friend who's an expert at Cocoa I could ask for advice.

Thanks for that, I appreciate your feedback and our discussions. At work we mainly use Java (Frontend & Backend) and TypeScript (Frontend) so we are in the same boat (learning experience) here.

Btw. I was unhappy with the default macOS altTab experience and had on my todo list to look into alternatives but an issue was their closed source nature. I just can't install that on my company MacBook without hesitation or approval (security, privacy, trust). Then Ive read about your application in an [article](https://t3n.de/news/macos-windows-fensterwechsel-alt-tab-1209176/) which was presented one day in my google news feed (I had googled for all the alternatives 😄 weeks or months ago). Thanks for this project/application!

alt-tab-macos/ui/PreferencesPanel.swift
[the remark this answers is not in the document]
lwouis: My point is that the code you quoted from before made sense when the user is facing an input field. They can type "aaa" in the max window input field for instance, and when they do that we want to reset it to the previous value as cheap way to handle input field validation.

For dropdowns and sliders, the user can't select a wrong value anymore as we can make sure all finite values are valid at compile time. Thus this code is no longer needed for these finite inputs. Imagine the slider being moved and then that code moving it back to the previous value on release. This is a logical bug, not input validation

alt-tab-macos/ui/PreferencesPanel.swift
lwouis: I'm fine with both so i guess keep it ;p

alt-tab-macos/ui/PreferencesPanel.swift
lwouis: > Btw. I was unhappy with the default macOS altTab experience and had on my todo list to look into alternatives but an issue was their closed source nature. I just can't install that on my company MacBook without hesitation or approval (security, privacy, trust). Then Ive read about your application in an [article](https://t3n.de/news/macos-windows-fensterwechsel-alt-tab-1209176/) which was presented one day in my google news feed (I had googled for all the alternatives 😄 weeks or months ago). Thanks for this project/application!

What an uplifting story! Thanks for sharing! :)

alt-tab-macos/ui/PreferencesPanel.swift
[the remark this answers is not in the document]
gingerr: Only now I get what this was for. Good we discussed that one.

> Imagine the slider being moved and then that code moving it back to the previous value on release. This is a logical bug, not input validation

My intention was not input validation but handling an issue while saving Preferences (e.g. disk issue or whatever else could go wrong).

Ive misinterpreted your intention behind that code line and didn't expect this to be a smart way for input validation 👍 

Ok I've gonna throw an NSAlert to inform the user about an issue. Is it ok to leave the keep the input values alone then (not reverting at all)?

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,211 @@
...
+    var iconSize: NSSlider?
+    var fontHeight: NSSlider?
+    var tabKeyCode: NSTextField?
+    var windowDisplayDelay: NSTextField?
     var metaKey: NSPopUpButton?
     var theme: NSPopUpButton?
     var showOnScreen: NSPopUpButton?
-    var inputsMap = [NSTextView: String]()
+
+    var invisibleTextField: NSTextField? // default firstResponder and used for triggering of focus loose
```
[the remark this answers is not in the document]
gingerr: I could find a other way to unfocus an input and to prevent the NSPanel from focusing the first input element when its opened.

Ive tried making other elements become firstResponder but only succeeded with input elements. I was not ready to invest more time on it and went for the invisibleTextField until a better way is found (more NS... knowledge). Do you know maybe how to do this? Iam probably doing something wrong here.

The programmatic unfocus is required when:
1. use case
 - click input element
 - click a dropdown
 - result: input element is still focused and will not trigger `controlDidEndEditing` until some other input is focused
2. use case
 - click input element
 - click restart button
 - result: `controlDidEndEditing` will again be not triggered and that way the user change not saved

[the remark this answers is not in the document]
gingerr: Maybe MVC could be the solution here but I did not look into it yet. Only saw Controllers being used together with Views. I hope there is some binding and automatic changeDetection similar to Angular.

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: Oh, you are absolutely right. Didn't come into my mind to doubt the existence of this view references. 👍

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: The slider reported back way to often to `controlDidEndEditing`: mouseDown, mouseMove, mouseUp. It got the latest user selected value already with mouseMove (and saved that one) and skipped the value from the mouseUp event.

This was one of the reasons for me adding the `newValue` & `previousValue` comparison. The other was to prevent unneeded saves.

I found an interesting attribute in `NSSlider` which I disabled now: `isContinuous`. This way it only calls the `controlDidEndEditing` callback on mouseUp 💯

alt-tab-macos/ui/PreferencesPanel.swift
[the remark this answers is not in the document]
gingerr: Ok, ive figured out the NSAlert. It gives the user 3 choices: 

1. Ok (ignoring the failure, not touching the controls value)
2. Try again (repeats `controlDidEndEditing`)
3. Revert value (to `previousValue`)

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: Thank you for the choice 👍 
Made it native again cause of the sheetModal (NSAlert) otherwise looking displaced.

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: You'r welcome 🥇

alt-tab-macos/ui/PreferencesPanel.swift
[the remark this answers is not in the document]
lwouis: > input element is still focused and will not trigger controlDidEndEditing until some other input is focused

Oh, that's why? I'm surprised by this behavior. I would have to reproduce and also read online what long time mac developers have been writing about it. I'm assuming it's either a bad setup on our part, of it's a silly Cocoa design limitation that must have known work-arounds at this point

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: Gotcha. Made it non atomic, so it executes this four commands.

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: Alright. Also a sub-directory structure could begin to make sense, like feature -> layer. ThumbnailPanel could get its own sub-directory and all the related classes.

gingerr:
@lwouis Still working on this. Made good progress yesterday evening. My hope is to have it ready today evening (EU time).

gingerr:
@lwouis Alright, the next iteration for the partial implementation of #49 is rebased and force pushed. Ive done all the changes we talked about, refactored some more and kept at implementing improvements and little features into this. Every time I thought its good enough there was something else popping up (bad UX or a different approach to solve the problem at hand). The obvious UX improvements are my attempt at a simple instant validation with visualisation for the TextFields, error handling via an sheet (modal alike) and visual representation of slider values.

Tell me what you think.
Btw. thanks for the interesting links and insights at the other issues!

Cheers
Mariusz

**Changelog from commit message**

> implements requested changes from Code-Review and contains also refactoring and more features and improvements
> - Application Wide: changes all exceptions to NSError (required for error handling via sheet in panel)
> - PreferencesPanel: adds display of current value of Sliders
> - PreferencesPanel: adds validation with visualisation for TextFields (whitelist for KeyCode and ranges for maxScreenUsage & windowDisplayDelay) during editing, window open and close
> - PreferencesPanel: adds error sheet (in-window modal) at all known possible errors to display error information and ask the user how to proceed
> - PreferencesPanel: removes Restart button (and relaunch code in Application)
> - PreferencesPanel: adds Preferences being now live
> - PreferencesPanel: removes invisibleText workaround
> - PreferencesPanel: adds TextField value save on each keyDown event (if valid and changed)
> - PreferencesPanel: adds window activation/deactivation
> - PreferencesPanel: changes label texts of some Preferences

lwouis:
Hey, sorry for the delay!

Here is some feedback from an e2e perspective:

* if you invoke the preferences panel, then focus something else, then invoke it again, the panel titlebar stays unfocused/greyed
* If you invoke the preferences panel, then focus something else, then alt-tab, the panel will move to the foreground until your release alt-tab

I tested and these 2 issues pre-date your PR, so I will open a separate ticket for them.

Other things:

* Keyboard focus of the inputs is almost perfect. The dropdowns are not getting the blue halo though. Every other input can be tabbed to.
* Validation of the text input is broken. I tried to put `2` but the border becomes red without further explanation. I tried to write `a`, same, then back to `4` same still red.
image
* Max size of the app icons is a bit strange at `62px`. I think it should stop at `64px` where the downscaling is optimal. I'm also thinking that some people may want bigger values like `256px`, especially if we allow to disable the thumbnails later to have an app switcher style UI. This should be another ticket/discussion though
* I think the number of thumbnails per row and delay could also benefit from being turned into sliders. For the max thumbnails I also think having a special case "Infinite" or something like that would enable the use-case where some people may prefer always 1 row. Turning text fields to sliders would also remove the need for text field validation which is a nice win.

I will review the code now

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,263 @@
...
+                .filter({ !$0.isValid() })
+        )
+        let focusedField = invalidFields?.filter({ $0.currentEditor() != nil }).first
+        let fieldToNotify = focusedField ?? invalidFields?.first
+        fieldToNotify?.delegate?.controlTextDidChange?(Notification(name: NSControl.textDidChangeNotification, object: fieldToNotify))
+
+        if fieldToNotify != focusedField {
+            makeFirstResponder(fieldToNotify)
+        }
+    }
```
lwouis: Ok I understand now what you do. You block the user from leaving the window if a field is invalid. I think you made it was harder for yourself!

* Easiest/strongest fix: remove these annoying text fields and put sliders. Done. User can make mistakes now
* Intermediate fix: keep the previous logic to reset the field as soon as an incorrect value is written and focus on that field goes gone. This guarantees the field has a correct value always.
* Too-intense fix: the approach you took here where you want to communicate more to the user but I believe you end up confusing them more than previous 2 approaches as the red field doesn't explain what's wrong next to it, and when I close the panel, it says it's wrong but not why, or which format is expected. Also what if I don't close the window but close the app, or simply switch the window to the background and expect my change to be active? I think you're going into a territory that would require integrating a full-blown validation framework, but in our case it's way simpler to do the solutions above.

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,263 @@
...
+            // non-special keys (mac & pc keyboards): https://eastmanreference.com/complete-list-of-applescript-key-codes
+            var whitelistedKeycodes: [Int] = Array(0...53)
+            whitelistedKeycodes.append(contentsOf: [65, 67, 69, 75, 76, 78, ])
+            whitelistedKeycodes.append(contentsOf: Array(81...89))
+            whitelistedKeycodes.append(contentsOf: [91, 92, 115, 116, 117, 119, 121])
+            whitelistedKeycodes.append(contentsOf: Array(123...126))
+            return whitelistedKeycodes.contains(int)
+        }
+        let maxThumbnailsPerRowValidator: ((String)->Bool) = { (3...16).contains(Int($0) ?? -1) }
+        let windowDisplayDelayValidator: ((String)->Bool) = { (0..<10000).contains(Int($0) ?? -1) }
```
lwouis: See, you go really detailed on the validation here, but you don't surface that to the user. I think my comment above about just putting sliders is the way to go. For the tab key, I think for now let's let the user type an incorrect int for now. We are gonna replace this soon with a lib, so no need to spend time here I think

alt-tab-macos/logic/PreferencesPanelNSTextFieldEditable.swift
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    // custom method
+    func visualizeValidationState(_ isValid: Bool) -> Void {
+        if !isValid {
+            wantsLayer = true
+            layer?.borderColor = NSColor.systemRed.cgColor
+            layer?.borderWidth = 1
+        } else {
+            wantsLayer = false
```
lwouis: I think this logic here is not correct. `wantsLayer` live on some lifecycle that I don't think come into play instantly here. That means if it becomes red, then goes into the `else`, it will not clear the red. I think do the same as I did for the thumbnails border: set `wantsLayer = true` on init always, then set either `.systemRed` or `.clear` in the `if/else`

gingerr: Thank you, corrected.

alt-tab-macos/logic/PreferencesPanelNSTextFieldEditable.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    // protocol method
+    func controlTextDidChange(_ obj: Notification) {
+        visualizeValidationState(isValid())
+        let textField = obj.object as! PreferencesPanelNSTextFieldEditable
+        sendAction(textField.action, to: textField.target)
+    }
+
+    // custom method
+    func visualizeValidationState(_ isValid: Bool) -> Void {
```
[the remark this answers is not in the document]
gingerr: Thanks for that improvement catch!

alt-tab-macos/logic/PreferencesPanelNSTextFieldEditable.swift
```diff
@@ -0,0 +1,34 @@
...
+            wantsLayer = true
+            layer?.borderColor = NSColor.systemRed.cgColor
+            layer?.borderWidth = 1
+        } else {
+            wantsLayer = false
+        }
+    }
+
+    // custom method
+    func isValid() -> Bool {
```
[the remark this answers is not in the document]
gingerr: `isValid()` is used `by PreferencesPanel`. Thats why this validation and visualisation is split into two little methods.

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: You are right, the special nature of this input is just hinted `KeyCode (48 = Tab)` wording. 

This is just an improvement pull request for the related issue and I required some kind of fast and simple validation when I added that red visualisation state to the other textfields: an invalid key code would be a valid CGFloat (so the panel would not throw any error for that). This keycodes array corrects that and is an attempt on an easy fix/improvement to the current situation.

I planned to link that nice keycodes url as a 'good enough' solution via a button or clickable text (in the grey suffix textfield) in an upcoming iteration of this PreferencesPanel since its planned to integrate/implement some keycode recording in the future.

I think Sliders and PopUps for everything are also the way to go here and TextFields only where it makes absolute sense to use one. They overcomplicate this Panel and open a can of worms with validation & visualisation.

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,263 @@
...
+    override func close() {
+        NSApp.deactivate()
+        (NSApp as! Application).preferencesPanel = nil
+        super.close()
+    }
+
+    public func windowShouldClose(_ sender: NSWindow) -> Bool {
+        windowCloseRequested = true
+        challengeNextInvalidEditableTextField()
+        return attachedSheet == nil // user is challenged
```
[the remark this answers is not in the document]
gingerr: Oups 😇 Made the wording more clear.

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,263 @@
...
-    var theme: NSPopUpButton?
-    var showOnScreen: NSPopUpButton?
-    var inputsMap = [NSTextView: String]()
+class PreferencesPanel: NSPanel, NSWindowDelegate {
+    let panelWidth = CGFloat(496)
+    let panelHeight = CGFloat(256) // auto expands to content height (but does not auto shrink)
+    let panelPadding = CGFloat(40)
+    var labelWidth: CGFloat { get { // derived convenience variable
+        return (panelWidth - panelPadding) * CGFloat(0.45)
+    }}
```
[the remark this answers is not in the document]
gingerr: Gotcha. Thank you.

alt-tab-macos/logic/PreferencesPanelNSTextFieldEditable.swift
```diff
@@ -0,0 +1,34 @@
+import Cocoa
+
+class PreferencesPanelNSTextFieldEditable: NSTextField, NSTextFieldDelegate {
+
+    var validationHandler: ((String)->Bool)?
+
+    // protocol method
```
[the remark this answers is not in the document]
gingerr: Yeah it is annoying. Without such comments one has to known the inside outs of the super classes and interfaces. Removed the comments.

alt-tab-macos/logic/PreferencesPanelNSTextFieldEditable.swift
```diff
@@ -0,0 +1,34 @@
+import Cocoa
+
+class PreferencesPanelNSTextFieldEditable: NSTextField, NSTextFieldDelegate {
```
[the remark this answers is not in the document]
gingerr: Oh, yeah sorry for that.

The naming is hard one this one. Thats why I went with naming this in relation to the PreferencesPanel. In generic wording this would be something like ValidatedInstantChangeNSTextField?

alt-tab-macos/logic/Extensions.swift
```diff
@@ -39,3 +35,25 @@ extension String {
...
+// add convenience to NSError
+extension NSError {
+    class func make(domain: String, message: String, code: Int = 9999) -> NSError {
+        return NSError(
+                domain: domain,
+                code: code,
+                userInfo: [NSLocalizedDescriptionKey: message, NSLocalizedFailureReasonErrorKey: message]
+        )
+    }
+}
```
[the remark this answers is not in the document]
gingerr: The motivation here was: I required something that is an NSError or sub-classes it for the error sheet in PreferencesPanel. Why I used NSError was because the Cocoa framework is using that for its exception during the json save (disk/permissions issue) and places very useful informations that describe the error at hand and can be used to be displayed to the user. The PreferencesPanel is using that `.localizedDescription` as a text.

The earlier String extension sub-classed Error but NSError does not sub-class Error (that surprised my a little bit, I expected NSError to sub-class Error to be honest).

Re-using NSError was one simple and fast way to throw application errors, of course without going into an app-wide error/logging overhaul (out of scope this pull request).

alt-tab-macos/logic/PreferencesPanelNSTextFieldEditable.swift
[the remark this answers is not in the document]
lwouis: If there is only 1 type of TextField in the app, I think we can just call it `TextField`, no? I think it's not very ambiguous, and you can just see the call-sites if the context is unclear

alt-tab-macos/logic/Extensions.swift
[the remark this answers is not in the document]
lwouis: Oh I see, interesting distinction between `NSError` and `Error`. These frameworks sure have accreted..

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,263 @@
...
         title = Application.name + " Preferences"
         hidesOnDeactivate = false
-        contentView = makeGridView(makeLabelsAndInputs(), makeWarningLabel())
+        contentView = makeContentView()
+        NSApp.activate(ignoringOtherApps: true)
     }
 
-    private func makeLabelsAndInputs() -> [[NSView]] {
+    override func close() {
+        NSApp.deactivate()
```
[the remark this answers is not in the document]
gingerr: I realised that special needs of ThumbnailsPanel when going for NSApp.activate and .deactivate here (after searching how to focus windows). The first issue from #80 I didn't come up with but know how to fix (thanks for that use case!), the second one from #80 requires some investigation with try/error and is not affected by this pull request.

I will remove the two lines so this pull request does not thinker with that behaviour. Thank you for the separate issue.

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -1,110 +1,263 @@
...
+        (NSApp as! Application).preferencesPanel = nil
+        super.close()
+    }
```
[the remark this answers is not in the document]
gingerr: The easiest (less code) way to let the panel display the actual values from Preferences instead of whatever the user has entered the last time he had the window open. This was a problem for the use-case, when the Preferences rejected the provided value (e.g. CGFloat exception) the PreferencesPanel was still showing the user input (which was rejected) so the the user would think his choice has been set correctly. 

I had many iterations with and without validation and this close (de-instantiation) and new PreferencesPanel instance creation solved this issue without requiring additional code.

alt-tab-macos/logic/PreferencesPanelNSTextFieldEditable.swift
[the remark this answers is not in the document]
gingerr: Sure. Btw. had to add an second NSTextField but named them ...Label as you proposed and this one just TextField.

alt-tab-macos/ui/PreferencesPanel.swift
gingerr: Gotcha. Removed the text fields for the other fields as mentioned in another code review comment. Tab key input is the only one remaining for the moment.

Thats focus/unfocus is my problem. I have issues to get notified when the input looses focus anyhow (panel closed, dropdown opened, another input selected) to trigger a save Preference operation. If we don't trigger the save the last somehow here the value the user entered will not be saved at all. The panel before this pull-request has the same issue. If the user reopened the panel, he saw whatever he entered, even if it was not saved or the Preferences rejected the value. The PreferencesPanel did only save if the user focused an other input. This is my attempt at fixing this without requiring the user to click something.

One of the workarounds I came up with was the invisibleTextField (to trigger an unfocus of the current input which then triggered a save). Now this invisibleTextField workaround is removed and the inputs are triggering a save on every key press. But we need to allow the inputs to be invalid while the user is editing them (or has the panel not closed yet), that means to skip execution of an save (or there will be an exception triggering the error sheet panel). 

One earlier iteration I had it that way, a user was not able to leave an input while it was not valid but that felt not right. Since there is now only one input text field remaining I could try to find out how I did it and reimplement it. 

Once we have some way of recording a key or key combination (#68) we could remove all the TextField handling/complications.

> dosn't explain what's wrong next to it, and when I close the panel, it says it's wrong but not why, or which format is expected. 

Fair enough.

> Also what if I don't close the window but close the app, or simply switch the window to the background and expect my change to be active? 

I handled that 😬  The field validates instantly and gives feedback in time. The Preference will be saved to Preferences and disk if its valid. If the user close forces the window or the application and reopens, he will see his last valid value for the said Preference.

> I think you're going into a territory that would require integrating a full-blown validation framework, but in our case it's way simpler to do the solutions above.

Yes, once the textfields are gone completely this will be a happy little code to look at.

gingerr:
@lwouis 

> Hey, sorry for the delay!

No problem at all! Thanks for reviewing the pull-request in detail once again!

> Here is some feedback from an e2e perspective:
> * Keyboard focus of the inputs is almost perfect. The dropdowns are not getting the blue halo though. Every other input can be tabbed to.

Yeah, coming from web development I was wondering about that one too but it looks like its the normal native behaviour so I stopped there: 

Screen Shot 2019-11-10 at 20 14 41

> * Validation of the text input is broken. I tried to put `2` but the border becomes red without further explanation. I tried to write `a`, same, then back to `4` same still red.
>   image

I can't reproduce that. If I repeat that steps the 4 will be valid for me. Anyway, this Preference is a slider now, as you suggested. I was wondering the same and had on my todo list to ask you, which preferences should become sliders (all of them?) and which min/max values you would choose for them.

> * Max size of the app icons is a bit strange at `62px`. I think it should stop at `64px` where the downscaling is optimal. I'm also thinking that some people may want bigger values like `256px`, especially if we allow to disable the thumbnails later to have an app switcher style UI. This should be another ticket/discussion though

Corrected. Had a issue with tickmarks here: having too much or to little of them and the slider not starting at 0 for this Preference and me picking the wrong tickmarks divider. I made the all the sliders without tickmarks now. Another divider for the amount of tickmarks would be also possible. My motivation for the tickmarks was, to prevent the hammering of Preferences.saveToDisk() since its get called pretty often with continuous sliders. I had them non-continuous before, but the current slider value visualisation implementation requires them to be continuous again. A debounce (as in rxjs observables) would be a better solution here I think.

256 sounds also interesting! I think we should also allow 0 for `Apps icon size` and `Window font size` for users that would like to just see the screenshots or application icons.  

Screen Shot 2019-11-10 at 20 13 02


> * I think the number of thumbnails per row and delay could also benefit from being turned into sliders. For the max thumbnails I also think having a special case "Infinite" or something like that would enable the use-case where some people may prefer always 1 row. Turning text fields to sliders would also remove the need for text field validation which is a nice win.

Done. The infinite value would require a different non-llinear scale (however this is done with NSSlider) or could be simply solved by an additional checkbox that disabled the slider?

> 
> I will review the code now
> 
> **Update:** I finished reviewing. I asked for change instead of just merging and putting the comments as informational-only. The main reasons are:
> 
> 2. I think the validation logic you added adds a lot of code to the project, and could be approached in a different manner (see my other comments for suggestions)

Thanks for the detailed review! Yes it does, sorry for that. I think my unfamiliarly with Cocoa does also add to that. 

Ive done the requested changes and additional little improvements, see other comments.

lwouis:
I tested the PR locally and I think it is in great shape! Thanks you for all the work on this! By the way if you have time to hang out in the Discord, I would like to discuss next tickets :)

gingerr:
Nice! Thanks for the merge 🤝 and making my adventure into macOS ui lands possible.
See you on Discord.

Issue #14, opened 2019-08-30
Display windows from other Spaces


lwouis:
Currently windows from other spaces, including fullscreen'd windows that are in their own space, are not displayed in the thumbnails

coxjc:
i'd like to bump this. i generally keep all my apps in full-screen, so this isn't uber useful until this enhancement is done.

lwouis:
I checked and HyperSwitch has this feature. You can see at the bottom of the preferences UI here:

![image](https://user-images.githubusercontent.com/106195/67002862-bbdee680-f117-11e9-97c3-b798bb27e52d.png)

However I'm not sure how they managed to implement that. The API call used currently is `CGWindowListCopyWindowInfo` which doesn't seem to have an option for querying all spaces. I'll try to find a way to implement it. If anyone has an idea, please share it here :)

lwouis:
I tested:

* HyperSwitch: always displays from other spaces correctly, with the Space number on the thumbnail
* WindowSwitcher: only displays if it was running during the space transition. It seems they maintain state in the app instead of asking the OS. If you move windows to a space and restart, the app doesn't show the windows.

okoeth:
Would really like to see this feature. Use Spaces a lot...

lwouis:
[Interesting discussion on the Rectangle project](https://github.com/rxhanson/Rectangle/issues/16) on why they don't support Spaces in general because there don't have public APIs.

I've played today with private APIs from [CGSPrivate.h](https://raw.githubusercontent.com/lswank/Tranquility/master/CGSPrivate.h) and it seems that even these are now buggy. [This tweet](https://twitter.com/puffnfresh/status/255518851452764160?lang=en) says that they stopped returning correct values in 10.8.

I'm trying to find a lead on which tech to use to implement this feature, but it's very hard to find literature on this. The big projects like [yabai](https://github.com/koekeishiya/yabai), but look at the crazyness they have to go through:

> It uses a scripting-addition, which is a bundle of code that we inject into Dock.app to elevate our privileges when communicating with the WindowServer. The WindowServer is a single point of contact for all applications. It is central to the implementation of the GUI frameworks and many other services. Because of this, System Integrity Protection must be disabled for yabai to function properly.

lwouis:
Interesting potential framework to switch spaces: https://github.com/bigbearlabs/SpaceSwitcher

lwouis:
[...]

PR #78, opened 2019-11-06
fix: show minimized windows (closes #11 closes #45 closes #62) 

### LINKED ISSUE #11: Handle minimized windows
It is not possible to display thumbnails for minimized windows, but we could display a message instead.

Listing the minimized windows would be done by removing `.optionOnScreenOnly` in `CGWindowListCopyWindowInfo([.excludeDesktopElements], kCGNullWindowID)`

### LINKED ISSUE #45: Recomputing Thumbnails is slow and expensive
First of all, thank you for this great app, I really like it!

However, on my 2016 MacBook Pro, getting the UI to show up feels very sluggish, even though I set the delay to 0ms. After investigating the code a bit and running some tests, it seems like recomputing the thumbnails every time the UI shows up is quite expensive.

In my testing I had around 10 windows open and the following lines of code resulted in around 200ms of time: https://github.com/lwouis/alt-tab-macos/blob/7ff17b5882f07d17f66d9019c78449cf9cf526a6/alt-tab-macos/ui/Application.swift#L151-L154
Most of that time was spent in the thumbnail function of `OpenWindow`. A few suggestions on how this could be fixed:

- Use multithreading for capturing the thumbnails.
- Capture thumbnails on background thread, then when done update the UI. This would really help getting the UI to show instantly. However, to be able to show something for each window initially I would also:
- Cache the previously captured thumbnails.
- Figure out how Mission Control does it and replicate that. If it's similar to iOS multitasking, it's probably taking spontaneous snapshots and storing them somewhere on disk. If we could figure out where, then those could be used.
- Have a continuously running background task that captures thumbnails every few seconds / minutes and caches them. (Maybe even checks if anything changed and if not increase the time for that specific window?)

### LINKED ISSUE #62: Switch to windows faster
Even if we ignore potential performance issues from #45, after the shortcut is released, the app is slow to switch focus to the target window.

In my tests (using a slow-mo camera), I see HyperSwitch has 3 frames between its UI disappearing, and the target window appearing. For AltTab, it's 10 frames. I'm assuming they use another way to request focus to the OS

Here is the current code (10ms):

```swift
app.activate(options: [.activateIgnoringOtherApps])
AXUIElementPerformAction(target!, kAXRaiseAction as CFString)
```

lwouis:
I'm opening a PR to share my progress on #11. Currently it works, but minimized images have no thumbnail. I included some commented code that use `screencapture` to grab thumbnails for minimized windows. The performance is bad, so we can't use this approach at the moment


gingerr:
Tried to compile on macOS High Sierra (10.13.6):

Screen Shot 2019-11-11 at 20 53 46

Looks like the references are missing in my files on that old macOS version. I am using Hackintosh but the macOS system files are not modified in any way. [Clover](https://sourceforge.net/projects/cloverefiboot/) does something before the unmodified macOS is booted. 

Ive changed naively (try&error) without any understanding of the code 
- `Int.zero` to `0`
- `CGWindowID.zero` to `CGWindowID(0)`
- `pid_t.zero` to `Int32(0)`

And it compiles and runs successfully. Does my change have the same meaning as the previous code?

Usage observations on single monitor on said macOS version (10.13.6):
- I get screenshots for minimised (cmd+m) and hidden (cmd+h) windows
- there is an (timing?) issue when an window is being minimised (animation in progress) or very shortly after its been minimised and the ThumbnailPanel is summoned
  - sometimes there is no screenshot: I think its when the minimise animation is in progress
  - sometimes the window is missing completely in the ThumbnailPanel
    - when I escape the ThumbnailPanel summon in this case and summon it again the missing window is shown
    - I think that happens when I try to summon the ThumbnailPanel just when the animation has finished
  - when I wait a little (half/full second) before summoning the ThumbnailPanel, the minimised application is always shown
- no feeling of lag with just a few windows open and something between half/full second with 20 windows open on desktop hardware (i7-7700k & Nvidia 1080 ti)

**Screenshots**
Ive minimised Discord.

Discord without screenshot
discord without screenshot

Discord missing
discord missing

Discord reappears on 2nd ThumbnailPanel summon
discord reappears

Excited about this one, it would be great for productivity and desktop management.

gingerr:
Tested now also on my MacBook Pro 2015 (current macOS and dual monitor setup: integrated & external). Compiles and runs with the checked out code. Performance is good. Same observation regarding the "timing" issue. Can you reproduce that?

alt-tab-macos/api-wrappers/PrivateApis.swift
```diff
@@ -0,0 +1,11 @@
+import Cocoa
+import Foundation
+
+class PrivateApis {
+    // CGS can capture screenshots of minimized windows, which CGWindowListCreateImage can't
+    static func windowScreenshot(_ windowId: inout UInt32) -> CGImage {
```
gcbw: maybe rename to `minimizedWindowScreenshot` to make the purpose clear without the need for the comment?

lwouis: This method can capture screenshots of any window, not just minimized. In this PR i only use this method for all windows as a matter of fact

lwouis:
Closing this PR as I'm working on an all-inclusive PR. See https://github.com/lwouis/alt-tab-macos/issues/14#issuecomment-569056863

PR #83, opened 2019-11-11
fix: code compile compatibility with old macos

gingerr:
- .zero in Int/Int32/UInt32 is missing here

Screen Shot 2019-11-11 at 20 53 46

This is just a pull request to show the required changes to make it compile again on a High Sierra macOS.


lwouis:
I can already imagine in the future seeing `Int(0)` and thinking "hey I can use the global const Int.zero here to optimize", and not realizing that they are breaking builds on High Sierra.

There must be a way that I can setup the xcode project so that it gives me the same warning as you get on High Sierra. That way there is no way to break build on older versions

lwouis:
> * Supporting 1 SDK means people stuck with an another XCode version need to download and point to this other SDK. I'm not sure if this actually works. I'm gonna try it out.

image

image

It seems the OS really locks everything :/

gingerr:
I like that about Apple, most devices get updates for many years to come (compared to android on third party devices), windows on desktops is of course best here. 

Thanks for this merge.

PR #91, opened 2019-12-27
feat: display other spaces/minimized windows (closes #14)

### LINKED ISSUE #11: Handle minimized windows

### LINKED ISSUE #14: Display windows from other Spaces

### LINKED ISSUE #45: Recomputing Thumbnails is slow and expensive

### LINKED ISSUE #62: Switch to windows faster

lwouis:
Also closes #11 closes #45 closes #62

BREAKING CHANGE: this brings huge changes to core parts of the codebase. It introduces the use of private APIs that hopefully are should be compatible from macOS 10.12+, but I couldn't test them. I reviewed the whole codebase to clean and improve on performance and readability


gcbw:
tested on 10.14 OK

but showing windows from all workspaces with not option to disable completely breaks this for me :(

Issue #39, opened 2019-10-17
Update the thumbnails live when windows get closed or new windows are opened


lwouis:
Right now the UI shows a snapshot of windows at the time the shortcut is pressed. This has pros: the UI is stable and not changing on the user, but also cons: the UI doesn't reflect the current state of windows.

It's a tradeoff between stability and change. I would love feedback from other people about this. Making it should be a preference?

mfn:
HyperSwitch shows the ones "currently cached"; if you've the thumbnails open and wait with your window choose action, it will update them with a delay.

Which it was always up2date like on Windows, but feels like a good compromise to me.

lwouis:
This was closed by #114

PR #94, opened 2019-12-27
WIP: Add quick-actions

### LINKED ISSUE #9: Add quick-actions on thumbnails
Add quick actions on thumbnails, similar to what WindowSwitcher is doing:

image

I think only these are actually useful:

* Resize half-screen left/right/up/down
* Resize full-screen
* Quit

In a first phase, keyboard only, then maybe some visual buttons can be added.

More details in [this video](https://www.youtube.com/watch?v=MdPuoe6oY1A), especially the minimize behaviour which shows limitation of the Mac API (can't get thumbnail of a minimized window)

Robinhuett:
This PR adds quick-actions to quickly manipulate the selected window/application.

~Current limitation aside from the progress below is, that windows in fullscreen can't be closed.
Ideas on how to achieve this are more than welcome~ Works now :)
~Also im not entirely happy with the delays in `./ui/Application.swift` _(LOC 110 & 119)_. Input regarding this is also appreciated.~ 

@lwouis: Could you take a look at it, if this is going in the direction you intended? :)

Progress:
- [x] Close windows (`w`)
- [x] Quit application (`q`)
- [ ] Snap window left (`arrow left`)
- [ ] Snap window right (`arrow right`)
- [ ] Enter full screen (`f`)
- [ ] Minimize Window (`m`)

Closes #9

### DIFF
--- a/alt-tab-macos/api-wrappers/AXUIElement.swift
+++ b/alt-tab-macos/api-wrappers/AXUIElement.swift
@@ -7,7 +7,9 @@ import Foundation
 enum AXAttributeKey: String {
     case windows = "AXWindows"
     case minimized = "AXMinimized"
+    case fullScreen = "AXFullScreen"
     case focusedWindow = "AXFocusedWindow"
+    case closeButton = "AXCloseButton"
     case subrole = "AXSubrole"
 }
 
@@ -53,6 +55,10 @@ extension AXUIElement {
         return attribute(.minimized, Bool.self) == true
     }
 
+    func isFullScreen() -> Bool {
+        return attribute(.fullScreen, Bool.self) == true
+    }
+
     func focus(_ id: CGWindowID) {
         var elementConnection = UInt32(0)
         CGSGetWindowOwner(cgsMainConnectionId, id, &elementConnection)
@@ -85,5 +91,15 @@ extension AXUIElement {
 
         SLPSPostEventRecordTo(&psn_, &(UnsafeMutablePointer(mutating: UnsafePointer<UInt8>(bytes1)).pointee))
         SLPSPostEventRecordTo(&psn_, &(UnsafeMutablePointer(mutating: UnsafePointer<UInt8>(bytes2)).pointee))
+    }
+
+    func close() {
+        if isFullScreen() {
+            AXUIElementSetAttributeValue(self, AXAttributeKey.fullScreen.rawValue as CFString, 0 as CFTypeRef)
+            return
+        }
+        if let closeButtonRef = attribute(.closeButton, AXUIElement.self) {
+            AXUIElementPerformAction(closeButtonRef, kAXPressAction as CFString)
+        }
     }
 }
--- a/alt-tab-macos/logic/Keyboard.swift
+++ b/alt-tab-macos/logic/Keyboard.swift
@@ -47,6 +47,8 @@ func keyboardHandler(proxy: CGEventTapProxy, type: CGEventType, event_: CGEvent,
             let isMetaDown = event.modifierFlags.contains(Preferences.metaModifierFlag!)
             let isRightArrow = event.keyCode == kVK_RightArrow
             let isLeftArrow = event.keyCode == kVK_LeftArrow
+            let isQ = event.keyCode == kVK_ANSI_Q
+            let isW = event.keyCode == kVK_ANSI_W
             let isEscape = event.keyCode == kVK_Escape
             if isMetaDown && type == .keyDown {
                 if isTab && event.modifierFlags.contains(.shift) {
@@ -57,6 +59,10 @@ func keyboardHandler(proxy: CGEventTapProxy, type: CGEventType, event_: CGEvent,
                     return dispatchWork(application, true, { application.cycleSelection(1) })
                 } else if isLeftArrow && application.appIsBeingUsed {
                     return dispatchWork(application, true, { application.cycleSelection(-1) })
+                } else if isQ && application.appIsBeingUsed {
+                    return dispatchWork(application, true, { application.quitTargetApp() })
+                } else if isW && application.appIsBeingUsed {
+                    return dispatchWork(application, true, { application.closeTarget() })
                 } else if type == .keyDown && isEscape {
                     return dispatchWork(application, false, { application.hideUi() })
                 }
--- a/alt-tab-macos/logic/TrackedWindow.swift
+++ b/alt-tab-macos/logic/TrackedWindow.swift
@@ -9,7 +9,18 @@ class TrackedWindow {
     var thumbnail: NSImage?
     var icon: NSImage?
     var app: NSRunningApplication?
-    var axWindow: AXUIElement?
+    private var _axWindow: AXUIElement? = nil
+    var axWindow: AXUIElement? {
+        set {
+            _axWindow = newValue
+        }
+        get {
+            if _axWindow == nil {
+                _axWindow = id.AXUIElementOfOtherSpaceWindow(ownerPid)
+            }
+            return _axWindow
+        }
+    }
     var isMinimized: Bool
     var spaceId: CGSSpaceID?
     var spaceIndex: SpaceIndex?
@@ -28,7 +39,7 @@ class TrackedWindow {
         if let cgImage = cgId.screenshot() {
             self.thumbnail = NSImage(cgImage: cgImage, size: NSSize(width: cgImage.width, height: cgImage.height))
         }
-        self.axWindow = axWindow
+        self._axWindow = axWindow
         self.isMinimized = isMinimized
         self.spaceId = spaceId
         // System Preferences windows appear on all spaces, so we make them the current space
@@ -37,9 +48,16 @@ class TrackedWindow {
     }
 
     func focus() {
-        if axWindow == nil {
-            axWindow = id.AXUIElementOfOtherSpaceWindow(ownerPid)
-        }
         axWindow?.focus(id)
+    }
+
+    func close() {
+        axWindow?.close()
+    }
+
+    func quitApp() {
+        if app != nil {
+            app?.terminate()
+        }
     }
 }
--- a/alt-tab-macos/ui/Application.swift
+++ b/alt-tab-macos/ui/Application.swift
@@ -5,11 +5,13 @@ let cgsMainConnectionId = CGSMainConnectionID()
 
 class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
     static let name = "AltTab"
+    let observer = Observer()
     var statusItem: NSStatusItem?
     var thumbnailsPanel: ThumbnailsPanel?
     var preferencesPanel: PreferencesPanel?
     var uiWorkShouldBeDone = true
     var isFirstSummon = true
+    var isOutdated = false
     var appIsBeingUsed = false
 
     override init() {
@@ -44,6 +46,7 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
     func hideUi() {
         debugPrint("hideUi")
         thumbnailsPanel!.orderOut(nil)
+        observer.clearObservers()
         appIsBeingUsed = false
         isFirstSummon = true
     }
@@ -53,6 +56,20 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
         if appIsBeingUsed {
             debugPrint("focusTarget: appIsBeingUsed")
             focusSelectedWindow(TrackedWindows.focusedWindow())
+        }
+    }
+
+    func closeTarget() {
+        if appIsBeingUsed {
+            debugPrint("closeTarget")
+            closeSelectedWindow(TrackedWindows.focusedWindow())
+        }
+    }
+
+    func quitTargetApp() {
+        if appIsBeingUsed {
+            debugPrint("quitTargetApp")
+            quitApplicationOfSelectedWindow(TrackedWindows.focusedWindow())
         }
     }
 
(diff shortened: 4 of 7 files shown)


alt-tab-macos/logic/Keyboard.swift
```diff
@@ -57,10 +59,14 @@ func keyboardHandler(proxy: CGEventTapProxy, type: CGEventType, event_: CGEvent,
...
                     return dispatchWork(application, true, { application.cycleSelection(-1) })
+                } else if isQ && application.appIsBeingUsed {
+                    return dispatchWork(application, true, { application.quitTargetApp() })
+                } else if isW && application.appIsBeingUsed {
+                    return dispatchWork(application, true, { application.closeTarget() })
                 } else if type == .keyDown && isEscape {
                     return dispatchWork(application, false, { application.hideUi() })
                 }
-            } else if isMetaChanged && !isMetaDown {
+            } else if isMetaChanged && !isMetaDown && application.appIsBeingUsed {
```
lwouis: I'm confused as to why you added this condition? Did you run into a corner-case that required this?

alt-tab-macos/logic/TrackedWindow.swift
```diff
@@ -42,4 +42,17 @@ class TrackedWindow {
...
+        if axWindow == nil {
+            axWindow = id.AXUIElementOfOtherSpaceWindow(ownerPid)
+        }
+        axWindow?.close()
+    }
+
+    func quitApp() {
+        if app != nil {
+            app!.terminate()
+        }
```
lwouis: I would suggest a simpler `app?.terminate()` here

alt-tab-macos/logic/TrackedWindow.swift
```diff
@@ -42,4 +42,17 @@ class TrackedWindow {
         }
         axWindow?.focus(id)
     }
+
+    func close() {
+        if axWindow == nil {
+            axWindow = id.AXUIElementOfOtherSpaceWindow(ownerPid)
+        }
+        axWindow?.close()
+    }
```
lwouis: What about adding the following logic in a getter method to avoid duplication?

```swift
if axWindow == nil {
    axWindow = id.AXUIElementOfOtherSpaceWindow(ownerPid)
}
```

alt-tab-macos/ui/Application.swift
```diff
@@ -93,4 +102,22 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
         hideUi()
         DispatchQueue.global(qos: .userInteractive).async { window?.focus() }
     }
+
+    func closeSelectedWindow(_ window: TrackedWindow?) {
+        DispatchQueue.global(qos: .userInteractive).async { window?.close() }
+        isOutdated = true
+        // slight delay needed to give the window time to close before updating the thumbnails
+        DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + .milliseconds(75), execute: {
```
lwouis: I'm afraid this is too fragile. Many windows will take longer to close so it will fail. I think there are way to listen to windows being closed and application being quit. That would be one way to follow the OS work completion.

Another way would be to optimistically update our UI to remove the thumbnail(s), and consider that the OS doing the work is an async task now that will happen eventually (which it should). We could either remove the thumbnails, or mark them somehow like opacity reduced, black, red cross, etc.

I think these are the 2 ways to go: (a) listen to window close events somehow, or (b) update UI optimistically and let the OS deal with execution of the order

alt-tab-macos/ui/Application.swift
```diff
@@ -93,4 +102,22 @@ class Application: NSApplication, NSApplicationDelegate, NSWindowDelegate {
...
+            self.showUiOrCycleSelection(TrackedWindows.focusedWindowIndex)
+        })
+    }
+
+    func quitApplicationOfSelectedWindow(_ window: TrackedWindow?) {
+        DispatchQueue.global(qos: .userInteractive).async { window?.quitApp() }
+        isOutdated = true
+        // slightly more delay needed to give the application time to quit before updating the thumbnails
+        DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + .milliseconds(150), execute: {
+            self.showUiOrCycleSelection(TrackedWindows.focusedWindowIndex)
```
lwouis: Same comment as above for the window delay: delay won't cut it as some apps take multiple seconds and some apps take a few dozen ms.

alt-tab-macos/ui/Application.swift
Robinhuett: Yeah, as I said, im not happy with the delays either.

I'll start with your second suggestion, as this is easily ruled out. Many applications ask you, if you really want to close the window. iTerm with multiple Tabs for example. In this case I think it would be bad UX if we change our UI when the application doesn't.

Now to the first suggestion, which I also thought about. I think this would definitely be the way to go. I briefly looked into [AXObserver](https://developer.apple.com/documentation/applicationservices/axobserver) with which we should be able to do this. Unfortunately I haven't got the notifications working. I'll take another look at this solution since this would be the best.

alt-tab-macos/logic/TrackedWindow.swift
Robinhuett: Good idea, I'll add an var `getAxWindow` with the getter :)

alt-tab-macos/logic/TrackedWindow.swift
Robinhuett: Good catch, I'll change that

alt-tab-macos/logic/Keyboard.swift
Robinhuett: I could probably have made a separate commit for this. It's just, that the application got triggered when only the modifier got pressed. And we don't have to check if the application is being used like here: https://github.com/lwouis/alt-tab-macos/blob/c69123b3d4d7b6884f3f6ebd7848a4d6d8e579e5/alt-tab-macos/ui/Application.swift#L53-L56

I can definitely change this back if you like.

alt-tab-macos/ui/Application.swift
Robinhuett: Update: I got the `AXObserver` working. I'm now looking into how I will implement it to be as flexible as possible. Who knows for what it's good in the future :)

Robinhuett: I'm now using the `AXObserver`. So far it seems to work beautifully. I even got closing fullscreen windows to work with it :)
But extensive testing is more than welcome 🚀

alt-tab-macos/ui/Application.swift
Robinhuett: Same as in my [previous comment](https://github.com/lwouis/alt-tab-macos/pull/94#discussion_r362091513).

Robinhuett:
Sorry, messed something up. Ignore the close und reopening of the PR

alt-tab-macos/logic/Keyboard.swift
Robinhuett: I have thrown this out for the moment

lwouis:
Hey @Robinhuett! Sorry for the delay in reviewing your PR! Holidays are quite a rare treat here in Japan :p

I tested your PR locally and here are my observations:

* Closing windows seems to work great 👍 

* Bug: quitting an app closes the thumbnails panel instead of keeping it open and updating it, like when closing windows

* Bug: closing a fullscreen window refreshes the panel. See these before and after screenshots:

image

image

* Bug: quitting an app by selecting a window on another space successfully quits the app, but switches the user to that window space. I think quitting apps or closing windows should not move the user to any space. Note that closing a window on another space works fine.

* Overall, I think there are some issues with quitting apps, but closing windows seems ready

* I haven't reviewed the code yet in depth, but I'm pretty sure it introduces very complex race conditions that are hard to test/reproduce like closing a window that's slow to close. Then what if you close it again while it's closing? what if there is an animation like minimizing another or the same window at the same time? What if you quit an app and close one of its window, or the opposite? etc. Overall we should write code here that is not strict, so that these messy race conditions due to delays/animations don't silently crash the app and the user is like "Wait why isn't AltTab in my menubar anymore? What happened?"

Robinhuett:
No problem, it's holiday season in Germany too :)

Your observations seam fitting, dealing with quitting apps is a lot worse than the windows.
But for the first bug I am not sure what's happening, the panel shouldn't close except earn releasing the modifier. 
And the switching of spaces is the focus event wich is triggered when releasing the modifier. This is annoying if one only wants to close a window / app. _Maybe we could dismiss the focus event if until another window is selected?_
As for the Bug with partial rendered Thumbnails, I have an idea how to work around this and will later look into it :)

And yeah, this whole closing windows and quitting apps unfortunately has a lot of potential race conditions. Currently I try to test as few things when changing something, especially with slowly or not closing windows. But my main goal is to get it working. But before any of this ships I plan to test everything as much as I can. I have already noted some edgecases which I want to make sure don't make problems.

For the currently not implemented shortcuts, I wanted to get the closing/quitting right first to not litter the commit history too much. I have started with minimizing / fullscreen in another branch. But as I said, one thing at time :)

lwouis:
> But for the first bug I am not sure what's happening, the panel shouldn't close except when releasing the modifier.

Are you able to reproduce that behavior of losing focus when quitting apps on your side?

> Maybe we could dismiss the focus event until another window is selected?

Well I think we should fix the root cause, then we would not have this issue of focusing like that.

> As for the Bug with partial rendered Thumbnails, I have an idea how to work around this and will later look into it :)

I updated #110 with some info regarding the OS APIs that get us the screenshots. You may find it interesting/related.

> But before any of this ships I plan to test everything as much as I can. I have already noted some edgecases which I want to make sure don't make problems.

I would be really interested in *how* you simulate/reproduce these corner-cases. I'm currently investigating a big refac of the whole app that would leverage AX observers that build up a state of the system as apps and windows get open/closed, instead of querying the system state on trigger every time. The advantage would be better performance since we process small deltas instead of the whole system every time, but current performance is good I think, so it's eh. The main motivation is to have a cleaner foundation: by using AX observers we may be able to get `AXUIElement`s of every single window, without resorting to private API black magic like is currently done, when focusing a window from another space for instance.

I just tested locally, and indeed the AX API triggers on a window getting closed on another space. That's encouraging! I used AppleScript to close the window from my current space without visiting the other space. These are the kind of tricks I would love if you could share with me if you have some: easy ways to interact with other space objects, or to simulate a window that's slow to close or don't close, etc.

Robinhuett:
> Are you able to reproduce that behavior of losing focus when quitting apps on your side?

Unfortunately not. Under what circumstances did this issue occur? Was the app you were trying to quit minimized, fullscreen, on the current space? 

> Well I think we should fix the root cause, then we would not have this issue of focusing like that.

I totally agree. When there is a bug which closes the thumbnails panel, this should be fixed before resorting to workarounds.

> I updated #110 with some info regarding the OS APIs that get us the screenshots. You may find it interesting/related.

This is really nice to know. In this case I thought of just removing the windows from the `TrackedWindows`. Since we get notified when the Application is closed it should be be safe to do so here.

> These are the kind of tricks I would love if you could share with me if you have some: easy ways to interact with other space objects, or to simulate a window that's slow to close or don't close, etc.

The reface sounds interesting. But I'm not sure if al that is needed for 'a simple window switcher'. Also I ran into some limitations while using the `AXOberserver`. For example is it triggered when you close a window, but not necessarily wen quitting an application. This is the reason I use KVO for application quitting.

As for the testing, one could say I have gone the low budget route. For not closing windows for example I opened 2 Tabs in a iTerm2 Window so it asks me if I really want to quit. Also since I have way to may Tabs open in Safari it's a kinda okayish testcase, even if it's not an edgecase. But iTerm, running `sleep 10 && exit` is a good test for windows that take a long time. 
And for closing windows / apps on different spaces I mostly use AltTab itself since that's the thing I'm currently interested in. But here `sleep 10 && exit` could come in handy too.
But I like the AppleScript way, since it seems a bit more controlled :D

lwouis:
@Robinhuett I just opened a PR for v3 (#114) which makes it much easier to implement the close window and quit app that you started working on in this PR. I would suggest that either you port your work here on top of my PR, or I'll probably do it some.

I think we should be able to implement both quick-actions and https://github.com/lwouis/alt-tab-macos/issues/39 with the same code since they are both about listening to the apps/windows in the background, which v3 does already for all apps all the time, and simply refreshing the UI when a change happens. I think it's like 5 lines of code + sending the close and quit message when the shortcut are pressed.

Let me know if you want to port your rebase or if you prefer I do it. I think the effort is minimal now that v3 is listening to everything

lwouis:
Hi @Robinhuett! I was wondering if now that the `master` branch is stable and contains everything needed to get to a nice and easy implementation of the ticket, you would be interested in implementing it.

I'm asking, because in case you are no longer interested or too busy, I have bandwidth to do it now. Please let me know this week! :)

Robinhuett:
Hi @lwouis 
I'm sorry for the late response. To be honest I haven't looked at the code for a while so I'm not really up to date on the most recent changes. 
I can definitely take another stab at this, if you haven't already started.

alt-tab-macos/api-wrappers/AXUIElement.swift
```diff
@@ -86,4 +92,14 @@ extension AXUIElement {
         SLPSPostEventRecordTo(&psn_, &(UnsafeMutablePointer(mutating: UnsafePointer<UInt8>(bytes1)).pointee))
         SLPSPostEventRecordTo(&psn_, &(UnsafeMutablePointer(mutating: UnsafePointer<UInt8>(bytes2)).pointee))
     }
+
+    func close() {
+        if isFullScreen() {
+            AXUIElementSetAttributeValue(self, AXAttributeKey.fullScreen.rawValue as CFString, 0 as CFTypeRef)
+            return
+        }
```
lwouis, line 100: I don't know how you found this private constant, but it works great! 👍

lwouis:
@Robinhuett Oh! I thought you were too busy so I started today. I'm porting the close/quit actions from your PR. I think i'll finish the basic ones. Then for the more advanced things, one of us can do it later. I'm talking about moving left/right, resizing, or displaying UI for mouse triggers.

lwouis:
Closing as it was subsumed by #251. Thanks again for your contribution, especially the private API for fullscreen windows! Really cool stuff!

PR #99, opened 2019-12-30
Add app icon

### LINKED ISSUE #38: Add an app icon
This is OSS, so for the icon creation process, I don't want to decide everything. However, I would like to share some important considerations:

**Asks:**

* I would like an editable file to be also provided in the PR (e.g. .psd, .svg, etc). The purpose is that the icon can evolve later
* I would like the icon to avoid using text to avoid i18n issues
* For the menubar, I think it is expected by user to have the choice between a monochrome or color icon. Also Dark Mode should be supported somehow. Not sure what's best practice here. The advantage of text here is that it was supporting everything out-of-the-box there

**Discussion:**

I looked at `/System/Library/CoreServices/Dock.app`, and I like the abstract icon:

image

Also the Mission Control icon (can't find where it is again today):

image

I think something similar may be a good idea: an abstract list of windows, and maybe some arrow showing a switch from one to another.

Also as inspiration, here is what similar apps have gone with:

Windows Switcher
image

HyperSwitch
image

Witch
image

Context
image

akx:
This PR:

* adds an icon for the app (fixes #38)
* switches the menu item to use it instead of the "AltTab" text
* moves the version number text to the menu item's quit option
* (somewhat extraneously) adds a convenience menu option for debugging the selector window

# Icon preview

Screenshot 2019-12-30 at 13 49 48


fturcheti:
> Mentioning the version number in the quit menu is a bit strange. I'm not sure that i truly prefer it, but i think it's smart and should be approved. If you open a dedicated PR i will merge it.

This seems strange for me too. What about displaying the version number in the Preferences window (although an About window would be the ideal)?

akx:
> I really appreciate you took some time to submit this PR and make this icon! First of all I would like to say that the other 2 commits are interesting:
> 
> * Mentioning the version number in the quit menu is a bit strange. I'm not sure that i truly prefer it, but i think it's smart and should be approved. If you open a dedicated PR i will merge it.
> * Having a UI menu to trigger the UI is actually an interesting use-case. I can imagine someone has a keyboard issue but still wants to access a window. For that use-case though, it would be nice to present a ui when clicking the menu, without having to hold `alt`. What do you think about making that menu a first class citizen by adding it as the first option in the menu? I think a clearer label would be "Show".

Sure thing – I'll spin these out as another PR. My UC for the alternate option for the show menu actually was simply to be able to open the selector when accessibility features aren't enabled (I don't know about older macOSes, but on Catalina you have to uncheck and re-check the privileges every time you recompile, so I'd just commented the hard capability checks out from app init).

> Now regarding the icon: I know that I haven't outlined any discussion around what the app icon should be in #38. That makes it difficult for me because here I would like to reject the app icon, even though it's a subjective appreciation from me, and there was no direction/discussion in the ticket.
> My thought process: I asked a friend to look at the icon and he had no idea what it was. I had to show him that this symbol is on the mac keyboard as "option". I think lots of people don't even know this kind of stuff. Another thing is that the key alt is part of the signature alt-tab that people from the Windows world know and love. I think alt is only half of it though, and the shortcut is customizable so for some people, they will not even use option to trigger the app.

I like that you asked for a second opinion! I didn't even think about doing that really... 
A couple counterpoints though: For ATM, the menu bar icon is just for preferences access really and not the main entry point to the app, so its "identity" doesn't really matter all that much. Once you've configured things to your liking, the menu bar button doesn't necessarily have to be there at all. (This of course is served by menubar apps like [Dozer](https://github.com/Mortennn/Dozer)/Bartender/Vanilla.) Secondly, no matter what you can just click on the icon to see "Quit AltTab", so by that point you'll know what the icon is for.

FWIW, my UC for having _some_ icon at all was to have a less bulky menubar item; the "AltTab" text is a little big. :)

fturcheti:
> the menu bar icon is just for preferences access really and not the main entry point to the app, so its "identity" doesn't really matter all that much.

Yep, for its limitation in size and colors the menu bar icon can't be approached as _the app icon_. But IMHO its identity is not irrelevant. It should resemble the real app icon, the one in full colors and sizes, in some way (presenting a simplified element from it, for example). 

So, if the main app icon was a representation of a macOS window with a big ⌥ (alt-option symbol) in it, the menu bar item could be a simple ⌥ without problems as the user could link those images even without knowing the meaning of that symbol.

PR #100, opened 2019-12-30
Soften privilege warnings

akx:
* When keyboard hooking fails, don't crash in `CFMachPortCreateRunLoopSource`
* When not a11y/screenshot privileged, show a warning in the preferences instead of crashing at startup

### DIFF
--- a/alt-tab-macos/logic/Keyboard.swift
+++ b/alt-tab-macos/logic/Keyboard.swift
@@ -23,9 +23,15 @@ func listenToGlobalKeyboardEvents(_ delegate: Application) {
                 eventsOfInterest: eventMask,
                 callback: keyboardHandler,
                 userInfo: UnsafeMutableRawPointer(Unmanaged.passUnretained(delegate).toOpaque()))
+        if eventTap == nil {
+            debugPrint("Could not enable keyboard hook; is accessibility enabled?")
+            delegate.keyboardWasHooked(false)
+            return
+        }
         let runLoopSource = CFMachPortCreateRunLoopSource(kCFAllocatorDefault, eventTap, 0)
         CFRunLoopAddSource(CFRunLoopGetCurrent(), runLoopSource, .commonModes)
         CGEvent.tapEnable(tap: eventTap!, enable: true)
+        delegate.keyboardWasHooked(true)
         CFRunLoopRun()
     }
 }
--- a/alt-tab-macos/logic/SystemPermissions.swift
+++ b/alt-tab-macos/logic/SystemPermissions.swift
@@ -4,25 +4,14 @@ import Cocoa
 // macOS has some privacy restrictions. The user needs to grant certain permissions, app by app, in System Preferences > Security & Privacy > Privacy
 class SystemPermissions {
     // macOS 10.9+
-    static func ensureAccessibilityCheckboxIsChecked() {
-        if !AXIsProcessTrustedWithOptions(["AXTrustedCheckOptionPrompt": true] as CFDictionary) {
-            debugPrint("Before using this app, you need to give permission in System Preferences > Security & Privacy > Privacy > Accessibility.",
-                    "Please authorize and re-launch.",
-                    "See https://help.rescuetime.com/article/59-how-do-i-enable-accessibility-permissions-on-mac-osx",
-                    separator: "\n")
-            NSApp.terminate(self)
-        }
+    static func checkAccessibility() -> Bool {
+        return AXIsProcessTrustedWithOptions(["AXTrustedCheckOptionPrompt": true] as CFDictionary)
     }
 
     // macOS 10.15+
-    static func ensureScreenRecordingCheckboxIsChecked() {
+    static func checkScreenshot() -> Bool {
+        // TODO: This may false-positive if the first window we encounter happens to be owned by the app itself
         let firstWindow = CGWindow.windows(.optionOnScreenOnly)[0]
-        if let cgId = firstWindow.value(.number, CGWindowID.self), cgId.screenshot() == nil {
-            debugPrint("Before using this app, you need to give permission in System Preferences > Security & Privacy > Privacy > Screen Recording.",
-                    "Please authorize and re-launch.",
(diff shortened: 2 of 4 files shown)


alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -2,6 +2,19 @@ import Cocoa
...
+
+See https://help.rescuetime.com/article/59-how-do-i-enable-accessibility-permissions-on-mac-osx
+"""
+    let SCREENSHOT_WARNING_TEXT = """
+Before using this app, you need to give permission in System Preferences > Security & Privacy > Privacy > Screen Recording.
+Please authorize and re-launch.
+
+See https://dropshare.zendesk.com/hc/en-us/articles/360033453434-Enabling-Screen-Recording-Permission-on-macOS-Catalina-10-15-
+"""
+
```
lwouis: I like the idea to move the console log to a more user-friendly UI, but I think the preference panel is not the right place. It should be about setting preferences. I think it would be better to have a dedicated window that shows if you start the app and didn't check the permission checkboxes. That new window would also need to either have a "restart the app" button, or listen to some OS event when the checkboxes are checked, so the window closes itself automatically

alt-tab-macos/logic/SystemPermissions.swift
```diff
@@ -4,25 +4,20 @@ import Cocoa
...
-                    separator: "\n")
-            NSApp.terminate(self)
+            return false
         }
+        return true
     }
 
     // macOS 10.15+
-    static func ensureScreenRecordingCheckboxIsChecked() {
+    static func checkScreenshot() -> Bool {
```
lwouis: I wonder if the private API I switched to in v2 still need this checkbox to be checked. I don't have a Catalina machine to test it though

alt-tab-macos/logic/SystemPermissions.swift
```diff
@@ -4,25 +4,20 @@ import Cocoa
...
         if !AXIsProcessTrustedWithOptions(["AXTrustedCheckOptionPrompt": true] as CFDictionary) {
-            debugPrint("Before using this app, you need to give permission in System Preferences > Security & Privacy > Privacy > Accessibility.",
-                    "Please authorize and re-launch.",
-                    "See https://help.rescuetime.com/article/59-how-do-i-enable-accessibility-permissions-on-mac-osx",
-                    separator: "\n")
-            NSApp.terminate(self)
+            return false
         }
+        return true
     }
```
lwouis: May I suggest a simpler:

```swift
return AXIsProcessTrustedWithOptions(["AXTrustedCheckOptionPrompt": true] as CFDictionary)
```

alt-tab-macos/logic/SystemPermissions.swift
```diff
@@ -4,25 +4,20 @@ import Cocoa
...
         let firstWindow = CGWindow.windows(.optionOnScreenOnly)[0]
         if let cgId = firstWindow.value(.number, CGWindowID.self), cgId.screenshot() == nil {
-            debugPrint("Before using this app, you need to give permission in System Preferences > Security & Privacy > Privacy > Screen Recording.",
-                    "Please authorize and re-launch.",
-                    "See https://dropshare.zendesk.com/hc/en-us/articles/360033453434-Enabling-Screen-Recording-Permission-on-macOS-Catalina-10-15-",
-                    separator: "\n")
-            NSApp.terminate(self)
+            return false
         }
+        return true
```
lwouis: May I suggest a simpler:

```swift
return firstWindow.value(.number, CGWindowID.self)!.screenshot() != nil
```

(you can `!.` the `CGWindowID` because in practice i never found one to return nil. It is always provided by the OS call)

alt-tab-macos/logic/Keyboard.swift
```diff
@@ -23,9 +23,15 @@ func listenToGlobalKeyboardEvents(_ delegate: Application) {
                 eventsOfInterest: eventMask,
                 callback: keyboardHandler,
                 userInfo: UnsafeMutableRawPointer(Unmanaged.passUnretained(delegate).toOpaque()))
+        if eventTap == nil {
+            debugPrint("Could not enable keyboard hook; is accessibility enabled?")
+            delegate.keyboardWasHooked(false)
+            return
+        }
```
lwouis: I'm wondering here if libraries like `ShortcutRecorder`, discussed in [this thread](https://github.com/lwouis/alt-tab-macos/issues/17#issuecomment-546788907) would work without the system permission in the first place. I think the old Carbon API don't require that permission.

Ideally we require those permission only because we need to focus the windows, not because we have a global hotkey, I think

alt-tab-macos/logic/SystemPermissions.swift
akx: On my macOS 10.15.2, if the privilege isn't allowed, the screen previews don't show up, and the privilege prompt shows up behind the selector panel. I think that's still better than preventing the app from starting at all :)

alt-tab-macos/ui/PreferencesPanel.swift
akx: The idea here was that the first thing you do when installing the app is inevitably opening the preferences, and you'd see the warning(s) there. 

I don't really necessarily have the Swift-fu for that dedicated window, so this was simpler :)

lwouis: I think that merging this PR as-is could be considered a regression. I remember when #33 was introduced to fix the issue on Catalina, people were saying that they were not getting the permission prompt every time. We added it at startup, so that if it's not checked, it stops, hopefully the user gets the OS prompt, and this can loop until the user has checked the box.

This PR as-is, puts permissions as a background thing for the user to do. It means that they can have the app running, not working, and potentially confused as to what's happening. I think it's not necessarily a better UX.

I think to truly improve on the UX, we need some dedicated onboarding UI to instruct the users to check the boxes.

akx:
Sure thing. I'll close this for the time being and we can work on a nicer solution such as a proper onboarding UI in the future :)

PR #101, opened 2019-12-30
Add option to hide the space number label

akx:
I have two displays (so ATM counts them as two Spaces), but I don't really need to see the space number in this case.

### DIFF
--- a/alt-tab-macos/logic/Preferences.swift
+++ b/alt-tab-macos/logic/Preferences.swift
@@ -12,7 +12,8 @@ class Preferences {
         "metaKey": metaKeyMacro.macros[0].label,
         "windowDisplayDelay": "0",
         "theme": themeMacro.macros[0].label,
-        "showOnScreen": showOnScreenMacro.macros[0].label
+        "showOnScreen": showOnScreenMacro.macros[0].label,
+        "hideSpaceNumberLabels": String(false)
     ]
     static var rawValues = [String: String]()
     static var minimumWindowSize = CGFloat(200)
@@ -39,6 +40,7 @@ class Preferences {
     static var windowCornerRadius: CGFloat?
     static var font: NSFont?
     static var showOnScreen: ShowOnScreenPreference?
+    static var hideSpaceNumberLabels = false
     static var themeMacro = MacroPreferenceHelper<(CGFloat, CGFloat, CGFloat, NSColor, NSColor)>([
         MacroPreference(" macOS", (0, 5, 20, .clear, NSColor(red: 0, green: 0, blue: 0, alpha: 0.3))),
         MacroPreference("❖ Windows 10", (2, 0, 0, .white, .clear))
@@ -110,6 +112,8 @@ class Preferences {
         case "showOnScreen":
             let p = try showOnScreenMacro.labelToMacro[value].orThrow()
             showOnScreen = p.preferences
+        case "hideSpaceNumberLabels":
+            hideSpaceNumberLabels = try Bool(value).orThrow()
         default:
             throw NSError.make(domain: "Preferences", message: "Tried to update an unknown preference: '\(valueName)' = '\(value)'")
         }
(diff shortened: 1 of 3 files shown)


alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -79,6 +79,7 @@ class PreferencesPanel: NSPanel, NSWindowDelegate {
             makeLabelWithSlider("Max thumbnails per row", rawName: "maxThumbnailsPerRow", minValue: 3, maxValue: 16, numberOfTickMarks: 0),
             makeLabelWithSlider("Apps icon size", rawName: "iconSize", minValue: 12, maxValue: 64, numberOfTickMarks: 0, unitText: "px"),
             makeLabelWithSlider("Window font size", rawName: "fontHeight", minValue: 12, maxValue: 64, numberOfTickMarks: 0, unitText: "px"),
+            makeLabelWithCheckbox("Show space number label", rawName: "showSpaceNumberLabel"),
```
lwouis: Could you rename it "Show space number label**s**"?

alt-tab-macos/logic/Preferences.swift
```diff
@@ -39,6 +40,7 @@ class Preferences {
     static var windowCornerRadius: CGFloat?
     static var font: NSFont?
     static var showOnScreen: ShowOnScreenPreference?
+    static var showSpaceNumberLabel: Bool = true
```
lwouis: No need for the `: Bool` here I think ;p

alt-tab-macos/ui/PreferencesPanel.swift
```diff
@@ -104,6 +105,12 @@ class PreferencesPanel: NSPanel, NSWindowDelegate {
         return makeLabelWithProvidedControl(labelText, rawName: rawName, control: input, suffixText: suffixText, suffixUrl: suffixUrl)
     }
 
+    private func makeLabelWithCheckbox(_ labelText: String, rawName: String) -> NSStackView {
+        let checkbox = NSButton.init(checkboxWithTitle: labelText, target: nil, action: nil)
+        setControlValue(checkbox, Preferences.rawValues[rawName]!)
+        return makeLabelWithProvidedControl(nil, rawName: rawName, control: checkbox)
```
lwouis: image

It looks a bit odd that the checkbox has labels on the right. I understand it's OS standard, but in the context of the preferences panel, I think it may look better to do put the text on the left. Did you try it out? Maybe it looks even more odd? I think the code would be:

```swift
let checkbox = NSButton(checkboxWithTitle: "", target: nil, action: nil)
setControlValue(checkbox, Preferences.rawValues[rawName]!)
return makeLabelWithProvidedControl(labelText, rawName: rawName, control: checkbox)
```

akx: Screenshot 2020-01-02 at 10 14 43

Sure, we can slide it over to the left. Trivial to change later if needed :)

alt-tab-macos/ui/Cell.swift
```diff
@@ -53,7 +53,7 @@ class Cell: NSCollectionViewItem {
         let fontIconWidth = Spaces.singleSpace && !openWindow!.isMinimized ? 0 : Preferences.fontIconSize + Preferences.interItemPadding
         label.textContainer!.size.width = thumbnail.frame.width - Preferences.iconSize! - Preferences.interItemPadding - fontIconWidth
         minimizedIcon.isHidden = !openWindow!.isMinimized
-        spaceIcon.isHidden = openWindow!.isMinimized || Spaces.singleSpace
+        spaceIcon.isHidden = openWindow!.isMinimized || Spaces.singleSpace || !Preferences.showSpaceNumberLabels
```
lwouis: I tested the PR again, and actually realized that this condition is not correct. If `Preferences.showSpaceNumberLabels` is `true`, which is the default, then `Spaces.singleSpace` is ignored, which is breaking the intent of not showing space labels for the single space use-case

I reflected on use-cases here. I thought we would need a dropdown menu like:

Show space number labels:
* Always
* Never
* If multiple spaces are used

However, I can't imagine anyone would want "Always". Why would someone want to see the space when they only use 1 space?

I think the only 3 useful use-cases are:

* I use 1 space at the moment: I don't need to see number labels since they are all the same
* I use multiple spaces at the moment: I prefer to see the labels to help me find which space a window belongs to
* I use multiple spaces at the moment: I prefer not seeing space labels; i'm fine

Thus I think the best UI here would be to rename the checkbox "Hide space number labels", and change the expression to:

```swift
spaceIcon.isHidden = openWindow!.isMinimized || Spaces.singleSpace || Preferences.hideSpaceNumberLabels
```

akx: Excellent points!

Done and done :)

lwouis:
I tested this PR after your changes, and I think I may have introduced a regression with another PR I merged today, but I'm not sure. I opened #107 that has this commit here (keeping your as the author to preserve credits of course) and another commit to bring back the single space logic that was lost previously today I think

PR #104, opened 2020-01-02
Menu changes

akx:
* Removes the unnecessary "Version #VERSION#" menu item
* Adds a new item (previously an alternate item) to open the selector window even when keyboard access fails for some reason

Spun out from #99 as requested.


lwouis:
I opened #106 to add to your work here. I'll merge that PR. Rest assured that your 2 commits are there, with your name to give you all credits! I added 2 commits; you can find more detail in that PR if you are curious as to what I added

Issue #95, opened 2019-12-28
When you alt-tab without any active app application crash


Xicy:
When list count equals 0 and step 0, try to modulo to 0 and return undefined and crash, you can use ` list.count == 0 ? 0 : ` like that function for prevent crashing 

```
 static func moveFocusedWindowIndex(_ step: Int) -> Int {
        return list.count == 0 ? 0 : (focusedWindowIndex + step < 0 ? list.count - 1 : (focusedWindowIndex + step) % list.count)
    }
```

lwouis:
Thank you very much for sharing this issue! I made a fix and merged it to release a new version today! It should fix that issue :)

PR #113, opened 2020-01-05
fix: app crashes when no windows are open

Robinhuett:
The application still crashed when there were no windows to be shown.

Related to #95 and #96


lwouis:
Thanks a lot for noticing this bug, and sharing a fix! This is such an issue issue to regress on when we change the windows listing logic. I wish we had e2e tests for that 🙏

Thanks!

Issue #92, opened 2019-12-27
Opening alt-tab during a Space transition is a bad UX


lwouis:
Spaces support was added in v2. During development I noticed an edge-case: if you invoke alt-tab *during* the space transition animation, it will bring alt-tab window on the previous space, then when you release the shortcut, it sometimes send the window in the void: you can never access it again by any mean.

At trigger time, alt-tab records the current space. I think when done during a space transition, it records the "transition space". I read in another codebase that macOS has a fake space for transition periods.

Even if this impactful bug doesn't happen, the UX is still bad because the alt-tab window is on the previous screen, but the user is still interacting with it, so when they release they get an unexpected behaviour.

HyperSwitch works around this issue by buffering the shortcut, and only triggering when the space transition is done. I think that's the way to go. Either that or ignoring triggers during the transition.

Issue #80, opened 2019-11-10
Preferences panel focus issues


lwouis:
* if you invoke the preferences panel, then focus something else, then invoke it again, the panel titlebar stays unfocused/greyed
* If you invoke the preferences panel, then focus something else, then alt-tab, the panel will move to the foreground until your release alt-tab

Issue #75, opened 2019-11-02
Add CD to Homebrew


lwouis:
Add a step to deliver builds to Homebrew on TravisCI, in addition to the current delivery on Github releases.

Also document the brew option in README.md

[Some preliminary discussion](https://github.com/Homebrew/homebrew-cask/issues/71674#issuecomment-547699742)

OranShuster:
according to him this option is not meant for this.
cask-repair is very easy i use it all the time. it totally automates PRing the new version (it enters the SHA and opens the PR for you after you give it the new version)
it can be automated pretty easily but can also be done manually in <1 minute

lwouis:
> it can be automated pretty easily

I would like to encourage you to open a PR here! Your know-how would be very useful here :)

OranShuster:
tried to automake cask-repair with `expect` and its totally not working...
i guess we can swap it for a simple python script and manually edit the .rb file

OranShuster:
Seems like we don't really need to do this 
https://github.com/Homebrew/homebrew-cask/issues/63171
there is atlest one automated process for updating casks that have an appcast automatically

you can see the output of one of them here
http://macupdater.net/cask-repair/

or here
https://github.com/Homebrew/homebrew-cask/issues/55268

lwouis:
I opened a [PR to open the current cask](https://github.com/Homebrew/homebrew-cask/pull/75859) with the correct minimum macOS version.

PR #114, opened 2020-01-08
v3 POC

### LINKED ISSUE #24: Thumbnails are ordered by z-axis, not recently-used-first
If you switch to an app via the dock icon, all windows of that app become ‘recently used’ in the alt-tab menu.

The documentation for the kCGWindowListOptionOnScreenOnly constant says:

> List all windows that are currently onscreen. Windows are returned in order from front to back. When retrieving a list with this option, the relativeToWindow parameter should be set to kCGNullWindowID.

Original author on [this HackerNews thread](https://news.ycombinator.com/item?id=21256369)

### LINKED ISSUE #93: Opening alt-tab during a window minimize/deminimize animation blocks the UI from appearing until the animation is over
When minimizing or deminimizing a window, if you press alt-tab during the animation where the windows goes in/out of the Dock, the alt-tab UI only shows up when the animation is over.

My hypothesis is that one of the OS API calls, probably `CGWindowListCopyWindowInfo` blocks during these animations. I'm not sure how to workaround that issue

lwouis:
Closes #93 #24 #80 #115 #92 #13 #75 #69 #160 #161 #131 #74 #134 #145 #130 

BREAKING CHANGE: Instead of asking the OS about the state of the whole system on trigger (what we do today; hard to do fast), or asking the state of the whole system on a timer (what HyperSwitch does today; inaccurate) - instead of one of 2 approaches, v3 observes the Accessibility events such as "an app was launched", "a window was closed". This means we build a cache as we receive these events in the background, and when the user trigger the app, we can show accurate state of the windows instantly.

Of course there is no free lunch, so this approach has its own issues. However from my work on it from the past week, I'm very optimistic! The thing I'm the most excited about actually is not the perf (because on my machine even v2 is instant; I have a recent macbook and no 4k displays), but the fact that we will finally have the thumbnails in order of recently-used to least-recently-used, instead of the order of their stack (z-index) on the desktop. It's a big difference! There are many more limitations that are no longer applying also with this approach.

More context: https://github.com/lwouis/alt-tab-macos/issues/45#issuecomment-571898826


lwouis:
@koekeishiya This PR implements that solution [you described a while ago](https://github.com/lwouis/alt-tab-macos/issues/14#issuecomment-550316911).

It has some downsides, but also amazing upsides such as better perf, no hacks to focus other-space windows, better handling of minimized or app-is-hidden windows (or minimized *and* app-is-hidden, which is handled too :p), and it opens the door to live-update of thumbnails if the app/windows are closed in the background while AltTab is opened, for example when doing #94.

My worry now is in the very complex corner cases of the OS, such as an app launching and during launch the screensaver triggers, or who knows which unthinkable cases like this. These would make the state of AltTab drift away from the OS.

I think a brute-force solution here would be to trigger out-of-sync checks once in a while, either on some actions, on a timer, or a mix of both.

Do you think this is the way to go? This POC looks amazing to me and a clear winner, but I'm worried about stability of the app after it has been up in the chaos of the OS for a few days. I'll beta-test myself for a few days before I merge the PR. If you want to review it that would be super cool of course, but I understand your time is limited so I'm mentioning just in case that's of interest to you

alt-tab-macos/logic/Application.swift
```diff
@@ -0,0 +1,99 @@
...
+        AXObserverAddNotification(axObserver, axUiElement!, kAXApplicationActivatedNotification as CFString, selfPointer)
+        AXObserverAddNotification(axObserver, axUiElement!, kAXFocusedWindowChangedNotification as CFString, selfPointer)
+        AXObserverAddNotification(axObserver, axUiElement!, kAXWindowCreatedNotification as CFString, selfPointer)
+        AXObserverAddNotification(axObserver, axUiElement!, kAXApplicationHiddenNotification as CFString, selfPointer)
+        AXObserverAddNotification(axObserver, axUiElement!, kAXApplicationShownNotification as CFString, selfPointer)
+        // TODO: when using the global queue, sometimes it breaks the `runningApplications` observation somehow
+        // To reproduce, open Terminal, quit it, open, quit, repeat. After a while the .insertion and .removal events will stop happening
+//        DispatchQueue.global(qos: .userInteractive).async {
+            CFRunLoopAddSource(CFRunLoopGetCurrent(), AXObserverGetRunLoopSource(axObserver), .defaultMode)
+//            CFRunLoopRun()
```
lwouis: @koekeishiya this piece of code here, and the general GCD system are really a mystery to me. I spent a huge amount of time documenting myself on the topic, but it's such a mess of layer on top of layers of legacy APIs.

I used 2 dedicated `DispatchQueue` in v3, and a few `DispatchQueue.global.async` and `DispatchQueue.main.async` calls. However the fact that I had to comment the global dispatch here is where I'm hitting a wall.

If you have experience with GCD and the run loops, could you possibly take a look at my approach here?

Robinhuett:
@lwouis I'll run this branch too an report any hiccups that may occur.

So far I have 2 things I noticed:
- Like cmd+tab the application shouldn't open when the user is in mission control. But I'm not sure if this is possible to detect.
- When launching the application, applications on "unvisited" spaces are not indexed

Other than that it seem promising :)

I'll look into redoing my PR on this branch, which as you said shouldn't be too hard.

lwouis:
> Like cmd+tab the application shouldn't open when the user is in mission control. But I'm not sure if this is possible to detect.

Wow, what a use-case! Nice catch 🙉 Could you open a ticket for this with a screenshot and some explanations? I think there may be ways to detect it. Also note that the other order: press alt-tab but hold alt, *then* open Expose, correctly hides AltTab UI. You can still release and focus though. I'm not sure if it's good UX or not. Such advanced corner-cases lol

> When launching the application, applications on "unvisited" spaces are not indexed

This was the big limitation that @koekeishiya warned me about from the beginning. First I think it's not very important because this app is meant as a OS utility. It is meant to be opened on startup, so the issue will never manifest again after the first time the user manually opens the app. Even in that scenario, they will go there manually and the app will register the all windows from that space then and it will be back on track.

On top of that issue being minor imo, we can also address it the same way we address the issue in v2: by using the CG apis once, on launch for the initial tracking, then we can resume to normal observation of the OS. What I don't like with this approach is that I will need to put back the old check on "windows are at least 200px tall and 200px wide" that I got to remove in v3 because I can use the AX subrole to identify actual windows in a more robust way.

Like I said, the impact is very minor here. If you disagree with my perception, please open a ticket, then we can discuss it further

> I'll look into redoing my PR on this branch, which as you said shouldn't be too hard.

I just pushed a new commit on the v3 PR that fixes the GCD issues, and has the thumbnails update when the windows change in the background (window closes, window opens, app closes with multiple windows, window title change because you received a notification, etc, etc).

Should be even more easy now to plug the quick-actions on top of the PR

Robinhuett:
> Like I said, the impact is very minor here. If you disagree with my perception, please open a ticket, then we can discuss it further

Nah, it's minor, I just wanted to point it out for completeness. But maybe this should be a disclaimer in the readme so new users trying the application don't think it's not working properly.

> Should be even more easy now to plug the quick-actions on top of the PR

Yeah, pretty easy now :D

> Maybe this issue is gone after my GCD fixes?

Unfortunately not. I think we need pretty sophisticated error checking to catch all those cases where there are animations and space switching. Which get only worse if we are the initiators of the window closing/moving/etc. For example when closing a fullscreen application we inevitably have a space switch when the window closes.

lwouis:
> Like cmd+tab the application shouldn't open when the user is in mission control. But I'm not sure if this is possible to detect.

@Robinhuett I added [a commit](https://github.com/lwouis/alt-tab-macos/pull/114/commits/ba3e22b559217d7d1e920a8d3259ab40b9799fb6) to do that. Here is the behavior, with the commit:

* User trigger AltTab and hold shortcut, now they trigger Mission control: AltTab window fades out. At this point
  * If the user releases alt, AltTab is closed but no focus is given (i.e. nothing happens)
  * If the user keeps holding alt, and stops Mission Control, AltTab window fades back in, and they can resume using it
* User triggers Mission Control. Then they trigger AltTab: shortcut is ignored and nothing happens

I'm very happy with the behavior, however, I'm not super thrilled to add an "expensive" check at the 2 critical points in the UX: when the user summons the app, and when they release to focus another window. I'm keeping it for the moment, but hopefully we find a more efficient way to detect Mission Control in the future.

Note that [transition between spaces](https://github.com/lwouis/alt-tab-macos/issues/92) is still not covered.

lwouis:
I worked more on this PR. The current state seems really robust to me. I added commits to close #93 #24 #80 #115 #92 🙏I'm pretty sure all multithreading/GCD issues are gone now.

@Robinhuett 

> Like cmd+tab the application shouldn't open when the user is in mission control. But I'm not sure if this is possible to detect.

This is supported now

> When launching the application, applications on "unvisited" spaces are not indexed

This is supported now

> When opening an application and while switching spaces, AltTab crashes.

I can't reproduce it. I did the `sleep 3 && exit` trick on a fullscreen'ed Terminal window, and pressed alt-tab before it triggered. On trigger the behavior is perfect: it closes the window, goes to the previous space, and the alt-tab window refreshes without losing focus.

Could you please test this PR a bit? I think it's very close to mergeable 👍

Robinhuett:
Hi @lwouis 
I'll build this branch and will run it as daily driver. If I find something I'll report back to you :)
My first impression is that it works pretty good 👍

Robinhuett:
For some reason often not all windows will be displayed on my machine. In the case of the screenshot below there should be 2 Firefox windows. I'm not entirely sure why this is the case.
Also, as you can see in the screenshot I get an empty row in the thumbnails panel.

 Screenshot 2020-01-18 at 11 35 58

Unfortunately I haven't had time to drill into it and where it's coming from. I should have some spare time tomorrow to investigate this further. I'll update this comment when I can reproduce this consistently or at least find some clues :)

Otherwise v3 pretty stable for me. Good job 👍

lwouis:
I noticed that issue yesterday! I reworked the layout code completely (it
was incorrect before. However the current code has some bug where windows
on the last row are not shown.
I'm aiming at fixing this sometime tomorrow or on monday (JST)!

lwouis:
For some reason, the app fails to detect new windows from IntelliJ. The `kAXWindowCreatedNotification` is not sent by the OS. I'm suspecting that either it's some Java magic, or the app has multiple processes, and the "com.jetbrains.intellij" isn't the main one somehow. I need to dig deeper here

koekeishiya:
You can check the result of AXObserverAddNotification to see whether it succeeded in observing the event or not.

IIRC there is a varying degree of time after launch before an application can be observed and that is likely the issue here.

Sent from my iPhone

On 20 Jan 2020, at 09:24, lwouis <notifications@github.com> wrote:

﻿

For some reason, the app fails to detect new windows from IntelliJ. The kAXWindowCreatedNotification is not sent by the OS. I'm suspecting that either it's some Java magic, or the app has multiple processes, and the "com.jetbrains.intellij" isn't the main one somehow. I need to dig deeper here

lwouis:
Thanks a lot @koekeishiya! You were spot on with your guess! The call returns `AXError.cannotComplete`. I looked at yabai's code and it seems that you simply retry if it fails. That's pretty brute-force but I guess there is no choice.I'll need to retry in a background thread to not block the UI. I'm surprised the AX API creators didn't provide an efficient way to handle the "not ready yet" use-case.

lwouis:
I used the same 10ms delay as in the yabai code I linked to, and it seems to work great!

lwouis:
> For some reason often not all windows will be displayed on my machine. In the case of the screenshot below there should be 2 Firefox windows. I'm not entirely sure why this is the case.
> Also, as you can see in the screenshot I get an empty row in the thumbnails panel.

Sorry @Robinhuett I forgot to update your in this thread: I noticed the issue too. It was a layout problem. I believe i fixed it today. The issue should be gone from this branch now

Robinhuett:
@lwouis No problem. I'm following this PR to at least run the latest version as daily driver. I've also seen your commit and build 7221b53 before heading to work. At the first look it worked fine. If I find something else, I'll let you now :)

lwouis:
I just found that pressing alt-tab while a youtube video is fullscreen'd doesn't show the UI. It still switches, but without minimizing the fullscreen'd window. I'll investigate

Robinhuett:
> I used the same 10ms delay as in the yabai code I linked to, and it seems to work great!

Hi @lwouis I just got to test a bit. Unfortunately the retry hasn't fixed it entirely.
I had the issue one time with the mail app but can't reproduce it. But with iTerm2 I can get the issue regularly (about 50% of the time). When launching iTerm2 it gets added to the applications and this is in the debug log:
```
"OS event: apps launched" [(10669, Optional("com.googlecode.iterm2"))]
"Adding app: com.googlecode.iterm2" []
```
But the window of iTerm2 won't get tracked. For additional windows I open from this one everything is fine.

From my testing, I think what happens is, if the window is spawned between the moment the application is ready and the 10ms delay, it won't get tracked.

lwouis:
Oh interesting. I will download it and try to reproduce on my laptop. I'm not sure what to do here though. With a retry loop, there is always a chance we miss the right timing. What else can I do to wait for the app to be observable, but not too much that a window already popped?

lwouis:
@Robinhuett I can't reproduce the issue with iTerm2. I notice you edited your comment and removed the issue with iTerm2. Was it a false positive?

It seems like you have noticed another now with a fullscreen window scenario. How do you launch the second app, using Launchpad or Spotlight?

Robinhuett:
@lwouis Yeah, sorry about that. I falsely credited die fullscreen issue to the delay. After playing around with the timing I noticed, that it's related to the fullscreen. I have missed it in that moment since I launch my apps using Alfred (should be the same behavior as spotlight).
Interestingly the application gets focused after launching. But I'll test your patch this evening and update this comment with the results.

Thank you for your hard work ❤️

lwouis:
If nobody reports an issue, I will merge this PR this weekend! I would like other PR such as #94 from @Robinhuett to have the new v3 on master to work on top of. I would also like to close #68 and to open a new PR that address the more ambitious #72 (and also #17)

Robinhuett:
@lwouis Sounds good to me :)
I already started some ground work on #94 on top of this branch. So it should be ready shortly after. Should I update my PR or should I open a new one?

lwouis:
> Should I update my PR or should I open a new one?

If the description and scope is the same then just force-push your new work on the existing PR. If it's different, then I would suggest closing the current PR and opening a new one

koekeishiya:
@lwouis I believe the check for `isMissionControlActive` is broken. If the Dock is not set to autohide, your check will trigger a false positive. I recently had to change the logic for yabai at least: https://github.com/koekeishiya/yabai/blob/0fb1d48bbe5b4d13653d646d6adf892ab5cededf/src/event.c#L1246

I don't actually know if it is only on Catalina or not. I am pretty sure that the old logic worked fine back when I was using Mojave, so I cannot say for certain if it is better for you to keep the logic the way it is, or change it to check something else. During my investigation on Catalina there is only one window that was usable to differentiate from the normal Dock window, which had layer == 18 while all the others had layer 20.

I also noticed that you are now using a lot of the private spaces APIs, e.g: CGSManagedDisplayGetCurrentSpace. These APIs only work if the setting *Displays have separate spaces* have been enabled. I don't know if you have a check for this already, so figured I|'d mention it: https://developer.apple.com/documentation/appkit/nsscreen/1388365-screenshaveseparatespaces

lwouis:
> @lwouis I believe the check for `isMissionControlActive` is broken. If the Dock is not set to autohide, your check will trigger a false positive. I recently had to change the logic for yabai at least: https://github.com/koekeishiya/yabai/blob/0fb1d48bbe5b4d13653d646d6adf892ab5cededf/src/event.c#L1246
> 
> I don't actually know if it is only on Catalina or not. I am pretty sure that the old logic worked fine back when I was using Mojave, so I cannot say for certain if it is better for you to keep the logic the way it is, or change it to check something else. During my investigation on Catalina there is only one window that was usable to differentiate from the normal Dock window, which had layer == 18 while all the others had layer 20.

Thank you for sharing that obscure issue! I'm tempted to add the layer 18 check, however I first tested it. I tested it on 10.14 since that's where I added the Mission Control check in the first place. I know for a fact that is was working properly. I remember having a check just on `owner == "Dock"` then I realized that the Dock being shown was triggering the check (I hide the Dock all the time, so I never checked that scenario). I then proceeded to [add the check on `kCGWindowName == nil`](35b48b998073ab74ffc703fab2379637e41889f8) and it fixed it.

Today I upgraded to 10.15, and I just tested now and with the current code (without the layer 18 check), it works in all scenarios. I checked the [ticket in the commit of the line your linked](https://github.com/koekeishiya/yabai/issues/319) and they mention fullscreen which i then tested. Basically it works on macOS 10.14 and 10.15 and I think in all scenarios. Do you think I should still add the layer check? I wish I could reproduce.

> I also noticed that you are now using a lot of the private spaces APIs, e.g: CGSManagedDisplayGetCurrentSpace. Based on my experience these APIs only work if the setting Displays have separate spaces have been enabled, so you might want to test & verify that.
I don't know if you have a check for this already, so figured I'd mention it: https://developer.apple.com/documentation/appkit/nsscreen/1388365-screenshaveseparatespaces

I remember reading stuff online about this, maybe actually on yabai README. However, one day I tested the app with and without that checkbox and it worked as expected. Do you remember what behavior you remember seeing without the checkbox checked, when calling this API? In any case, I'll test again and report back asap.

koekeishiya:
> By the way, in yabai you poll this check every 10ms. Isn't that a bit expensive on the machine? For alt-tab I check when the shortcut is hit and when it's released. I hate that it adds latency, but I thought polling would be too intense on the machine.

I subscribe to an event that notifies me when mission control is entered, so the polling you see here only happens while the user is inside mission control.

So I noticed that the issue on my side is that in Catalina the application is required to have access to some ScreenRecording permission for `CGWindowListCopyWindowInfo` to return window titles on some windows ?? Not sure what's going on there. I wonder if that was the same reason for https://github.com/koekeishiya/yabai/issues/278

Anyway, assuming you have that permission your check should be working just fine. I was pretty sure I did the exact same thing back in Mojave, so, well now we know.

> However, one day I tested the app with and without that checkbox and it worked as expected.

That is probably just me misremembering.I think I got some of the functions you use mixed with one of these: 
```
CGSCopyManagedDisplayForSpace
CGSCopyManagedDisplayForWindow
```
So you probably don't require that setting.

lwouis:
> in Catalina the application is required to have access to some ScreenRecording permission for `CGWindowListCopyWindowInfo` to return window titles on some windows ??

Oh! I went through that one a while ago! Some users said the app is not working on Catalina (#29). Then someone nicely opened a PR (#33) to fix it. Now when the app starts, we check the Accessibility permission *and* the new Catalina screensharing permission. The code is very simple, you can [check it here](https://github.com/lwouis/alt-tab-macos/blob/master/alt-tab-macos/logic/SystemPermissions.swift). For some reason Apple didn't include a permission check for Screen Recording, unlike the Accessibility permission which has an API to check it. You can see my reaction in the PR where I thought the way we check is a bit hacky and there must be an official way, but there is not.

Actually writing this, I realized that I should only do the Catalina check if the user is on Catalina. I will add the check. Anyway I hope this gave you context! Apple is such a poorly documented ecosystem! If only they had billion of dollars to pay some intern to document changes

lwouis:
@koekeishiya regarding the "Displays have separate Spaces" checkbox, I double-checked:

There are [4 APIs](https://github.com/phracker/MacOSX-SDKs/blob/9fc3ed0ad0345950ac25c28695b0427846eea966/MacOSX10.12.sdk/System/Library/Frameworks/CoreGraphics.framework/Versions/A/CoreGraphics.tbd#L779) that look super similar:

```
CGSCopyManagedDisplayForSpace
CGSCopyManagedDisplayForWindow
CGSCopyManagedDisplaySpaces
CGSCopyManagedDisplays
```

In alt-tab, we only use `CGSCopyManagedDisplaySpaces` in that list. As you mentioned, we also use `CGSManagedDisplayGetCurrentSpace`. Actually there are others private APIs we use now. I tried to [document all private APIs in use](https://github.com/lwouis/alt-tab-macos/blob/v3-poc/alt-tab-macos/api-wrappers/PrivateApis.swift), and also past experimentation (at the bottom).

I just tested the app with and without the "Displays have separate Spaces" checkbox checked, and all app functionality seem to work for me, so I think it's safe to say that all private APIs that we do use today are not broken if that checkbox is unchecked.

koekeishiya:
Not sure if you care as you have a solution, but on Catalina there is a private function to request screen capture access:

`extern CGError SLSRequestScreenCaptureAccess(void);`

The function returns 0 if the application does not have access, and it will trigger the "<App> would like to record this computers screen". If the prompt is denied, it won't show again until `tccutil reset ScreenCapture app.bundle.identifier` has been run. If the application already has access, the result of the function call is 1, and the prompt does not appear.

If you think this is more appropriate than the check you have now, keep in mind that this function would might need to be looked up dynamically during runtime using dlopen and dlsym, because you don't want it to fault for older versions of macOS where it (to my knowledge) does not exist.
Using macOS guard #available might work here - not sure.

The method you have implemented seems to work fine, so this is just more an FYI than a recommendation.

lwouis:
> `extern uint8_t SLSRequestScreenCaptureAccess(void);`

Thanks for the tip! That's way cleaner that the current implem. Also `tccutil` is interesting.

I tried to add both for a smoother permission onboarding experience, but I ran into some issues. I clarified them in ticket #127. Apple really makes it a terrible experience for users. I can't understand why their APIs are so poorly designed! It's not rocket science to have a yes/no permission! Very weird state of affaires

koekeishiya:
The permissions resetting can be resolved by CodeSigning the binary.

Sent from my iPhone

On 26 Jan 2020, at 17:12, lwouis <notifications@github.com> wrote:

﻿

extern uint8_t SLSRequestScreenCaptureAccess(void);

Thanks for the tip! That's way cleaner that the current implem. Also tccutil is interesting.

I tried to add both for a smoother permission onboarding experience, but I ran into some issues. I clarified them in ticket #127<https://github.com/lwouis/alt-tab-macos/issues/127>. Apple really makes it a terrible experience for users. I can't understand why their APIs are so poorly designed! It's not rocket science to have a yes/no permission! Very weird state of affaires

lwouis:
> The permissions resetting can be resolved by CodeSigning the binary.

Thanks for bringing that point! I had actually bookmarked [this README from yabai](https://github.com/koekeishiya/yabai/blob/master/CODESIGN.md), and after you posted this, I thought "ok let me finally get to looking at this and understand it". Then I realized you have removed it and updated yabai's main README! Also you added an icon I believe. It looks nice! Well done! The new readme makes some of the requirements/constraints much clearer as well.

Regarding code-signing, I added self-signed code-signing to debug builds. For production builds, I think the only option is #13 and paying Apple for a certificate, unfortunately. Still there may be a middle-ground I discuss in #127.

lwouis:
[AltTab-5.9.4.zip](https://github.com/lwouis/alt-tab-macos/files/4202588/AltTab-5.9.4.zip)

Uploading this file here to test Sparkle auto-update in a more e2e fashion

lwouis:
I can confirm auto-update works! It was hard to test e2e without actually releasing but I'm pretty confident now after the tests I conducted! 🎉

lwouis:
Almost 100 commits and 100 files changed later, I think this PR is ready to be merged on master, which will auto-release v3! 🎉 

image

gingerr:
Congratulations, appreciation and thanks for this massive update guys. Ive updated just now to 3.0; multimonitor use case on Macbook with latest OS working perfectly.

lwouis:
@gingerr note that I've had some issues with CI after the release, as you can see in the updates above. I introduced many hard-to-test CI steps with this releases. For instance, CI now:

* Updates alt-tab on homebrew
* Signs and notarized the app (no more "this app is going to destroy your computer" warning)
* Update `appcast.xml` (which allows the new auto-updater to notice when new versions are released)

I just fixed all the issues _right now_. I had to re-release v3 a few times as a result, unfortunately. I think the version you downloaded is fine, but if you wanna be 100% safe, I suggest you re-download the current v3 :)

gingerr:
Thanks for the hint! Ive updated to 3.0.2 and then 3.0.5 with the integrated auto-update process, love it 💯

I expected to re-allow the screen recording and accesibility permissions after the update but the OS didnt ask me to. I assume this is also no longer needed cause of the app being now signed with your paid developer certificate which is great 🥇 

Thank you for this awesome release.

Improvement idea for the README.md:
You could mention the possibility to install alt-tab-macos via homebrew to onboard homebrew users faster.

lwouis:
> I assume this is also no longer needed cause of the app being now signed

Yes indeed!

> Improvement idea for the README.md

Done! Thanks for pointing it out!

PR #120, opened 2020-01-20
Removed isMetaDown check on Escape keydown (closes #119)

### LINKED ISSUE #119: Escape key on touch bar does not hide UI
On my MBP 15 2019 running Catalina, the key down/up event from the Escape key on the TouchBar does not contain the `modifierFlags` from the keyboard. Therefore `isMetaDown` flag in the `KeyboardHandler` is false and `hideUI` is not called.

phungtuanhoang1996:
This is due to TouchBar's escape key down event does not have modifier flags from the keyboard

Since I chose (^ Control) as Alt key in the preference, I expect all keyboard events to have 
`Event modifier: NSEventModifierFlags(rawValue: 262401)`
when the (^ Control) key is pressed down

However, while keeping (^ Control) down, an keydown event from Esc key on the touch bar does not have the expected modifier flags, instead:
`Event modifier: NSEventModifierFlags(rawValue: 256)`
And since the modifier flags are wrong, hideUI() is not called

I could not find a way to distinguish whether the Esc key down event is coming from touch bar or actual keyboard, so I skipped the isMetaDown check for this case.

### DIFF
--- a/alt-tab-macos/logic/Keyboard.swift
+++ b/alt-tab-macos/logic/Keyboard.swift
@@ -48,7 +48,10 @@ func keyboardHandler(proxy: CGEventTapProxy, type: CGEventType, event_: CGEvent,
             let isRightArrow = event.keyCode == kVK_RightArrow
             let isLeftArrow = event.keyCode == kVK_LeftArrow
             let isEscape = event.keyCode == kVK_Escape
-            if isMetaDown && type == .keyDown {
+            
+            if type == .keyDown && isEscape {
+                return dispatchWork(application, false, { application.hideUi() })
+            } else if isMetaDown && type == .keyDown {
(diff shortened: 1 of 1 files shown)


lwouis:
Thank you for opening a PR! Please make sure that: (a) the title and description of the PR convey what the change is about, by mentioning a ticket for instance; and (b) the commits messages [follow the convention](https://www.conventionalcommits.org/en/v1.0.0-beta.4/#summary), so your PR can be merged.

phungtuanhoang1996:
I have updated my commit message / PR title + elaboration. 
Thanks for pointing me to the convention.

lwouis:
Thank you very much for contributing to the project! Macbook pro users will be grateful! :)

Issue #217, opened 2020-04-08
Crash, Preferences related?


gingerr:
Hi @lwouis,

ive just updated to the latest version. Some of the dropdown options were empty in the Preferences panel. Ive set them and then the application crashed when i summoned AltTab and later it crashed already when i opened the Preferences panel.

Ive deleted the defaults `defaults delete com.lwouis.alt-tab-macos` and restarted the application and was able to use the application with its default settings.

![image](https://user-images.githubusercontent.com/13162840/78759860-932a0200-7980-11ea-8561-611d2b08cc54.png)

But in the preferences panel all dropdowns were empty (nothing is selected) and when i changed any of them the application crashed when i summoned AltTab.

When i select something for all of the available dropdowns the application crashes already when opening the Preferences panel.

Can you observe the same behavior?

Mariusz

lwouis:
It looks like it's localization-related. Let me try to reproduce this issue

lwouis:
I can reproduce it when the app is in non-English language. The problem is that the dropdown preferences are stored in the local language, meaning when it's time to check what's selected, it fails to find the key, which coincides with the English value.

I'm surprised that this issue is surfacing now, since we had the same logic for the "Show on display:" dropdown for quite a while now. I guess it shows that very few users use this.

I'll work on a fix as soon as possible!

lwouis:
I release an update with a fix. @gingerr and @buttercrab, you could please tell me if it's fixed?

Please run `defaults delete com.lwouis.alt-tab-macos` beforehand, to remove old preferences. I will open a ticket to discuss how to deal with old preferences in a nicer way

buttercrab:
@lwouis It's working great!

gingerr:
@lwouis Can also confirm. Dropdown preferences were pre-selected and a change is also working nicely.

I did not change any preferences in the last few months so I did not observe this earlier. Today i took a look inside to see what new features you have implemented and especially the ShortcutRecorder keybindings.

Thank you! 💯

gingerr:
How did you fix it? Do you save an english key in the preferences instead of the localised string now?

PR #218, opened 2020-04-08
fix: avoid text flickering on main ui (closes #197)

### LINKED ISSUE #197: Thumbnail titles sometimes flicker on UI display
Similar to #115, but it is happening under a very different technical context now that the main UI is no longer using a `NSCollectionView`. This ticket is about trying to stop or at least mitigate this flickering.

The flickering is due to the `NSView`s being reused from a pool for better performance. We show the UI as soon as possible to limit latency, but that means sometimes the user sees the title being updated.

Mitigating this could include trying to update the titles in advance in the background

lwouis:


lwouis:
I fixed #217 also in that PR

Issue #211, opened 2020-04-03
Correct the readme / Make a landing page


lwouis:
The readme hasn't been reviewed in a while. Its `Configuration` section for example is outdated and wrong.

This ticket could also include making a real landing page for the project, using Github Pages or others. This page could improve the experience by:

* Being localized or at least showing AltTab language support in a personalized way, based on user browser language
* Provide builds optimized for the macOS version. See #158 
* Show a video of the app in motion
* Explain the preferences

ezity:
@lwouis If I fork the repository, create a landing page using GitHub Pages, and then merge my fork back, does the site created get merged back in? The way to create the site (through the repo settings) makes me suspicious that might not be the case.

I can work on this, but will probably need some guidance from you. Can we discuss in chat of some sort?

lwouis:
Hi @ezity! Could you please [join our discord](https://discordapp.com/invite/mHvmcqT)?

lwouis:
Also, we might be distant timezones, so live communication might be difficult.

I checked out Github Pages today, and documentation is lacking and unclear on how the whole system works. That being said, it seems safe to be merged from a fork, since this process involves a folder in the repo, either at the root or in the `docs/` branch.

lwouis:
I released the [new website](https://alt-tab-macos.netlify.app/) today. I'm keeping this ticket open, as there are a lot of "nice to have" things we can do now that the bulk of the work is done. See the list in the OP for details.

lwouis:
I've thought about adding code on the landing page to detect if the macOS version of the browser downloading the app is within the compatible range, or to check if it's an Apple Silicon mac:

```js
<script>
function checkCompatibility(event) {
  const version = navigator.appVersion.match(/OS X (\d+)_(\d+)/)
  const majorVersion = Number(version[1])
  const minorVersion = Number(version[2])
  if (majorVersion > 10 || (majorVersion === 10 && minorVersion < 12 || minorVersion > 15)) {
    if (!confirm('AltTab is compatible with macOS 10.12 to 10.15. It may still work on a different version, but with some bugs. Are you sure you want to download?')) {
      event.preventDefault()
    }
  }
  if (majorVersion > 10 || (majorVersion === 10 && minorVersion >= 15)) {
    const context = document.createElement('canvas').getContext('webgl')
    const extension = context.getExtension('WEBGL_debug_renderer_info')
[...]

PR #223, opened 2020-04-08
Hotfix for 3.6.2 crashing on old preferences

lwouis:


lwouis:
@ryenus there is no config file. There used to (the readme is not up-to-date, see #211). Now it's using AppKit [`NSUserDefaults`](https://developer.apple.com/documentation/foundation/nsuserdefaults).

I'm going to add a preference version there indeed. I actually used to have all this implemented a long time ago. Checking if the preference format has changed, and deleting if it was too old. Then when I migrated to `NSUserDefaults` I removed all this.

I'm going to implement it now. I don't know how to do it in a really nice way. I imagine some libraries may be covering that already. I will check online. This PR is a hotfix to avoid having other people upgrade and be stuck, so time was of the essence

lwouis:
@ryenus I raised a PR to have versioning of the preferences and handle migration up and down better: #224

PR #414, opened 2020-07-11
feat: add preference to hide menubar icon (closes #103)

### LINKED ISSUE #103: Add preference: remove AltTab from the menubar
I don't like to keep lots of apps on the menubar, and AltTab has no crucial feature to be sitting there. Actually, I think it should "invisible" to the user, just showing up when activated to switch windows. So it would be nice to have an option to remove its icon from the menubar.

L1cardo:
feat: add preference to hide menubar icon (closes #103)
屏幕录制2020-07-11 上午1 04 09 2020-07-11 01_09_23

### DIFF
--- a/src/ui/App.swift
+++ b/src/ui/App.swift
@@ -13,7 +13,7 @@ class App: NSApplication, NSApplicationDelegate {
     static let repository = "https://github.com/lwouis/alt-tab-macos"
     static var app: App!
     static let shortcutMonitor = LocalShortcutMonitor()
-    var statusItem: NSStatusItem!
+    static var statusItem: NSStatusItem!
     var thumbnailsPanel: ThumbnailsPanel!
     var preferencesWindowController: PreferencesWindowController!
     var feedbackWindow: FeedbackWindow?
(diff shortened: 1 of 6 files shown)


src/ui/App.swift
```diff
@@ -50,8 +50,19 @@ class App: NSApplication, NSApplicationDelegate {
         // TODO: undeterministic; events in the queue may still be processing; good enough for now
         DispatchQueue.main.async { () -> () in Windows.sortByLevel() }
         preloadWindows()
+        
+        #if DEBUG
+        showPreferencesPanel()
+        #endif
+
```
L1cardo, line 57: If we set `hide menubar icon` to `true`, it will do nothing when we debug the app, we won't be able to open the preferences panel, so I think we may just leave this code.

L1cardo:
@lwouis 
I reopened this PR with decent commits, I think it is beautiful now.

lwouis:
Hey @L1cardo! I just merged these commits as well as a few others in PR #415. I'm closing this PR as its content are now merged from the other one.

This will trigger a new release in a moment. I want to thank you again for your great contribution! It's been one of the most requested feature since the beginning of the app, but I never took the time to do it as i felt it's non-essential, and I was focusing on bugs and performance. Now it's finally here for everyone to enjoy! Thanks again for contributing! 👍

PR #468, opened 2020-07-24
Use more inclusive language

lumaxis:
Following a trend across the industry and also [Apple's recently updated style guide](https://help.apple.com/applestyleguide/#/apsg1a3a0436?sub=apd9aaca4623), this replaces "blacklists" with "blocklists".

I also went ahead and adjusted the German localization accordingly 🙂

Alternatively, I had considered using "Ignorelists" instead, let me know if you prefer that.


lumaxis:
Hmm, I guess this doesn't work with the way that the localizations are pulled from POEditor in CI 😅

lwouis:
Hi @lumaxis! Thank you for contributing this PR!

I would like to address the PR in 2 separate angles:

### Political motivation of the PR

I personally disagree with this trend. I think it is silly to censor non-discriminatory language. A black sheep, a black swan, a blacklist, none of these are related to BLM and skin color discriminations. I don't like that we start censoring everything willy-nilly without thinking.

Some cases like "slave/master" could negatively affect some people, and I understand that it's creating a nicer environment to updates these. However, I find the specific replacement of blacklist/whitelist to be a bad idea that caters to people who decide to be offended by everything and anything, even when there is no actual offense to be found objectively (see [origin of the words](https://en.wikipedia.org/wiki/Blacklisting)).

That being said, it is not a big deal, so I'm fine with the change. If it helps some people feeling welcome using the app, great. Note that change has a non-zero impact though, as it invalidates a bunch of already done localization work, that each language contributors are going to have to re-enter.

### Review of the PR

As you have noticed, we use poeditor.com to handle localization. If I were to merge this PR, we would have the issue where poeditor.com would now be out of sync with the codebase.

I will update poeditor.com myself, using contents from this PR for German. I'll open a PR that cherry picks the refactors you did in the Swift code. I'll make sure to commit with your name/email for this, so you get full credits. It's a bit tricky at the moment to do the full poeditor.com flows which involves 3 scripts in `scripts/`, so I thought it's easier if I do it. As I said, it's important to me to preserve full credits, so I'll keep your name/email in the commit as it is in this PR. If you also want to be credited for your contribution to the German localization, please go here and register as for the German localization. The scripts will pick up your account there to credit you automatically on the following release 👍

lwouis:
I realized that I should be able to simply update poeditor.com, and then restart the CI build for this PR, and it should pass.

In the meantime, I thought a bit about this "Blocklist" approach. I think a more elegant solution to this conundrum would be to call it "Exceptions" as this tab in the preferences indeed treats of exceptions to the rule; exceptional cases to handle differently. It has 2 lists today, but could host more in the future, as AltTab integrates better with specific apps.

I think "Exceptions" would be easier to translate in other languages as well. I considered how to translate "Blocklist" in French for instance and could only find very long translations, whereas I think "Exceptions" is a pretty clear and understood concept in most languages that should have a 1 word translation.

What do you think?

lumaxis:
Thanks for looking into this! I don't really want to discuss the motivation of these changes since ultimately, it comes down to this for me as well:
> If it helps some people feeling welcome using the app, great.

(However, I don't think black- and whitelists ever made any semantic sense. What's "black" about any of this? 😄)

Finding a good replacement that actually makes sense and is hopefully even more intuitive and self-explanatory is a bit tricky though, indeed. 
I find "Exceptions" a bit odd too. Exception from what? Plus, at least to developers, it always has the "error" connotation, too. After thinking about it a while, I think I like "Ignorelists" best semantically but it's indeed a bit tricky to translate.
All of them are usable though 👍🏼

lumaxis:
That is true for general and generic cases but imo does not really fit this specific case. Renaming at all provides an opportunity to pick a name that is way more self-explanatory and intuitively more sense to a user.

koekeishiya:
I think the word blacklist *is* the proper word to use in this context. Looking at the history of the word blacklist/whitelist it is pretty clear that they do not originate from a place of malice, but rather due to the properties of the "colours" themselves. 

A black object absorbs all light and thus has a negative connotation, where as white reflects all light and has a positive connotation to it. The property that black/white absorbs/reflects all light also make them the inverse of each-other, which is another property that carries into the meaning of the word black/white-list.

At the end of the day this is @lwouis  project and if he thinks an alternative is reasonable to use, that is fine, but I think that censoring certain words in our language by intentionally misinterpreting them is bad. Are we no longer "allowed" to use the term *a black hole* to refer to said phenomenon?

lwouis:
@koekeishiya I share your perspective that there is no discrimination in these words, and that putting discrimination into them is reaching. However, I want to be pragmatic here, and if it helps some people feel better, I want to support it.

The pragmatic *issue* here though is cost of change and UX. I tried to think about how to translate ignore-list and deny-list in French for instance. I can't think of a single good way to translate this concept without a full sentence like "Liste d'applications refusées" and even then it's a mess compared to the equivalent for "blacklist" that's already available, or "Exceptions" which is literally the same as in English.

I tried to find on the website you shared earlier @lumaxis for international recommendations but found nothing. I searched within Apple [reference glossaries](https://help.apple.com/applestyleguide/#/apsg1a3a0436?sub=apd9aaca4623) and found 1 entry for deny-list: "Liste d'interdiction" which I'm sure no-one would understand. I found 94 entries for "blacklist" though. It being spread out macOS like this shows how wildspread the word is, and how understandable it is for users.

frypf:
Interesting discussion, and interesting also that Apple's own suggestions are so far entirely Anglocentric. Indeed, even their guidelines linked by @lumaxis suffer from the burden of longer sentences @lwouis alluded to in his French translations. Perhaps "Omissions" or even "Exclusions"?

lumaxis:
Hey, sorry for not replying here sooner. Was on vacation for a bit and lost track of this issue.

> Perhaps "Omissions" or even "Exclusions"?

I only speak a few languages somewhat fluently but this seems like a promising direction for possible translations.

@lwouis Thank you for your thoughtful replies, I very much appreciate them, especially since you don't necessarily share the urge to change these. I obviously understand that this is not the most important thing in the world and changing a few words wouldn't solve system racism anyway. Still, I think words influence our thinking and therefore matter.

But this is a free, volunteer project after all and I won't blame anyone for not having the time to invest in this.
I still think changing this would be a valuable change, so if you think that there's anything I could do or help with to get it done, let me know! If not, my feelings won't be hurt if we just close this PR 🙂

lumaxis:
Maybe one thing to still mention: The other original reason why I started looking into this was that to me, the German translation, "Schwarze Liste", does not actually make much sense in the context of this app. To me, it has a very strong "_persona non grata_" connotation whereas in AltTab, it seems to me to really semantically fit better with something along the lines of _exclusion_ or _exception_.

lwouis:
> I still think changing this would be a valuable change, so if you think that there's anything I could do or help with to get it done, let me know! If not, my feelings won't be hurt if we just close this PR 🙂

I'm really overloaded with [important tickets at the moment](https://github.com/lwouis/alt-tab-macos/issues), unfortunately. Some of these are big crashes at launch, regressions introduced in v5, v6, or v6.2. I can't justify spending time on changing wording when some users have the app crash on them, or have their workflows broken.

Let's keep this open, and I'll revisit when I've worked my way through the current urgent issues. If you want to help move the other issues along, of course that would directly help move this very issue quicker of course. Let me know if you're interested in doing that 👍

AfzalivE:
>A black object absorbs all light and thus has a negative connotation, where as white reflects all light and has a positive connotation to it. The property that black/white absorbs/reflects all light also make them the inverse of each-other, which is another property that carries into the meaning of the word black/white-list.

@koekeishiya This is going a bit into philosophy. Why is absorbing light a negative thing and why is reflecting light a positive thing? I think we should consider that while using terms like blacklist and whitelist don't cause tangible harm, they do, cause us to think that black is bad (and should be excluded) and that white is good (and should be included). The origin might not have been racist but as language adapts and changes, so should we. Many swear words started as just normal words, but have now become offensive due to their usage in prevalent context.

@lwouis, thank you for considering updating the language :)

lwouis:
I've decided that the alleged benefits of this PR don't outweigh its costs. Furthermore, there hasn't been activity on this in a year, which is in itself probably demonstrating the non-essential nature of the topic. I'm closing the PR 🙇

Issue #169, opened 2020-03-10
Reduce size of app icon file


lwouis:
The app icon is a 1024x1024 export from Figma. It weights 562kB. Yet, after I used apple's `iconutil --convert icns resources/icons/app-icon.iconset` to generate the icns file, it weights 859kB.

There may be a way to avoid this unjustified bloating.

lwouis:
I noticed that the `Autoupdate.app` file embedded into the Sparkle framework is only 36kB even though it has multiple png within, one of them being 512x512. Looking at their repo, I found this [very interesting Makefile](https://github.com/sparkle-project/Sparkle/blob/b42642690535a43f459511e175716d5ad17fc5b8/Makefile-icon).

akx:
Running pngquant (as in the linked makefile) even on the lightest, gentlest settings leads to some unfortunate artifacts on our icon:

![Screenshot 2020-03-30 at 10 09 07](https://user-images.githubusercontent.com/58669/77884862-8c0f3f80-726e-11ea-8d15-a84dd334c539.jpg)

OxiPNG (via [ImageOptim](https://imageoptim.com/mac)) has lossless savings of 800kb -> 580kb, but I'm quite unable to get a ICNS smaller than the current one. Speaking of which, how was the .icns file generated in the first place though? I see no script (nor the original PNG file, but it was easy to extract from the icns) in the repo.

lwouis:
@akx I'm very happy to see a fellow performance enthusiast! Too often people add megabytes and bloat apps for no reason. Let me try to explain the icon situation a bit more. There is some behind the scenes that's not documented anywhere, unfortunately.

The original icon was designed by @fturcheti in Figma. Here lies the first issue: Figma uses some advanced SVG features that have wildly varying implementation in popular engines

Which icon sizes to ship?


Assets catalog (i.e. `.xcassets`)
https://github.com/lwouis/alt-tab-macos/issues/38#issuecomment-596416020

lwouis:
@akx I'm very happy to see a fellow performance enthusiast! Too often people add megabytes and bloat apps for no reason. Let me try to explain the icon situation a bit more. There is some behind the scenes that's not documented anywhere, unfortunately.

**SVG-world and Figma**

[...]

Issue #544, opened 2020-08-19
AT 5.0.0 crashes if I open Preferences


mparry:
**Describe the bug**

AltTab's 5.0.0 generally seems to be working unless I attempt to open the Preferences window via the menu bar icon's menu, at which point AT immediately crashes, every time. This was fine pre-5.0.0 (though I don't know exactly when I would have last tried opening it).

This didn't seem the same as #543, so hopefully this isn't a duplicate.

[at_crash.zip](https://github.com/lwouis/alt-tab-macos/files/5095344/at_crash.zip)

**Steps to reproduce the bug**

1. Click on the AT menu bar icon and select Preferences...
2. AT crashes

mparry:
The further investigation on #543 suggests this is the same thing, so I'm going to close this

Issue #543, opened 2020-08-19
5.0.0 crashes right on startup


mfn:
**Describe the bug**

This morning I was created with notification to update to 5.0.0 which I did.

After the upgrade step, nothing happened. No app icon visible but also <kbd>cmd-tab</kbd> didn't work. **Not even the native one 😅**

Luckily I could use <kbd>cmd-space</kbd> and also HyperSwitch was still activated <kbd>cmd-tab</kbd> from there (I still use it for in-app switching as AltTab doesn't support it).

After re-downloading (same effect) I found out some things:
- the app does start for a brief moment, showing what instead should be an app icon for ~1s white rectangle in the top bar which then disappears
- I found something in the console.app

```
default	08:55:54.737328+0200	Finder	LAUNCHING:0x0-0x11f11f AltTab foreground=0 bringForward=0 seed=1522 userActivityCount=0
default	08:55:54.744582+0200	loginwindow	-[ApplicationManager checkInAppContext:eventData:] | ApplicationManager: Checked in app : AltTab
default	08:55:54.747206+0200	runningboardd	Resolved pid 18846 to [executable<AltTab(501)>:18846]
default	08:55:54.748687+0200	runningboardd	[executable<AltTab(501)>:18846] This process will not be managed.
default	08:55:54.748718+0200	runningboardd	Now tracking process: [executable<AltTab(501)>:18846]
default	08:55:54.749180+0200	runningboardd	Acquiring assertion targeting executable<AltTab(501)> from originator [daemon<com.apple.coreservices.launchservicesd>:146] with description <RBSAssertionDescriptor; uielement:18846; ID: 322-146-1995; target: 18846> attributes = {
    <RBSDomainAttribute: 0x7fbee0910040; domain: com.apple.launchservicesd; name: RoleUserInteractive; sourceEnvironment: 0x0>;
}
default	08:55:54.749293+0200	runningboardd	Assertion 322-146-1995 (target:executable<AltTab(501)>) will be created as active
default	08:55:54.749763+0200	runningboardd	Acquiring assertion targeting executable<AltTab(501)> from originator [daemon<com.apple.coreservices.launchservicesd>:146] with description <RBSAssertionDescriptor; uielement:18846; ID: 322-146-1996; target: 18846> attributes = {
    <RBSDomainAttribute: 0x7fbede626920; domain: com.apple.launchservicesd; name: RoleUserInteractive; sourceEnvironment: 0x0>;
}
[...]

PR #545, opened 2020-08-19
Quick fix for #543 (removes quit button from prefs dialog)

akx:
This is really just band-aid; something in laying out that button breaks things, but it's not strictly required.

Refs #543
Refs #544


lwouis:
Thanks again for finding the root cause for the crashes on v5 @akx. I'm closing this PR as I reverted the code for the quit button as I thought I would rather have some warnings about layout than lose the quit button.

Thanks for all your support!

Btw this reminds me, did you see that I closed #169? I know you were interested by this issue. I found [a solution](https://github.com/lwouis/alt-tab-macos/commit/81a3553af2eb2daef22dcb90242d7baf1134f6fd#diff-96710857f25ded53c3f908ada7e21fb0R5) and now the icon has the correct footprint

PR #725, opened 2020-12-11
Update brew cask syntax

karbassi:
`Warning: Calling brew cask install is deprecated! Use brew install [--cask] instead.`

### DIFF
--- a/docs/Introduction.md
+++ b/docs/Introduction.md
@@ -12,7 +12,14 @@ permalink: /
 
 [**Download the latest release**]({{ site.github.latest_release.assets[0].browser_download_url }})
 
-Alternatively, you can use [homebrew](https://brew.sh/): `brew install --cask alt-tab`
+Alternatively, you can use [homebrew](https://brew.sh/): 
+```sh
+# Latest homebrew
+brew install --cask alt-tab
+
+# Homebrew 2.5 or below
+brew cask install alt-tab
+```
 
 ## Compatibility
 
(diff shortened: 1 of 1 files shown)


karbassi:
@lwouis That sounds great to me. Let this sit for the new few months then review and merge if it still applies then.

i3p9:
Just thought I'd let you know that the `brew cask` has been disabled. We must use `brew install --cask` now. 

```bash
Error: Calling brew cask install is disabled! Use brew install [--cask] instead.
```

lwouis:
@i3p9 man the homebrew maintainers are moving way too fast in my opinion. What are we supposed to do now? If we migrate to the new syntax, then it will break users of previous versions of homebrew, which is probably most users by far today.

I think we should keep the old command for now, as it clearly says what to do and gives the new command to run. That way users of recent versions get this, and can work around the warning. If we migrate, then users of previous versions get no clear explanation, and think that AltTab is broken or something of that nature.

mnin (approved):

ayroblu:
Just gonna chime in here with something I saw in another repo, might be a reasonable compromise:
image

karbassi:
> Just gonna chime in here with something I saw in another repo, might be a reasonable compromise:
> image

I like this.

I also noticed that you don't need the `--cask` part. You can do `brew install alt-tab`

mnin (approved):

lwouis:
Closing this PR. I implemented similar changes directly as part of a bunch of other changes. You can [see the commit here](https://github.com/lwouis/alt-tab-macos/commit/01683af8ef9c7b7ecf654226376da1788ebedf43). I committed with your name/email, so that you get credited for your contribution:

image

Next release you will also be added to credits on the website and in the app itself 👍

Thanks for the contribution!

PR #731, opened 2020-12-14
Show window controls on hover, even when option is disabled

samdenty:
Awesome project!

This PR makes the window controls always appear on hover - regardless of the "Mouse hover" option.
I think this makes the most sense, as you are still able to click-to-focus windows when the option is disabled anyway.

### DIFF
--- a/src/ui/main-window/ThumbnailView.swift
+++ b/src/ui/main-window/ThumbnailView.swift
@@ -68,7 +68,7 @@ class ThumbnailView: NSStackView {
         if let shouldShowWindowControls = shouldShowWindowControls_ {
             self.shouldShowWindowControls = shouldShowWindowControls
         }
-        let shouldShow = shouldShowWindowControls && isHighlighted && !Preferences.hideColoredCircles && !window_!.isWindowlessApp && !Preferences.hideThumbnails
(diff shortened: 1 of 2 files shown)


lwouis:
Hey @samdenty! Thanks for sharing this PR! Really appreciate you contributing to this project!

I tested you PR, and indeed I think it's a clear-cut improvement. I don't see a downside to it.

It has been [discussed to have another kind of visual UI](https://github.com/lwouis/alt-tab-macos/issues/286#issuecomment-739615500) to signify "the mouse is hovering this thumbnail" compared to "the mouse is hovering, and moved selection to this thumbnail". Something like a thin border framing the thumbnail under the mouse, as Windows 10 does. This may be an improvement to clarify hovering, for user who unchecked mouse hovering selection. However in the current state, this PR is still an improvement.

I will merge your commit bundled with other commits, in https://github.com/lwouis/alt-tab-macos/pull/728/commits. Don't worry though, you remain the author of the commit, and you will get credited on the website, and in the app itself. You will be added in the next-next release though, as credits lag 1 release behind due to how they are automated.

Thanks again! Let me know if you're interested in working on something else. I can help you if some areas of the codebase are not easily accessible 👍

PR #741, opened 2020-12-23
Disable standard tab detection for all JetBrains apps, not just EAPs

### LINKED ISSUE #716: ATM breaks certain rename functionality in PyCharm 2020.3
## Describe the bug

This is a weird one...

When ATM is enabled, using the `Ctrl+T, Enter` shortcut in the PyCharm 2020.3 IDE to rename an identifier doesn't work correctly anymore. Where you should be able to edit text using multiple carets, only arrows and delete/backspace work anymore, so you can't quite rename identifiers (unless you like very short identifiers). Only clicking in the editor restores normal functionality.

I initially thought this was a JetBrains problem/regression and raised an issue on their tracker: https://youtrack.jetbrains.com/issue/PY-45867

## Screenshots / video

I overlaid my keypresses on top of the gif to make things hopefully a little clearer.

### Broken (when ATM is active)

rename_fail_keys

### Working (ATM not active)

This was tried a couple hours later, hence the code isn't the same.

rename_work_keys

## Steps to reproduce the bug

1. Have ATM running.
2. Open PyCharm 2020.3 (Professional in my case).
3. Set up a document where you can attempt to rename an identifier.
4. Hit Ctrl+T to open the Refactor This popover, then hit enter to use "Rename".
5. Attempt to modify text. It doesn't work correctly.

## Workarounds

* If I use the context menu -> Refactor -> Rename option, this doesn't occur
* If I use shift+f6 instead of ctrl+t and enter, this doesn't occur

akx:
Fixes #716 (likely)


akx:
@lwouis Can I get a review for this? I think the CI fail is unrelated...

lwouis:
@akx I'm sorry, I saw it during the holidays, and I'm trying to find time to review it. I'll try my best to work on it this week 👍

scripts/update_contributors.sh
```diff
@@ -11,7 +11,6 @@ projectId="316051"
   echo -e "They helped [develop the app](https://github.com/lwouis/alt-tab-macos/graphs/contributors):\n"
 
   curl https://api.github.com/repos/lwouis/alt-tab-macos/contributors \
-    -H "Authorization: token $GITHUB_TOKEN" |
```
lwouis, line 14: It may seem like this is unnecessary, but I actually added this because from the CI machine (TravisCI), requests to github get quota'd and denied. Using a token means it doesn't fail because of that, as it deducted from that token's quota vs from the IP quota.

lwouis:
Thanks for the PR @akx! I'm going to cherry-pick the commits on the `develop` branch to bundle changes in the next release. Of course you stay these commits's author 👍

Closing this PR

Issue #371, opened 2020-06-09
Add CLI support


v-braun:
First: thanks for this app!
I tried HyperSwitch for long time but it was so annoying that the order of windows (most recent window) is based on the active space, if you switch windows between spaces the order is messed up (maybe something that you can to the compare table on your website).

No my question:
I wanted to built a Alfred Workflow to list and search within windows, this app would be a good starting point to get all windows if it would have a CLI.

Is there something like this?

lwouis:
Hi @v-braun! Thanks for sharing this feedback!

I'm not sure i get a clear picture of what that CLI would look like.

Could you please specify it more? Like example commands you would need to build your workflow, and what they output, in more details?

Currently there is none of that. The app is only usable through its GUI

v-braun:
Hey, what I imagine is a CLI interface to query Windows, Apps and Spaces.
For my specific use case only windows are needed.

This interface would be cool:
``` sh
# list all windows from all spaces
# options: 
# -o / --output (table or json): specify output format
# -s / --space: filter by space 
# -d / --display: filter by display

#example: 
alt-tab window list -o json


# output:
[{
   "id": "unique window id",
   "appId": "process id of the app",
   "title": "window title",
   "size": {
      "height": 100,
      "width": 100
   }
}]


# activate a window
#example: 
alt-tab window activate <<window id>>

```

Other subcommands could be:
- alt-tab window
  - subcommands
    - alt-tab window list
    - alt-tab window activate
    - alt-tab window resize
    - alt-tab window move
- alt-tab space
  - subcommands
    - alt-tab space list
    - alt-tab space activate

- alt-tab display
  - subcommands
    - alt-tab display list

lwouis:
That's interesting! I have a few questions:

* You said: "to query Windows, Apps and Spaces", but then there is nothing about Apps in the subcommands. Instead there is a subcommand for Displays. I guess you also said that you personally only have interest in Windows?

Regarding filtering, you wrote:

> # -s / --space: filter by space 
> # -d / --display: filter by display
> # -t : filter by title (regex??, fuzzysearch??)

I wonder if filtering wouldn't be better done by the user. Like doing regex search in titles or fuzzysearch could be done by piping `grep` or `fzf`, or writting a program that feeds off the JSON output and sorts/filters it.

Do you care about the windows order? Should it be ordered by latest-used like in the graphical AltTab?

[...]

PR #761, opened 2021-01-03
Persist window state to directory

ayroblu:
In a first attempt to tackle #371, I thought that writing to the filesystem would be useful.

Menu layout:
image

(I'm using a ram disk as volatile fyi)

I then hooked this up to alfred and it worked as expected!:

image

TODO:

- [x] Allow selecting any directory rather than using a hardcoded `Documents/alt-tab-windows.json`, probably with button + NSPathControl
- [ ] Move this persist to another preference tab, maybe Advanced?
- [ ] Fix layout with NSPathControl shrinking when longer path selected as well as handling long paths
- [x] Cocopods update + workspace version updates shouldn't be in this PR
- [ ] Define what actually should be exported rather than just exporting the random fields I wanted (spaceIndex looks wrong, and maybe others)
- [x] Hook on to all the correct events, I only put it on one update statement, perhaps it's reasonable to put it more / different places
- [ ] Benchmarks for the performance degradation
- [ ] Tests?
- [x] Use async
- [ ] Write better swift (no experience here, hope it's okay?)

### DIFF
--- a/appcast.xml
+++ b/appcast.xml
@@ -5,6 +5,20 @@
     <language>en</language>
 
     <item>
+      <title>Version 6.12.0</title>
+      <pubDate>Tue, 17 Nov 2020 16:51:35 +0000</pubDate>
+      <sparkle:minimumSystemVersion>10.12</sparkle:minimumSystemVersion>
+      <sparkle:releaseNotesLink>https://alt-tab-macos.netlify.app/changelog-bare</sparkle:releaseNotesLink>
+      <enclosure
+        url="https://github.com/lwouis/alt-tab-macos/releases/download/v6.12.0/AltTab-6.12.0.zip"
+        sparkle:version="6.12.0"
+        sparkle:shortVersionString="6.12.0"
+        sparkle:edSignature="XSiVRJQOMOGxEaq9Z405VSTvmN6FOHhRp1azCNceY8xoVu+8CLwBDRu8hjPtHsoC+Mn8gnB2Z6m2cPlUEzvwDg==" length="7238803"
+        type="application/octet-stream"/>
+    </item>
+
+
+    <item>
       <title>Version 6.11.0</title>
       <pubDate>Wed, 11 Nov 2020 15:48:36 +0000</pubDate>
       <sparkle:minimumSystemVersion>10.12</sparkle:minimumSystemVersion>
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -1,5 +1,17 @@
 # Release notes
 
+# [6.12.0](https://github.com/lwouis/alt-tab-macos/compare/v6.11.0...v6.12.0) (2020-11-17)
+
+
+### Bug Fixes
+
+* window was not shown after closing tab (closes [#696](https://github.com/lwouis/alt-tab-macos/issues/696)) ([a7e96f2](https://github.com/lwouis/alt-tab-macos/commit/a7e96f2))
+
+
+### Features
+
+* add slovak localization ([06027dc](https://github.com/lwouis/alt-tab-macos/commit/06027dc))
+
 # [6.11.0](https://github.com/lwouis/alt-tab-macos/compare/v6.10.0...v6.11.0) (2020-11-11)
 
 


lwouis:
Hey thanks for sharing this first draft! It's nice to see you got useful results so fast!

Regarding the preferences, I'm a bit worried as I've learned on this project how they can grow out of control. I'm thinking that an opt-in checkbox is necessary because of the potential perf impact, but also the certain "intrusiveness" (as in, the user may expect the app to not write anything on disk, given its functionality).

Regarding your todo list:

> Benchmarks for the performance degradation
> Write better swift (async? etc)

I think that you should use async IO to write to disk. Then I expect the performance impact should be almost null.

To benchmark it, you could use Instruments and do a before/after the change. That's how I get a general performance understanding of changes that I suspect would impact performance.

> Tests?

Before this app is 100% side-effects from OS interactions, I gave up on automated testing from the get go. I kind of regret it sometimes, as probably some unit tests here and there would have helped, but at the same time even in retrospect I feel I made the right call back then. So there are no automated tests today; it's all manual QA. I have a [section if you're curious](https://alt-tab-macos.netlify.app/contributing#qa) to see the kind of fun macOS had in store when I built this app.

> Define what actually should be exported rather than just exporting the random fields I wanted (spaceIndex looks wrong, and maybe others)

You may want to discuss this with @v-braun and see what kind of usage he would have of it, to decide what to expose. A good start would be to expose the list of windows observed, with all their attributes. Other useful stuff may include the info on Spaces, Screens, running Applications.

By the way, I realized that your approach of exporting the state of the app as JSON is useful to get info on the state, but will not allow the user to interact with the windows or other elements. For the window for instance, even if you exported their ID, and passed it to some AppleScript, it couldn't work in all cases. I'm using [private APIs](https://github.com/lwouis/alt-tab-macos/blob/master/src/logic/Window.swift#L159) and [private SPIs](https://github.com/lwouis/alt-tab-macos/blob/master/src/logic/Window.swift#L172) to focus windows, as there is no public API to do it correctly for some edge-cases (like windows on other Spaces, or minimized in the Dock on other Spaces, I forgot)

> Hook on to all the correct events, I only put it on one update statement, perhaps it's reasonable to put it more / different places

If you serialization code is fast enough, you probably want to rewrite the whole file on any change in the state, instead of updating specific lines by doing delta changes. You may put the update code [after this switch](https://github.com/lwouis/alt-tab-macos/blob/master/src/logic/events/AccessibilityEvents.swift#L15). If you're exposing Applications state, you may want to add the update code [here as well](https://github.com/lwouis/alt-tab-macos/blob/master/src/logic/events/WorkspaceEvents.swift#L22).

ayroblu:
Preferences: This is what I have right now, I thought I need the button, but actually I might not for the path control + I need to make it show the icons, it's a bit bland right now. I think it also fails silently on permissions issues right now to I need to put that in somewhere too:

image

This worked well enough for me with osascript for alfred (including changing to windows in different spaces) (yes ALOT of stackoverflow there). I never minimize / hide so I haven't tested those specifically

```
on alfred_script(q)

set argv to extract_argv(q, "|||||")

set proc to item 1 of argv
set windowId to item 2 of argv as integer
set windowName to item 3 of argv

tell application "System Events" to tell process proc
    set frontmost to true
    windows where title contains windowName
    if result is not {} then perform action "AXRaise" of item 1 of result
	click (first menu item whose name contains windowName) of menu "Window" of menu bar 1
end tell

end alfred_script

on extract_argv(source_string, new_delimiter)
	set backup to AppleScript's text item delimiters
	set AppleScript's text item delimiters to new_delimiter
	set argv to every text item of source_string
	set AppleScript's text item delimiters to backup
	return argv
end extract_argv
```

Regarding cocoapods, I definitely had to upgrade my targeting atleast, I might not have needed to do the `pod install`, so I can try revert that, but xcode won't compile for me without some config change. I'm using Big Sur, and I installed XCode yesterday (it took like 2 hours 😭️) @lwouis

ayroblu:
Oh looks like I didn't need to pod install, sorry I must've been mistaken!

Stvad:
@ayroblu I see you've discarded this approach in favor of #768. I'm not quite sure what's the state of that one, but I given that you got to a working Alfred WF here I was wondering if you had this somewhere on the branch that I can build for myself? 🙂

ayroblu:
Any particular reason you don't want to use #768 with a HTTP? Otherwise I can see if I can revert the changes on the reflog for this PR

Stvad:
Mainly laziness 😅. You have an Alfred WF for this one & all 😛. Do you have a WF that works with the other PR? 🙂   
Also, I think it can be nice to have code of this preserved for posterity 🙂 

(also got scared by other pr's size initially, but it seems most of the code is copying source of a library? btw I'm Swift ignorant, so I'm curious on why you'd want to do that vs dependency?).

ayroblu:
@Stvad I made #886 (sorry couldn't do it to this pr) for posterity. I've only made my workflow work for the HTTP one now so see that PR for details.

Dependencies are checked into git, otherwise they work in a similar way to other languages. I'm not sure if it's [strictly necessary](https://stackoverflow.com/questions/45587084/should-we-commit-pod-files-to-version-control-system-git-or-svn), but this is called "vendoring" and it's pretty common especially for lower level languages. Obviously most useful if you want to make changes, whether for performance or compatibility reasons, or for more of a "it just works" when sharing a repo

Stvad:
@ayroblu thanks a lot!

lwouis:
@ayroblu 

> This worked well enough for me with osascript for alfred (including changing to windows in different spaces) (yes ALOT of stackoverflow there). I never minimize / hide so I haven't tested those specifically

I'm pretty sure your osacript will not work in every case, unfortunately. AltTab uses [private API and private SPI](https://github.com/lwouis/alt-tab-macos/blob/master/src/logic/Window.swift#L173) to enable focusing every window in [every situation](https://alt-tab-macos.netlify.app/contributing#qa). Maybe good enough for your particular usage though, which is already great!

ayroblu:
Yes! I think with the http server implementation in #768 it was so easy to hook in to the APIs that I just did that instead 😅 
You can see it [here](https://github.com/lwouis/alt-tab-macos/pull/768/files#diff-406ff24dd944c0b9993ba782366f964f001a67d80ae25782e1223d55dad94e96R46)
Then it's just an http call away!

PR #769, opened 2021-01-09
Add Swift Formatter

ayroblu:
Added `swift-format` configuration. Tutorial [here](https://exyte.com/blog/how-to-start-working-with-swift-format). Looks like someone made an xcode extension [here](https://github.com/kuglee/Swift-format)

Configuration set in `swift-format-configuration.json` which is just the default configuration with the indent changed from 2 to 4. Default configuration is dumped with the following command:

```sh
swift-format -m dump-configuration > swift-format-configuration.json
```

Format command is:

```sh
swift-format -m format --configuration swift-format-configuration.json --recursive -i src
```

- [ ] Is this okay?


lwouis:
Hey @ayroblu! Thanks for another PR!

I wasn't aware that there was a swift formatter from Apple. Interesting. After reading their website, I see some negatives:

> NOTE: No default Swift code style guidelines have yet been proposed. The style that is currently applied by swift-format is just one possibility, and the code is provided so that it can be tested on real-world code and experiments can be made by modifying it.

Not great, but fine I guess

> swift-format depends on SwiftSyntax and the standalone parsing library that is distributed as part of the Swift toolchain. The SwiftSyntax version in use must match the toolchain version, so you should check out and build swift-format from the branch that is compatible with the version of Swift you are using. This version dependency is also expressed in the SwiftSyntax dependency in Package.swift.
>
> Xcode Release | Swift Version | swift-format Branch
> -- | -- | --
> – | Swift at main | main
> Xcode 12.0 | Swift 5.3 | swift-5.3-branch
> Xcode 11.4 | Swift 5.2 | swift-5.2-branch
> Xcode 11.0 | Swift 5.1 | swift-5.1-branch

That sounds like lots of work to handle this, as time passes and we bump XCode and/or Swift.

Another thing is that we probably want to add a run of this tool on CI if we roll with it. After all, CI is the reference environment. This means automating the download and versioning of this tool in the CI script.

ayroblu:
Yup! Up to you, for me, I don't like coding choices for things that have no effect on the final result so I always use a formatter for my own stuff. They pretty much automatically work with terminal apps like vim and usually pretty easy on CI like GitHub actions

PR #826, opened 2021-02-17
refactor: Use 'blocklist' for the list of things to be blocked

bomberstudios:
This PR replaces all occurrences of an offensive term in the code with a more descriptive term, “blocklist”.

I have used Google Translate to update the localizations for the languages I do not understand, so maybe this should be followed up with a revision by the project translators (if any).

I’ve also not included any code to migrate existing preferences to use the new term, but could do so if I can find the time.


bomberstudios:
Turns out I should check first before writing code, and this is in fact a duplicate of https://github.com/lwouis/alt-tab-macos/pull/468

Closing!

lwouis:
Sorry about that @bomberstudios! Thank you for contributing! If you'd like to work on other topics, let me know 👍

Issue #870, opened 2021-03-18
Brew installation on site should be `brew install --cask alt-tab`


Luckey-Elijah:
**Describe the bug**

The latest versions on homebrew prefer cask like `--cask` rather than `cask` in the install.

The site says to install like: `brew cask install alt-tab`
When the brew-preferred method is: `brew install --cask alt-tab`

**Screenshots / video**

N/A

**Steps to reproduce the bug**

N/A

meiser79:
Over and over the same bug reports, see #725, #780, #829 and #870.

PR #884, opened 2021-03-26
docs: fix brew install command

klausbayrhammer:
See https://stackoverflow.com/questions/30413621/homebrew-cask-option-not-recognized


lwouis:
This keeps happening, unfortunately 😅

Please see https://github.com/lwouis/alt-tab-macos/issues/870#issuecomment-803264620

PR #892, opened 2021-04-02
docs: Correct Homebrew cask command

danilofuchs:
`brew cask` is not a command anymore
https://stackoverflow.com/questions/30413621/homebrew-cask-option-not-recognized

### DIFF
--- a/docs/Introduction.md
+++ b/docs/Introduction.md
@@ -12,7 +12,7 @@ permalink: /
 
 [**Download the latest release**]({{ site.github.latest_release.assets[0].browser_download_url }})
 
-Alternatively, you can use [homebrew](https://brew.sh/): `brew cask install alt-tab`
+Alternatively, you can use [homebrew](https://brew.sh/): `brew install --cask alt-tab`
 
 ## Compatibility
 


danilofuchs:
I see it was already fixed in other PRs. Why were they not merged? This is an issue in the first command the user ever attempts

lwouis:
Hi @danilofuchs, you can see the [rationale for waiting here](https://github.com/lwouis/alt-tab-macos/pull/725#issuecomment-753416944).

danilofuchs:
Maybe time enough has passed? 3 months is a lot of time for people to upgrade homebrew.

Keeping the old command just makes it seem like the project is not well maintained, IMHO

Perhaps adding a note for old version users?
I am led to believe more than 50% of people have already updated, hence the large number of PRs addressing this.

lwouis:
> Maybe time enough has passed? 3 months is a lot of time for people to upgrade homebrew.

3 months may seem like a long time to you, but the world at large is slower paced. Things take years, not months. Look at [macOS versions of homebrew users](https://formulae.brew.sh/analytics/os-version/365d/) for instance.

> Keeping the old command just makes it seem like the project is not well maintained, IMHO

I would rather give this impression, and have users read a warning and follow up, rather than break the command fully for some users on older versions.

Now, it's indeed extra work for me to see people opening the same PR and tickets. I will implement [what's suggested in the first PR here](https://github.com/lwouis/alt-tab-macos/pull/725#issuecomment-762481765), as it seems like it has the least downsides of all the routes we could take.

PR #900, opened 2021-04-07
feat: added NICE DCV to the don't show list

xanathar:
This adds [NICE DCV](https://docs.aws.amazon.com/dcv/index.html) to the list of applications for which alt-tab-macos is disabled by default if full-screen (i.e. the `disableShortcutsBlacklist`); NICE DCV is a remote display protocol on the same vein of VNC, RDP and other applications already in the list.


lwouis:
Hi @xanathar! Thank you for sharing this here. Not everyone would take some time to think about others facing the same issue and sharing their fix like this. I appreciate the nice gesture!

I cherry-picked the commit in the other PR so I can release it as part of a bundle of changes. You will get credited for your contribution on the website and in the app after the next release 👍

PR #1047, opened 2021-07-24
docs: fix typo in Introduction

nella17:


lwouis:
Thanks @nella17 for fixing this typo. I looked at how Apple is presenting this in macOS, and it looks like this:

image

Maybe it would be even better to write `简体中文`, `繁體中文` instead of `中文 (简体)`, `中文 (繁體)`? What do you think?

nella17:
I think this is a good idea.

lwouis:
I cherry-picked the commit in [this other PR](https://github.com/lwouis/alt-tab-macos/pull/1058).

Don't worry, you'll get credited in the git history as well as on [the website](https://alt-tab-macos.netlify.app/), and in the app itself 👍

PR #1080, opened 2021-08-19
feat: add Vim-style window selection shortcuts

david-r-cox:
This change introduces a new "Vim keys" option for window selection
controls. With this option enabled, it is possible to select windows
in AltTab without leaving the home row of the keyboard.

The keys H, J, K, and L move the window selection left, down, up, and
right, respectively. AltTab already uses 'H' as the default key to
hide/show apps. If "Vim keys" is enabled when hide/show is in its
default configuration, the hide/show action will be shadowed by "Vim
keys." It is possible to enable both hide/show and "Vim keys" by
overriding the default key for hide/show.

For more information on Vim, see https://www.vim.org/about.php

### DIFF
--- a/src/ui/generic-components/CustomRecorderControl.swift
+++ b/src/ui/generic-components/CustomRecorderControl.swift
@@ -74,10 +74,15 @@ class CustomRecorderControl: RecorderControl, RecorderControlDelegate {
             }
             return $0.shortcut.keyCode == shortcut.keyCode && $0.shortcut.modifierFlags == shortcut.modifierFlags
         }) {
-            let existing = ControlsTab.shortcutControls[shortcutAlreadyAssigned.id]!
             let alert = NSAlert()
             alert.alertStyle = .critical
             alert.messageText = NSLocalizedString("Conflicting shortcut", comment: "")
+            if shortcutAlreadyAssigned.id.starts(with: "cycleSelectionLeftVim") {
+                alert.informativeText = String(format: NSLocalizedString("Shortcut already assigned in window selection with Vim keys.", comment: ""))
+                alert.runModal()
+                return
+            }
+            let existing = ControlsTab.shortcutControls[shortcutAlreadyAssigned.id]!
             alert.informativeText = String(format: NSLocalizedString("Shortcut already assigned to another action: %@", comment: ""), existing.1.replacingOccurrences(of: " ", with: "\u{00A0}"))
             if !id.starts(with: "holdShortcut") {
                 alert.addButton(withTitle: NSLocalizedString("Unassign existing shortcut and continue", comment: "")).setAccessibilityFocused(true)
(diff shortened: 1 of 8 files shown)


lwouis:
Thank you very much for sharing this PR! I took a look at the code, and everything seemed great to me!

I'll need more time to QA it. As you've noticed, the app is quite complex with users being able to remap shortcuts very flexibly.

src/ui/generic-components/CustomRecorderControl.swift
```diff
@@ -74,10 +74,15 @@ class CustomRecorderControl: RecorderControl, RecorderControlDelegate {
...
-            let existing = ControlsTab.shortcutControls[shortcutAlreadyAssigned.id]!
             let alert = NSAlert()
             alert.alertStyle = .critical
             alert.messageText = NSLocalizedString("Conflicting shortcut", comment: "")
+            if shortcutAlreadyAssigned.id.starts(with: "cycleSelectionLeftVim") {
+                alert.informativeText = String(format: NSLocalizedString("Shortcut already assigned in window selection with Vim keys.", comment: ""))
+                alert.runModal()
+                return
+            }
+            let existing = ControlsTab.shortcutControls[shortcutAlreadyAssigned.id]!
```
lwouis, line 85: I was able to crash the app by:

* Activating `vim keys` in the preferences
* Clicking to change `Select previous window` and picking `K`

It then crashed as `ControlsTab.shortcutControls[shortcutAlreadyAssigned.id]` was `nil`

I think there may be issues in the lifecycles of shortcut bindings.

lwouis:
@david-r-cox Someone opened https://github.com/lwouis/alt-tab-macos/issues/1229 today, and I was reminded of this PR. I think we both forgot about this PR for a while.

Do you think you could look at my comment above, and maybe update it?

It was also closed automatically when I deleted the `develop` branch. Could you please point the PR to the `master` branch instead? PRs should target it. You'll probably need to rebase on top of `master` as it's been a while as well.

david-r-cox:
Hey @lwouis, thanks for reaching out. 
I've been using my fork of alt-tab daily and wanted to revisit this PR at some point because alt-tab + vim keybindings is now essential to my workflow. 😁 

It's great to see that others are also interested in an option for vim-style movement shortcuts.

I'm going to start working on a fix for the edge-cases you found, and hope to have it completed by the end of next week.

david-r-cox:
@lwouis I've opened a new PR here:
https://github.com/lwouis/alt-tab-macos/pull/1272

src/ui/generic-components/CustomRecorderControl.swift
david-r-cox: @lwouis I believe the arrow key crash predates this PR. I was able to reproduce it using https://github.com/lwouis/alt-tab-macos/commit/f5fc5ba8131fb65a2e54cf5c1795626b5118c1c4.

PR #771, opened 2021-01-10
Add support for Apple Silicon

ayroblu:
Inside Podfile, update Sparkle and AppCenter, then pod install

Also, AppCenter requires a major version update, and so requires following [this migration guide](https://docs.microsoft.com/en-us/appcenter/sdk/getting-started/migration/apple-sdk-update)

I set the release to use a self signed certificate (not included in this PR) and build release works

Issue #1062, opened 2021-08-01
Big Sur: Preferences panel UI broken on low-dpi screen (1440p LCD monitor)


Rabattkarte:
**Describe the bug**

I have the following setup:

- macOS Big Sur (v. 11.5.1)
- 13,3" integrated Retina-Display @ 2560 × 1600; MacBook Pro (2020)
- 27" external Display @ 2560 × 1440; acer B276HUL

Both screens render the settings menu of AltTab differently. This only affects the **Appearance settings**.

I believe this has something to do with the way macOS renders UI elements at different resolutions and pixel densities. I'm no expert but I've read about similar issues [in this blog post here](https://tonsky.me/blog/monitors/).

**Screenshots / video**

- integrated display (not broken):
![macbook_retina](https://user-images.githubusercontent.com/16440184/127768630-ea18d2cf-5813-41b9-8ce1-abf25c3ff7ec.png)

- external display (broken):
![lcd_1140p](https://user-images.githubusercontent.com/16440184/127768627-9d4dfeca-462c-4450-906c-211b129aaf57.png)

**Steps to reproduce the bug**

1. Open settings.
2. Click on 'Aussehen (Appearance)'
3. Move the settings window across screens

lwouis:
I put the app in German, and tried to reproduce with:

* External monitor: AW2521HF, 1920 x 1080 @ 60 Hz
* Macbook pro internal monitor: Built-In Retina LCD, 2560 x 1600 Retina

But it was working fine. I think it's a layout bug that expresses itself only in rare cases, when exact dimensions and screen sizes happen. I'm afraid there is not much to be done here. Hopefully it will go away after some work on layout on these page, like #351 for example.

rasmusbe:
I'm having the same problem on my external display and I'm using English, so it's not a German bug.
![image](https://user-images.githubusercontent.com/2124386/131636348-3be0d87a-f306-44e8-8af6-635b8ff727e9.png)

![image](https://user-images.githubusercontent.com/2124386/131636489-a9d22618-8e14-45fd-bc3f-5a8e65f283c1.png)

![image](https://user-images.githubusercontent.com/2124386/131636449-06fb431e-ebb8-4603-90cd-d007cf5a3ce9.png)

lwouis:
I'm guessing it's either pixel rounding, or not taking into account the size of the menubar somehow, on some monitors. I can't really address it if I can't reproduce it locally, fix it, then confirm it's fixed.

[...]

Issue #158, opened 2020-02-11
Release builds without embedded swift libraries


lwouis:
Switch 5 is natively supported starting with macOS 10.15. This means there is no need to embed the 10-12MB of Swift support files for uses on that OS or later.

[See more details here](https://www.google.com/amp/s/eclecticlight.co/2019/02/24/last-week-on-my-mac-apple-quietly-pre-announces-major-change-in-macos-10-14-4/amp/)

On the Mac AppStore there is app thinning: the app is thinned depending on the environment of the use downloading it. I'm not sure how we are supposed to handle this outside the MAS. Maybe having multiple downloads for each OS version? It may be possible to guide the user by making some web frontend with some JS detecting their OS and suggesting the best download for them. Hopefully Github Pages supports that. I need to double-check.

It would be nice in general having a more accessible download page anyway. Github releases is pretty unwelcoming for non-developpers

lwouis:
Migrating to Swift 5 is just changing the language setting from 4.2 to 5. No other action is required 👍

I experimented, and indeed, the Swift libraries are no longer embedded in the build starting from 10.14.4:

![image](https://user-images.githubusercontent.com/106195/80300677-c0c4d700-87d9-11ea-9a69-460d8e7fdedc.png)

I'm worried about providing the benefits of a build targeting a more modern OS version (i.e. smaller app, better performance) vs the issues with dealing with multiple builds. Here are some issues that worry me:

* CI has to build 2 times now. Longer CI builds, more potential for issues
* The UX for end-users is now confusing. How do we communicate the 2 options effectively. This requires #211 to have browser user-agent detection to help suggest the best build, with a fallback to manual selection
* How to handle homebrew? I imagine they have solutions for this scenario

gingerr:
Why just not drop support for older OS versions? I think with the usage of the private APIs it could become painful to support/test/dev and also build now for multiple OS versions.

lwouis:
> Why just not drop support for older OS versions?

Dropping support for older OS versions = dropping support for some people. The end-goal of this app is to be useful to as many people as possible. For this reason, I take compatibility very seriously. I would only remove backwards compatibility for a very large benefit. Here reducing the app size is not worth it for me. I would rather support 2 distribution channels and deliver an optimized builds for people on newer versions, rather than distribute only 1 build for them, and drop support for everyone else.

PR #1094, opened 2021-08-31
Add support for Apple Silicon #2

mnin:
Follow up for #771


mnin:
@lwouis can you check this build?

lwouis:
@mnin thank you very much for sharing this! I'll review it. I see already that there are [some minor issues to fix](https://github.com/lwouis/alt-tab-macos/pull/1094/commits/70176ab329c75fe53915934fde23f738bea205fa), but I'll need more time to really review some [more complex stuff](https://github.com/lwouis/alt-tab-macos/pull/771#discussion_r558337608) like CI, the pods, etc

mnin:
> I see already that there are [some minor issues to fix](https://github.com/lwouis/alt-tab-macos/pull/1094/commits/70176ab329c75fe53915934fde23f738bea205fa), but I'll need more time to really review some 

I will retract this commit, when we are done. But I needed it because otherwise the build didn't run (linked in the commit message).

Can you see if we can get the build from Travis here?

lwouis:
I see that you had to upgrade Sparkle for Universal compatibility. Unfortunately, Sparkle is missing a fix I made on my fork of Sparkle back in the day.

I submitted a PR to them today, but that's on their `master` which is v1.27.0 I think. This version changes the webview and other things, so not sure if it works with AltTab setup.

In the meanwhile, I updated the my fork to host a fixed v1.24.0. This `Podfile` should now include `pod 'Sparkle', :podspec => 'https://raw.githubusercontent.com/lwouis/Sparkle/fix-iframe-popup-1.24/Sparkle.podspec'`

Podfile
```diff
@@ -3,8 +3,8 @@ platform :osx, '10.12'
 target 'alt-tab-macos' do
   use_frameworks!
   pod 'LetsMove', :git => 'https://github.com/lwouis/LetsMove.git', :commit => '7abf4daed1a25218f2b52f2dfd190aee5a50071c'
-  pod 'Sparkle', :podspec => 'https://raw.githubusercontent.com/lwouis/Sparkle/master/Sparkle.podspec'
+  pod 'Sparkle', '1.24.0'
```
lwouis, line 6: I see that you had to upgrade Sparkle for Universal compatibility. Unfortunately, Sparkle is missing a fix I made on my fork of Sparkle back in the day.

I submitted a [PR](https://github.com/sparkle-project/Sparkle/pull/1938) to them today, but that's on their `master` which is v1.27.0 I think. This version changes the webview and other things, so not sure if it works with AltTab setup.

In the meanwhile, I updated the my fork to host a fixed v1.24.0. This `Podfile` should now include `pod 'Sparkle', :podspec => 'https://raw.githubusercontent.com/lwouis/Sparkle/fix-iframe-popup-1.24/Sparkle.podspec'`

Podfile.lock
```diff
@@ -38,12 +37,12 @@ CHECKOUT OPTIONS:
...
   ShortcutRecorder: fdf620aca34101b0cba3b10fca815e0459254189
-  Sparkle: ea073ba9c69cee3ebaab8685ba1e732e14d249f7
+  Sparkle: 270cd27377bf04e9c128af06e3a22d0f572d6ee3
   SwiftyMarkdown: 033abc9dad9a37903f81aec5188cade436f40520
 
-PODFILE CHECKSUM: 78df0a34910c4c031a65ac3b58dc004c21b552d0
+PODFILE CHECKSUM: f7be5f3250126444c8ff49c41543c32894eae84c
 
-COCOAPODS: 1.9.1
+COCOAPODS: 1.10.2
```
lwouis, line 48: What was the motivation behind 2e0685bb where you bumped cocoapods? I try to not change versions of anything unless there is an established benefit to it

src/ui/preferences-window/tabs/AppearanceTab.swift
```diff
@@ -33,7 +33,6 @@ class AppearanceTab {
             LabelAndControl.makeLabelWithCheckbox(NSLocalizedString("Hide apps with no open window:", comment: ""), "hideWindowlessApps"),
         ])
         grid.column(at: 0).xPlacement = .trailing
-        grid.rowAlignment = .lastBaseline
```
lwouis, line 36: Could you please elaborate on what the issue was, and which macOS version / arch it was happening on?

src/ui/preferences-window/tabs/GeneralTab.swift
```diff
@@ -1,4 +1,5 @@
 import Cocoa
+import LoginServiceKit
```
lwouis, line 2: This change puzzles me. It seems you switching from existing working code to a third-party [library which uses the same Apple APIs](https://github.com/Clipy/LoginServiceKit/blob/master/Lib/LoginServiceKit/LoginServiceKit.swift).

What was the motivation to do this?

Also, how can you be using `LoginServiceKit` without having it listed in the `Podfile`?

docs/Contributors.md
```diff
@@ -2,24 +2,6 @@
 
 They helped [develop the app](https://github.com/lwouis/alt-tab-macos/graphs/contributors):
 
-* [adamnemecek](https://github.com/adamnemecek)
```
lwouis, line 5: This happens if you run `scripts/update_contributors.sh` without setting `$GITHUB_TOKEN`. You can simply remove [this line](https://github.com/lwouis/alt-tab-macos/blob/master/scripts/update_contributors.sh#L16) before running the script (and change the `\` on the previous line with `|`) to make the github call without a token. See the comment above the line for more explanations on why the token is needed in the first place

lwouis:
@mnin I've finished reviewing your PR. Please check out my comments 👍

Podfile.lock
MoNTE48: The new version is always (almost) better than the old one. Nothing should break from this. 🤓

lwouis: Better how?

I recall having issues with upgrading cocoapods in the past, as the CI would then be desync'd. Now the project cocoapods is aligned with the CI version and unless there is a clear benefit, I'd rather not touch things

mnin:
> @mnin I've finished reviewing your PR. Please check out my comments 👍

@lwouis thanks, were you able to test the build of Travis?

lwouis:
> @lwouis thanks, were you able to test the build of Travis?

I'm sorry, I'm not sure what you mean here?

Travis builds are temporary. They get persisted only for releases on the `master` branch. They get persisted as [Github releases assets](https://github.com/lwouis/alt-tab-macos/releases) then.

Regarding Apple Silicon, as I've said in the ticket and the other PR, I don't own a AS machine. I only have an Intel macbook pro. However, people interested in having an AS/universal builds said that your build (as well as the other PR's build) are working for them, and also that it is passes the compiler, probably it will run.

The PR Travis build has passed on this PR, so that's a good sign. I'm not sure what you're looking for in your question. Could you please elaborate?

Thank you

mnin:
> > @lwouis thanks, were you able to test the build of Travis?
> 
> I'm sorry, I'm not sure what you mean here?
> 
> Travis builds are temporary. They get persisted only for releases on the `master` branch. They get persisted as [Github releases assets](https://github.com/lwouis/alt-tab-macos/releases) then. So, if what you mean here is "could run the CI build from this PR on your local machine?", then the answer is no. I can't access it as it's deleted after the CI job is done.
> 
> Regarding Apple Silicon, as I've said in the ticket and the other PR, I don't own a AS machine. I only have an Intel macbook pro. However, people interested in having an AS/universal builds said that your build (as well as the other PR's build) are working for them, and also that it is passes the compiler, probably it will run.
> 
> The PR Travis build has passed on this PR, so that's a good sign. I'm not sure what you're looking for in your question. Could you please elaborate?
> 
> Thank you

Hey, I'm interested of course if the build works and actually includes both architectures, I thought that was understandable before doing anything else here.

Like:

```shell
file /Applications/AltTab.app/Contents/MacOS/AltTab
/Applications/AltTab.app/Contents/MacOS/AltTab: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64]
/Applications/AltTab.app/Contents/MacOS/AltTab (for architecture x86_64): Mach-O 64-bit executable x86_64
/Applications/AltTab.app/Contents/MacOS/AltTab (for architecture arm64): Mach-O 64-bit executable arm64
```

Podfile.lock
mnin: To use the current/latest version of Cocoapods I think is self-evident and not worth mentioning here. And obviously this version also worked on Travis without any apparent problems.

src/ui/preferences-window/tabs/GeneralTab.swift
mnin: I used the Swift Package Manager to include the dependency. And I think, all other dependencies should use the Swift Package Manager too (in the future).

And I used this dependency to solve an issue in the code https://github.com/lwouis/alt-tab-macos/blob/master/src/ui/preferences-window/tabs/GeneralTab.swift#L32 for now.

This API can be removed soon by Apple. But I didn't want to remove any functionality through this PR.

src/ui/preferences-window/tabs/AppearanceTab.swift
mnin: This line caused me to be unable to open the Appearance Tab without creating a crash.

In the crash log there were errors regarding the UI constraints. But no more details what exactly triggers this behaviour.

After that I had to debug for a while to get further here. I realised that `LabelAndControl` and this line are related.

After removing this line, there were no more crashes and the UI looked like before.

I suspect here is a deviation in Apple's UI API that if in `LabelAndControl` `makeLabelWithProvidedControl` leaves `labelPosition` on the default `.leftWithSeparator`, then there is probably a bug with the constraints.

Podfile
mnin: Cool, thanks, there was a new comment on your PR 7 days ago https://github.com/sparkle-project/Sparkle/pull/1938#issuecomment-912801855

If nothing else comes up for the time being, I would then switch to your version.

docs/Contributors.md
mnin: I assumed if I built a PR, it would have a chance to run on Travis once. I removed the contributors so I could have a successful RUN there once. I didn't see that info, I will try that out, thanks.

src/ui/preferences-window/tabs/GeneralTab.swift
lwouis: > And I used this dependency to solve an issue in the code

As I said above, the library you introduced uses the same Apple APIs as AltTab uses currently. Thus your change doesn't improve on the issue you perceived

Regarding the deprecation, nothing can be done for now. Apple marked it as deprecated, but the alternative introduced since are different and not really replacement. I've spent hours understand the various APIs, their history, their pros and cons before landing on the current implementation.

When the day comes (if it comes) when Apple removes it, then we will have ample time to find a creative solution to replace it. For now this API is the best implementation in my opinion.

But again, here we are discussing a library that uses the same exact deprecated APIs, so it's literally replacing the API calls by a third-party lib that wraps the same API calls.

docs/Contributors.md
lwouis: > I assumed if I built a PR, it would have a chance to run on Travis once.

Yes, PRs on this repo trigger a CI build. You can see the builds generated by this PR here: https://github.com/lwouis/alt-tab-macos/pull/1094/checks. You can then access travis itself by clicking the links, and seeing the logs. It's all open to anyone to peruse.

> I removed the contributors so I could have a successful RUN there once

It's a tricky situation for PRs. There are scripts like `scripts/update_contributors.sh` that exist to make sure that contributors are credited; that adding them to the list is not forgotten. That runs on PRs as well. It can be annoying because unrelated to the PR. On the other hand, if we disable those for PRs, then someone could add a new library and forget to credit it, or share commits from someone else and forget to credit them. Same for localization scripts. They make sure people are aware and do the right things when introducing new strings in the app which will need to be localized (easy to forget, especially for first-time contributors)

My flow here is to run the `scripts/update_contributors.sh` locally with the token line removed, then the contributors are up to date and I can move on.

If you have any idea to improve this, please feel free to share

src/ui/preferences-window/tabs/GeneralTab.swift
mnin: Hey, I already said that with `This API can be removed soon by Apple.`.

But the current implementation just doesn't work and I didn't want to invest much time there, since LoginServiceKit also runs with macOS 11.5 without problems.

Screenshot 2021-09-10 at 15 51 00

docs/Contributors.md
mnin: That's very honorable in principle, but just doesn't meet my expectation when I create a PR that then never build the binary. Because I don't want to put my name in there unless I could build a successful build on Travis with the setup.

lwouis:
> And yes, we can just run file on Travis, but I asked for a way to get the binary out of the build.

As far as I know, Travis deletes any file in the workspace after a job. For a file to persist, it has to be manually deployed somewhere, like what's done today with releases being stored on Github Releases.

If you know of a way to keep CI builds, at least for a while, on Travis somehow, please do share 👍

Podfile
lwouis: My PR, if merged, would only fix `master`. Then they would need to release v1.28. Then we would need to move to that, and I think it has breaking change, so we would need to work on that.

I don't know the Sparkle project enough to be sure, but I suspect they don't support fixes on old releases. Something like v1.24.1 releases today, which is what we want ideally. I think it's simpler to use the branch on my fork which is exactly that: v1.24 + the fix = v1.24.1. That way we get Universal Build support without the popups bug, and without having to deal with the breaking changes introduces after v1.24

docs/Contributors.md
lwouis: How would you suggest we improve this?

src/ui/preferences-window/tabs/GeneralTab.swift
mnin: Again, I know we're talking about a long deprecated API here. Which is why I don't want to waste more time here.

And the current implementation has a bug, the external library does not have that, what is the idea here? Fix the current implementation? Is there a reason why the few lines must not come from a dependency?

We'll have to remove that soon anyway.

docs/Contributors.md
mnin: I would suggest doing it only on the `master` branch.

src/ui/preferences-window/tabs/GeneralTab.swift
lwouis: > And the current implementation has a bug, the external library does not have that, what is the idea here? Fix the current implementation? Is there a reason why the few lines must not come from a dependency?

I think it's interesting to understand the root cause of the bug you found on macOS 11. The API is still present on that system, so why is it crashing? What's the difference between the code-path in the implementation vs the code-path in the library code? It may reveal another issue actually. I'm interested in understanding why suddenly an issue appeared in an area of code that has been stable and working for a long time now.

> We'll have to remove that soon anyway.

It's been deprecated for 6 years. That puts the "soon" in perspective I think.

rahul0705:
Just wondering when this branch will get merged in? With the M1 Pro and M1 Max laptops being released. I suspect there will be a larger influx of developers who would love to see an M1 native version of this app.

lwouis:
@rahul0705 there are some concerns regarding the PR (e.g. https://github.com/lwouis/alt-tab-macos/pull/1094#discussion_r706550231) that i would like to see addressed before it's merged

src/ui/preferences-window/tabs/GeneralTab.swift
ayroblu: Couple things about this by the way, I found that this wasn't building for me on both master, and this pr
master on Xcode 13.1:
image

On master this is fixable by adding a few nil handlers. But on this branch I couldn't fix it due to the reliance on the package. imo, this package seems unnecessary, I found I was having a crash on startup, and so I commented out `kLSSharedFileListItemBeforeFirst` and then things worked as expected. I don't see this line in LoginServiceKit, so I think it seems reasonable to remove this line @lwouis ? And that way you don't need to download a whole new package to handle all this

lwouis: > I found that this wasn't building for me on both master, and this pr master on Xcode 13.1:

Do you know why there was a change (from compiling to no-longer compiling)? What's the new error with `nil` here?

> I don't see this line in LoginServiceKit, so I think it seems reasonable to remove this line @lwouis ?

Do you mean this line?

```
let _ = LSSharedFileListInsertItemURL(loginItems, kLSSharedFileListItemBeforeFirst.takeRetainedValue(), nil, nil, itemUrl, nil, nil).takeRetainedValue()
```

This line is the most important of the method. It adds the login item. It can't be removed otherwise we lose the start-at-login functionality. Or am I misunderstanding what you meant?

ayroblu: > > I found that this wasn't building for me on both master, and this pr master on Xcode 13.1:
> 
> 
> 
> Do you know why there was a change (from compiling to no-longer compiling)? What's the new error with `nil` here?
> 

No clue but I suspect that apple forces updates on you when you update Xcode?
> 
> 
> > I don't see this line in LoginServiceKit, so I think it seems reasonable to remove this line @lwouis ?
> 
> 
> 
> Do you mean this line?
> 
> 
> 
> ```
> 
> let _ = LSSharedFileListInsertItemURL(loginItems, kLSSharedFileListItemBeforeFirst.takeRetainedValue(), nil, nil, itemUrl, nil, nil).takeRetainedValue()
> 
> ```
> 
> 
> 
> This line is the most important of the method. It adds the login item. It can't be removed otherwise we lose the start-at-login functionality. Or am I misunderstanding what you meant?

WELL haha, that's the line that crashes every single time for me, I wonder how LoginServiceKit and others do it

lwouis: I [shared above](https://github.com/lwouis/alt-tab-macos/pull/1094#discussion_r706216085) LoginServiceKit implementation, which is only around 50 lines, which are basically the same as what I did. They have some guard closes though, which mean they may just do nothing sometimes if some call goes wrong. Which means silently failing for the user.

I would like to see someone going to the bottom of this, and understanding why it's apparently crashing on BigSur. The APIs are still there AFAIK, so what is going on?

Podfile
mnin: @lwouis which version should I take now? The MR is closed, shall we use the beta v2 version of Sparkle?

src/ui/preferences-window/tabs/GeneralTab.swift
mnin: Sorry but I didn't get the idea behind here, why should not use the LoginServiceKit, it does the same and it's fixed for the current macOS version? Why re implement the same?

mnin:
> > And yes, we can just run file on Travis, but I asked for a way to get the binary out of the build.
> 
> As far as I know, Travis deletes any file in the workspace after a job. For a file to persist, it has to be manually deployed somewhere, like what's done today with releases being stored on Github Releases.
> 
> If you know of a way to keep CI builds, at least for a while, on Travis somehow, please do share 👍

Just trying https://transfer.sh/ to get the build out of travis, let's see in the build log...

mnin:
The build on travis doesn't contain the arm64 executable:

```shell
** BUILD SUCCEEDED **
+ls -lah /Users/travis/build/lwouis/alt-tab-macos/DerivedData/Build/Products/Debug/AltTab.app/Contents/MacOS
total 8144
drwxr-xr-x  3 travis  staff    96B Nov 10 22:48 .
drwxr-xr-x  8 travis  staff   256B Nov 10 22:48 ..
-rwxr-xr-x  1 travis  staff   4.0M Nov 10 22:48 AltTab
+file /Users/travis/build/lwouis/alt-tab-macos/DerivedData/Build/Products/Debug/AltTab.app/Contents/MacOS/AltTab
/Users/travis/build/lwouis/alt-tab-macos/DerivedData/Build/Products/Debug/AltTab.app/Contents/MacOS/AltTab: Mach-O 64-bit executable x86_64
```

Will just do some testing again to see, if we can create the arm64 build on Travis in general, please ignore my last commits for now.

mnin:
@lwouis I need some help here:

```shell
error: No signing certificate "Mac Development" found: No "Mac Development" signing certificate matching team ID "M3YVS85XC2" with a private key was found. (in target 'alt-tab-macos' from project 'alt-tab-macos')
```

dvcrn:
@mnin were you able to build this with xcode13.1? Tried to give your branch a go here on Monterey but running into a bunch of Swift compiler errors

mnin:
> @mnin were you able to build this with xcode13.1? Tried to give your branch a go here on Monterey but running into a bunch of Swift compiler errors

Yes, just pull my latest commits which removes the general tab for testing purposes.

dvcrn:
Great, that worked, I was able to build it here! Running into other issues with constant crashes but at least it's compiling now 😄

mnin:
> Great, that worked, I was able to build it here! Running into other issues with constant crashes but at least it's compiling now 😄

As I wrote, these changes are only for building the arm64 executable and the x86_64 on Travis, not for running somewhere. Please use the build here https://github.com/lwouis/alt-tab-macos/pull/771#issuecomment-942341213 which is compiled for both architectures and with all features running fine.

src/ui/preferences-window/tabs/GeneralTab.swift
mnin: Hey @lwouis @ayroblu,

I replaced `LoginServiceKit` with the `LaunchAtLogin` swift package, which uses the Helper solution / `SMLoginItemSetEnabled` / `ServiceManagement` which works really well.

Please review this part again.

lwouis: I evaluated ServiceManagement during the initial implementation. I think of it as a [regression](https://github.com/sindresorhus/LaunchAtLogin#my-app-doesnt-show-up-in-system-preferences--users--groups--login-items) over the previous approaches. Currently, the user can manage their Login Items in all transparence. They could script or use apps to manage their Login Items, and AltTab is a part of that, making it accessible. With ServiceManagement, apps are no longer part of the Login Items standard approach, and users can't manage "login at launch" in the standard central place.

There are also other possible regressions. I know that the current implem is smart enough to [handle multiple versions](https://github.com/sindresorhus/LaunchAtLogin#my-app-doesnt-launch-at-login-when-testing) for instance.

I also worry about notarization. I've had numerous problems with it in the past, when introducing new frameworks for example. You can't test for it in the PR as well; it would only fail during CD on `master` as this is the only time/place where we use my paid Apple Developer account to notarize the app.

I'd prefer staying with the old API implem which I have tested for many many hours, and I know handles exotic edge cases. It's also light (few lines of code vs embedding an helper-app). If the `LoginServiceKit` library truly worked, then it should be pretty easy to peruse its implementation and adjust the current implem here to work on macOS 11.

This would be a path of least risk/change, with no regressions in terms of UX, app size, rare use-cases, etc.

lwouis:
Can someone clarify for me if the crashes/issues with `kLSSharedFileListItemBeforeFirst` are starting with macOS 11? starting with macOS 12? when building for AppleSilicon?

Podfile
lwouis: They [closed by MR indeed](https://github.com/sparkle-project/Sparkle/pull/1938), and mentioned that the fixes may be included in future releases.

In September, I prepared a fork of Sparkle 1.24.0, which is AppleSilicon compatible, with the fixes needed by AltTab cherry picked. It is probably the best approach for now: https://github.com/lwouis/Sparkle/blob/fix-iframe-popup-1.24/Sparkle.podspec

lwouis:
Sharing here the issue with the Login-at-startup API for people who may be interested:

I used Xcode 13 on macOS 12, and multiple compilation errors that don't exist on Xcode 12 on macOS 10.15. The reason is that some APIs were changed

Here is the API on Xcode 12:

```swift
public func LSSharedFileListCreate(_ inAllocator: CFAllocator!, _ inListType: CFString!, _ listOptions: CFTypeRef!) -> Unmanaged<LSSharedFileList>!
```

And here is the API on Xcode 13:

```swift
public func LSSharedFileListCreate(_ inAllocator: CFAllocator?, _ inListType: CFString, _ listOptions: CFTypeRef?) -> Unmanaged<LSSharedFileList>?
```

The optionality of the API calls have changed. This explains why the third-party library was checking and exiting some of these calls early.

I'll look into this further.

lwouis:
I still don't understand how SDK work for macOS. According to [this page](https://developer.apple.com/support/xcode/), Xcode 12 which I use on my machine shouldn't be able to make apps for macOS 12, as it doesn't contain the macOS 12 SDK yet. Indeed I can verify this locally:

```bash
$ xcodebuild -showsdks
macOS SDKs:
	macOS 11.1                    	-sdk macosx11.1
```

Yet, I can run a .app build from Xcode 12 on my macOS 12 VM. I guess it's some kind of forward-compatibility where if you don't use any new API, and you got lucky that the old ones are the same, it just runs?

In any case, I also build and ran the app from Xcode 13 on the macOS 12 VM. Because the macOS 12 SDK has changed the optionality inside `CoreServices.SharedFileList`, I added `!` after the 3 API calls (see diff above). When I ran it, it works fine.

So it seems that the crash that people reported is specific to AppleSilicon arch, as the API works well on any version of macOS, on Intel, compiled from any SDK.

src/ui/preferences-window/tabs/GeneralTab.swift
mnin: I know, I saw the issue. And yes, I also know that what Apple offers is not ideal, not even for the user.

But unfortunately this is the only supported way to start an app automatically when logging in.

And yes, introducing more things might lead to more problems.

I had implemented the helper myself first, if that helps, I can undo that as well. However, this framework looks like it is exactly what you want.

mnin: But in fact, all the apps are getting rid of `LSSharedFileListInsertItemURL`. Also Apple is supporting `SMLoginItemSetEnabled` for Sandboxed apps.

Podfile
mnin: Is it fixed in the beta version 2?

lwouis:
@mnin I was able to debug on an M1 VM (thanks @mnin!), and here is how the runtime crash happens:

image

I thought that maybe the memory model had changed, so tried `takeUnretainedValue` but same crash:

image

I also found out that using `kLSSharedFileListItemLast` instead of `kLSSharedFileListItemBeforeFirst` crashes on my Intel mac! It seems these 2 virtual item references are bugged.

I tried to replace this virtual item by the last item in the Login Items, and that works on the M1 VM!

image

The issue is that some people have 0 Login Item in their list, thus it would not work for them. At least it shows that the API is still available on the M1 arch. It's just a bug with these virtual item constants.

I'll see if I can work around it.

lwouis:
I've looked into workarounds for the runtime bug in the macOS API (i.e. `kLSSharedFileListItemBeforeFirst `) on M1.

I've looked into using:

* **`defaults`** to set Login Items. It seems that it [used to be possible](http://hints.macworld.com/article.php?story=20111226075701552) to control Login Items using the `defaults` binary. I tried many variations, but I think `loginwindow AutoLaunchedApplicationDictionary` is now yet another abandoned facility on macOS.
* **Apple Events** to set Login Items. Here is are some findings:

While the `LSSharedFileList` API has been deprecated for a while, the same functionality is available through Apple Events, and thought AE is clearly a forgotten side-note in Apple APIs, it's been stable and available on all versions macOS since the early days.

Of course, it's quite a mess to achieve it. After hours of investigation, I was able to get it running in a nice and maintainable shape. I used `OSAKit` to execute a JXA (JavaScript variant of AppleScript) script which I added to the app's bundle resources.

Now this approach has one major downside: to do its job, the script needs to send an AE to `System Events`, which requires the permission in `System Preferences > Security & Privacy > Privacy > Automation` to be granted. This would be yet another permission we would need to ask the user. However this one is not critical for AltTab to run. It's just to make it able to toggle itself from the Login Items. So we would need to handle it differently than the current wall-of-permissions popup that's shown on first launch.

Sharing here this [fantastic post explaining in great details how AE permissions work](https://www.felix-schwarz.org/blog/2018/08/new-apple-event-apis-in-macos-mojave), and even critiquing it and showing its many limitations.

---

@mnin I can say I tried my best, and I'm giving up on the Login Items. Dealing with the new AppleEvents permission would be more UX work for us, and I suspect a worse experience for users anyway.

I've looked into your MR, and how you added the `LaunchAtLogin` package. I'll review it in-depth now (e.g. size of the helper, compatibility with older macOS, CI pipeline, etc).

lwouis:
@mnin this `LaunchAtLogin` is pretty rough. They embed a 12MB helper. To compare, AltTab is currently 21MB:

* Around 12MB of swift runtime. No way to get rid of it, unfortunately. It's distributed with macOS from macOS 15 onwards, but I want to support previous versions, so we keep it (see #158)
* Sparkle is 4MB, and way worth it. These guys did a wonderful job, and the package is packed with features and localizations
* ShortcutRecorder is 1MB, and way worth it. Great job there also.
* 1.6MB of images and localization. Everything is carefully, manually compressed here.

The fact that their helper is 12MB just because they wrote its 20 lines in Swift is a bit preposterous. I'll look if there are slimmer helpers out there

ayroblu:
Hmm is this a Monterrey issue? I think it's interesting cause previous builds of this have worked fine for everyone, is issue specific to new Xcode etc, or does this issue affect old AltTab releases too?

lwouis:
@ayroblu the M1 VM I investigated on was on Monterey. So yeah it's possible that M1 + Big Sur doesn't have this runtime bug.

Anyway, I've decided to go with a LaunchAgent implementation. It's pretty simple (roughly the same 30 lines of implementation as the previous API) and compatible. I've tested it on the M1 VM, and it works (see https://github.com/lwouis/alt-tab-macos/pull/771#issuecomment-970418912)

src/ui/preferences-window/tabs/AppearanceTab.swift
lwouis: I tested with ad without, and I couldn't see any difference. So since it fixed it for you, let's bring that change forward.

Also, I'm wondering if it could actually help with https://github.com/lwouis/alt-tab-macos/issues/1062

mnin:
Close this PR in favour of #1239

Issue #760, opened 2021-01-02
Fade out animation has a white flash starting with Big Sur


Pythomiere:
![alttab](https://user-images.githubusercontent.com/69103204/103455216-1d995b00-4ca0-11eb-8a52-17fcca8797ff.gif)

lwouis:
I'm not sure what's happening on your gif. There is no distinction between Dark and Light mode for AltTab.

It looks like this on my machine, in Dark then Light mode:

https://user-images.githubusercontent.com/106195/103475659-9ad4e500-4df2-11eb-9b1c-319921495558.mov

You can see the smooth fade-out animation.

Some questions that may help diagnose the issue:

* Are you on Big Sur maybe?
* What happens in Light mode?
* Did it always act like this, or is this happening after an update of macOS or AltTab?

Pythomiere:
It will only happen in a dark background.

https://user-images.githubusercontent.com/69103204/103488884-3f403280-4dc5-11eb-90f4-f63cc55f807f.mov

lwouis:
The animation is handled entirely by macOS (by AppKit, technically), so I'm not sure what I can do about this. It's not reproducible on my machine as well.

* Are you on macOS Big Sur (macOS 11)?
* Does it happen with the default AltTab preferences? You can reset preferences by running this command in the Terminal: `defaults delete com.lwouis.alt-tab-macos`

Pythomiere:
I am in Big Sur now. I set up a new account and run with the default setting but the problem still occurs, I do not think this is the setting's problem. 

<img width="686" alt="Screen Shot 2021-01-04 at 1 01 43 PM" src="https://user-images.githubusercontent.com/69103204/103579808-db357100-4e8d-11eb-9342-23fa0e78e5ab.png">
<img width="686" alt="Screen Shot 2021-01-04 at 1 01 57 PM" src="https://user-images.githubusercontent.com/69103204/103579817-dcff3480-4e8d-11eb-8215-a7a0363c9340.png">

lwouis:
The preferences on your screenshot are heavily customized. Could you try with the default preferences and only activating the fade out animation?

I'm trying to diagnose if it's a Big Sur issue, or maybe your preferences, especially relative to layout, or maybe both.

Pythomiere:
Yeah, it still occurs even if I use the guest account with default settings.

awefuiawef:
[...]

PR #1201, opened 2021-11-03
Use a custom fade out animation for Big Sur and later

zacharee:
This should fix https://github.com/lwouis/alt-tab-macos/issues/760.

On macOS 11 and 12, the default utility window animation in AppKit doesn't respect dark mode, so the window flashes white when fading out. This fix replaces the AppKit animation on 11 and 12 with a simple alpha fade animation to work around the white flash.

As an effect of the custom animation, there's now a fade-in animation as well.


lwouis:
> As an effect of the custom animation, there's now a fade-in animation as well.

Do you think you could work out a way to only have fade-out? A strength of AltTab is how fast is shows when summoned. Some people enjoy a fade out animation, but i don't think anyone ever asked for a fade-in. People have always asked for less latency / quicker display though, historically. In addition, the current preference is labelled specifically "Fade out animation" so it would become incorrect on macOS 11+ 

I'm on holidays at the moment. When i'm back in a few days i'll review this PR and test it on my macOS 11 and macOS 12 VMs

src/ui/main-window/ThumbnailsPanel.swift
```diff
@@ -35,12 +37,34 @@ class ThumbnailsPanel: NSPanel, NSWindowDelegate {
...
+            )
+        } else {
+            super.orderOut(sender)
+        }
+    }
+
     func updateFadeOutAnimation() {
-        animationBehavior = Preferences.fadeOutAnimation ? .utilityWindow : .none
+        if #available(OSX 11, *) {
+            alphaValue = (Preferences.fadeOutAnimation && !isVisible) ? 0 : 1
```
lwouis, line 57: I don't understand the check on `isVisible` here. `updateFadeOutAnimation` is called when the preference has been changed by the user, not every time the window is shown/hidden.

I'm not sure why we need to update `alphaValue` in the first place actually here. It seems that it should only be updated on `show` and on `orderOut` (by the animator), no?

zacharee: `updateFadeOutAnimation()` is also called in the constructor, so this sets an initial state. When the preference changes, the app also needs to make sure that the alpha value is correct. I think the other logic was mostly for having it fade in before I removed that (to have the proper state), but it shouldn't cause any harm.

isametry:
If I may ask, what is the progress on this merge?

lwouis:
@isametry this has to be tested more. It's not easy because these animations behave differently in every version, and even within the same major macOS version, it differs. In addition, when I do the testing, it's inside a VM, so animation suffer in performance, and I can't judge about the quality of an animation for instance, with the same precision I can judge on my local machine (macOS 10.15).

So it's about QA now, and making sure we ship something solid that doesn't break (this PR [introduces some concurrency](https://github.com/lwouis/alt-tab-macos/pull/1201/files#diff-20286aee98cdeb923ac692044b9b829963acdff75f32a8f2fa4dca635fd498b7R46) that I'm not so sure can't crash the app), and that looks good.

isametry:
@lwouis thanks for letting me know. I am running the latest macOS (on Apple Silicon) and I would love to help with any tests – unfortunately I’m not a developer so I can’t really do a lot of dabbling myself, but whenever there’s any builds to try out, don’t hesitate to ping (I assume in #760 rather than here)

> You mentioned this yourself on the relevant ticket
I didn’t realize this PR was the same as that testing build, sorry for that.

PR #1206, opened 2021-11-07
Check for Android Studio windows similarly to JetBrains

zacharee:
Android Studio is a fork of JetBrains IntelliJ IDEA, so it has the same issue where tooltip windows are detected as normal windows. Since it has a different bundle ID, the existing JetBrains check doesn't work. This adds a check that does the same thing as the JetBrains check, just using the Android Studio bundle ID template.

The wildcard regex at the end is also needed, since the Canary versions of Android Studio have `-EAP` appended to the bundle ID.

### DIFF
--- a/src/api-wrappers/AXUIElement.swift
+++ b/src/api-wrappers/AXUIElement.swift
@@ -41,11 +41,11 @@ extension AXUIElement {
 
     func axCallWhichCanThrow<T>(_ result: AXError, _ successValue: inout T) throws -> T? {
         switch result {
-            case .success: return successValue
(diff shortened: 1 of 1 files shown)


lwouis:
Thanks for sharing this PR! I've taken the liberty to slightly alter the code. I've committed with your name/email, in order to credit your work.

The fix will ship next time I merge the `develop` branch on `master`. See this PR here: https://github.com/lwouis/alt-tab-macos/pull/1208/files

Closing this PR as I'll wait for a new changes before merging to master. (It triggers a release automatically)

Issue #563, opened 2020-08-28
AltTab's window sometimes "hangs" for 0.5 to 1.5 sec after choosing a window or cancelling


mfn:
Version: 6.0.0
**Describe the bug**
I did not notice this with the version before 6 I was using and I just upgraded.

I would say 1 out of 5 times, when I choose (keyboard or mouse) or not choose to select a window (ESC), the over of the thumbs hangs around a bit longer before it disappears (i.e. it's not instant).

Hard to properly _reproduce_, it just happens sometimes.

- _Maybe_ it's after some time _not_ bringing up AltTab and then quickly choosing a window. That is what just happened while I wrote the issue and quickly wanted to switch somewhere else.
- Yep, same happened again. I finished above sentence and _then_ I could immediately reproduce it again

I use "RecordIT" and tried to capture it, but it doesn't properly reflect the behaviour due to low FPS :/

----
![jbETlzsvkC](https://user-images.githubusercontent.com/87493/91526906-fc48ec80-e904-11ea-97d0-beb8631c5c5e.gif)

mfn:
I see everyone using loom or so, but it needs registering AFAICS which I didn't want to use. I use the free version of RecordIT (due to low FPS) but that's _usually_ enough for me.

If anyone can recommend another screen recording app which doesn't need registering and produces better results, I'm open to suggestions!

lwouis:
Hey again @mfn! Thanks for sharing this issue!


I noticed this during development for 2 days. I thought I was going crazy because it's so inconsistent like why is it happening? I tried to capture it in Instruments to profile it, but it only happened once there, and there was basically no activity in AltTab, so I thought "ok it's the WindowServer being slow to respond for some reason, nothing I can do about that".

It's either some multi-threading issue, or OS interaction like stressing out the WindowServer in burst or something. I'm sad to hear it's happening on other people's computer too. I thought it was only on my system. I'll try to work at it, somehow.

For the video software. I used OBS in the past. It records in top quality, but then you have the issue of where to host the content to share it. Loom is incredibly plug-and-play. Yes you need an account, but then it's a one-click to record and share with others. They pay the video hosting, so I feel it's fair to sign up. They are trying to run a business, and their product is good. Any free software is going to put the burden of hosting the video on you.

lwouis:
Questions to help me investigate:

* Do you remember these freezes happening on previous releases, or is it a new thing?
* Do you have BetterTouchTool running?

[...]

PR #1211, opened 2021-11-08
Make sure kAXFocusedUIElementChangedNotification is only allowed when the trigger PID isn't alt-tab's

zacharee:
This seems to fix https://github.com/lwouis/alt-tab-macos/issues/563.

I noticed that `voiceOverFocusedWindow()` was being called twice whenever the focused window would change, and traced it back to `kAXFocusedUIElementChangedNotification` being triggered and acted upon when the focused window changed. 

I modified the conditional when handling accessibility events to disallow `kAXFocusedUIElementChangedNotification` from running unless the triggering PID isn't alt-tab's. I've done some testing and it seems to fix the freezing issue without any other side effects.

### DIFF
--- a/src/logic/events/AccessibilityEvents.swift
+++ b/src/logic/events/AccessibilityEvents.swift
@@ -11,8 +11,7 @@ fileprivate func handleEvent(_ type: String, _ element: AXUIElement) throws {
     debugPrint("Accessibility event", type, type != kAXFocusedUIElementChangedNotification ? (try element.title() ?? "nil") : "nil")
     // events are handled concurrently, thus we check that the app is still running
     if let pid = try element.pid(),
-       try ((pid != ProcessInfo.processInfo.processIdentifier && element.subrole() != kAXUnknownSubrole)
-               || (type != kAXFocusedUIElementChangedNotification)) {
+       try (pid != ProcessInfo.processInfo.processIdentifier || (element.subrole() != kAXUnknownSubrole && type != kAXFocusedUIElementChangedNotification)) {
         switch type {
             case kAXApplicationActivatedNotification: try applicationActivated(element, pid)
             case kAXApplicationHiddenNotification,
--- a/src/logic/events/AccessibilityEvents.swift
+++ b/src/logic/events/AccessibilityEvents.swift
@@ -11,7 +11,8 @@ fileprivate func handleEvent(_ type: String, _ element: AXUIElement) throws {
     debugPrint("Accessibility event", type, type != kAXFocusedUIElementChangedNotification ? (try element.title() ?? "nil") : "nil")
     // events are handled concurrently, thus we check that the app is still running
     if let pid = try element.pid(),
-       try (!(pid == ProcessInfo.processInfo.processIdentifier && element.subrole() == kAXUnknownSubrole)) {
+       try ((pid != ProcessInfo.processInfo.processIdentifier && element.subrole() != kAXUnknownSubrole)
+               || (type != kAXFocusedUIElementChangedNotification)) {
         switch type {
             case kAXApplicationActivatedNotification: try applicationActivated(element, pid)
             case kAXApplicationHiddenNotification,


lwouis:
> I noticed that voiceOverFocusedWindow() was being called twice whenever the focused window would change

There is a guard at the beginning of `voiceOverFocusedWindow` though: https://github.com/lwouis/alt-tab-macos/blob/98c144efe8a2d61419bece8d0d650537fd588329/src/logic/Windows.swift#L102

Debugging shows that it passes the guard only once per action in my tests. Once when I summon AltTab, once when I move to the next window. It doesn't call it at all on focusing windows by releasing AltTab or with the mouse.

Regarding the conditional you mention here https://github.com/lwouis/alt-tab-macos/issues/563#issuecomment-962697594, the logic is the following:

We treat events if:

(they don't come from AltTab) OR (they come from AltTab AND they are not `kAXUnknownSubrole`)

(AltTab main window has subrole `kAXUnknownSubrole` [on purpose to mark it](https://github.com/lwouis/alt-tab-macos/blob/98c144efe8a2d61419bece8d0d650537fd588329/src/ui/main-window/ThumbnailsPanel.swift#L23), so we can ignore it in the accessibility events)

which can be simplified to:

(they don't come from AltTab) OR (they are not `kAXUnknownSubrole`)

it's also equivalent to NOT (they come from AltTab AND they are not `kAXUnknownSubrole`), which is the current form.

So I think the current logic is correct.

You also added a check on `type != kAXFocusedUIElementChangedNotification`. What is your intention here? You added this with an `OR` which means you are handling _more_ events, not fewer. I can't quite get what the intent is there

zacharee:
For me, the guard was passing twice. I added a `debugPrint()` after it and got two successive prints in the log. I'm not sure what exactly is going on with the conditional, but alt-tab's window was firing the focus event with the original conditional. The new conditional has the check on the focus event to make sure it doesn't run if it's that event and the PID is the same. Otherwise, it lets the event propagate, which I found out was needed to update the list when the alt-tab preference window is added or removed.

To reproduce the freezing, you need something intensive going on, like a Zoom call or streamed video. Then just spam the switcher shortcut until it freezes. Sometimes it will also freeze if you wait a second before letting go of the shortcut.

zacharee:
It's one of those coding things I can't fully explain, but I've been running with that change for over 12 hours now, and I haven't seen any freezes.

lwouis:
> It's one of those coding things I can't fully explain, but I've been running with that change for over 12 hours now, and I haven't seen any freezes.

I mean that sounds great! I'm just a bit worry of the combo of (A) i can't observe it on my machine + (B) i don't understand it theoretically by looking at the code change.

Could you maybe share a local build with that PR patch on the ticket, and ask other participant to beta-test it? I've done this a lot in the past for fixes that i can't test on my machine. Usually i tag people who've shown interest, and they do help 👍

src/logic/events/AccessibilityEvents.swift
```diff
@@ -11,7 +11,8 @@ fileprivate func handleEvent(_ type: String, _ element: AXUIElement) throws {
     debugPrint("Accessibility event", type, type != kAXFocusedUIElementChangedNotification ? (try element.title() ?? "nil") : "nil")
     // events are handled concurrently, thus we check that the app is still running
     if let pid = try element.pid(),
-       try (!(pid == ProcessInfo.processInfo.processIdentifier && element.subrole() == kAXUnknownSubrole)) {
+       try ((pid != ProcessInfo.processInfo.processIdentifier && element.subrole() != kAXUnknownSubrole)
```
lwouis, line 14: In the ticket, people have reported that this change fixes the issue for them. While I'm *really* excited about these reports, I'm struggling to see how this change fixes it.

Previous code truth table:

Event is from AltTab | window subrole is unknown | event is focusedUIElement | process event?
-|-|-|-
true | true | true | no
true | true | false | no
true | false | true | yes
true | false | false | yes
false | true | true | yes
false | true | false | yes
false | false | true | yes
false | false | false | yes

New code truth table (changes in bold):

Event is from AltTab | window subrole is unknown | event is focusedUIElement | process event?
-|-|-|-
true | true | true | no
true | true | false | **yes**
true | false | true | **no**
true | false | false | yes
false | true | true | **no**
false | true | false | yes
false | false | true | yes
false | false | false | yes

If I didn't make a mistake in the tables, the three cases that changed are:

* Handling previously excluded events:
  * From AltTab, window subrole is unknown, event is not focusedUIElement
    * => It seems to me that we don't want to handle any of these events. They are from AT itself, from its main window. The **main window events should all be ignored**
* Not handling previously handled events:
  * event is focusedUIElement, from AltTab, window subrole is not unknown
    * => AT doesn't use tabs, so it's fine to ignore focusedUIElement events on its non-unknown windows. **I'm wondering if this one is not the key!** I'm thinking to test a build that only prevents this.
  * event is focusedUIElement, not from AltTab, window subrole is unknown
    * => I don't think we want to handle events from unknown subrole windows. I think this will **lead to issues** with some apps.

Could you please test a build with this condition? It changes only the case which I think is desirable, while not introducing the 2 cases which I think are not desirable. It may be that this change there is the key to fix the ticket. I can't test it on my machine, so could you? 🙇‍♂️

```swift
pid != ProcessInfo.processInfo.processIdentifier || element.subrole() == kAXUnknownSubrole || type != kAXFocusedUIElementChangedNotification
```

zacharee: Testing with this new conditional now.

zacharee: `voiceOverFocusedWindow()` gets called twice when the window first appears, which I think is because of `element.subrole() == kAXUnknownSubrole`. It doesn't seem to cause the same performance issues, and the method only gets called once per focus event afterwards.

lwouis: I'm sorry, I shared the wrong conditional. I got confused after making the tables I think. Here's the correct conditional that fixes the issue with `kAXFocusedUIElementChangedNotification` without introducing regressions, I believe:

```swift
pid != ProcessInfo.processInfo.processIdentifier || (element.subrole() != kAXUnknownSubrole && type != kAXFocusedUIElementChangedNotification)
```

I tested it yesterday and I triggered only one call to `voiceOverFocusedWindow` when summoning AT.

zacharee: I'll add a new build to the issue.

lwouis:
From the encouraging results in https://github.com/lwouis/alt-tab-macos/issues/563#issuecomment-966461192, I'm taking this change in the `develop` PR. I committed with your ID to credit you for the amazing fix of course.

If it fully fixes https://github.com/lwouis/alt-tab-macos/issues/563, I will be in your debt truly. This freeze/lag issue has been weighting on me for a while, and I couldn't find the root of it for months.

Let's hope it's over. Thank you again for your wonderful contribution!

Closing this PR (fix will be bundled in the `develop` PR)

PR #1214, opened 2021-11-09
Allow switching to Parallels windows

zacharee:
This should fix https://github.com/lwouis/alt-tab-macos/issues/1213.

I'm noticing a weird issue where switching between two Parallels windows when they're adjacent in the focus order only works if you very quickly press and release the switcher shortcut. Otherwise, the focus doesn't change. I'm guessing from the code comments that that's an OS bug.

### DIFF
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -157,25 +157,17 @@ class Window {
             // but quickly switches back to another window in that space
             // You can reproduce this buggy behaviour by clicking on the dock icon, proving it's an OS bug
             BackgroundWork.accessibilityCommandsQueue.asyncWithCap { [weak self] in
-                do {
-                    guard let self = self else { return }
-                    var elementConnection = UInt32(0)
-                    CGSGetWindowOwner(cgsMainConnectionId, self.cgWindowId, &elementConnection)
-                    var psn = ProcessSerialNumber()
-                    CGSGetConnectionPSN(elementConnection, &psn)
+                guard let self = self else { return }
+                var elementConnection = UInt32(0)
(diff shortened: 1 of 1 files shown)


lwouis:
From the other ticket, you wrote:

> It looks like CGSGetWindowOwner() doesn't set elementConnection to anything. The returned CGError has a status of 0, so it think it's succeeding.
>
> Since elementConnection stays 0, CGSGetConnectionPSN() returns a CGError with status -600, which is apparently procNotFound (which makes sense).

What you're saying is that `CGSGetWindowOwner` is unreliable. In this PR, you use `GetProcessForPID` to get the PSN from a PID. If that way is reliable, then let's replace the current approach with `GetProcessForPID`.

In this PR you use `GetProcessForPID` as a fallback, but if it's reliable, let's just replace the implementation with it. In the past, the PID was trickier to obtain, but now it's always available in `self.application.pid`.

How about this?

```swift
guard let self = self else { return }
var psn = ProcessSerialNumber()
GetProcessForPID(application.pid, &psn)
_SLPSSetFrontProcessWithOptions(&psn, self.cgWindowId, .userGenerated)
self.makeKeyWindow(psn)
self.axUiElement.focusWindow()
```

zacharee:
I wasn't sure if just GetProcessForPID was reliable, but it seems like it is, so I've replaced CGSGetConnectionPSN completely.

lwouis:
I've taken the fix, and added it to https://github.com/lwouis/alt-tab-macos/pull/1208 as a commit with you as the author. This will it will be part of the next release. Closing this PR. Thank you!

Issue #522, opened 2020-08-11
Controls tab layout broken on macOS 10.12 Sierra


lwouis:
**Describe the bug**

On macOS 10.12 Sierra, the app has 2 issues:

* The app icon isn't shown properly  
  ![image](https://user-images.githubusercontent.com/106195/89922072-8290dd80-dc39-11ea-9238-a13bf5aa5eea.png)
  Note that the icon is correct in the permission panel:  
  ![image](https://user-images.githubusercontent.com/106195/89922180-a6ecba00-dc39-11ea-80ea-4a09a488404a.png)

* The Controls tab layout is broken  
  ![image](https://user-images.githubusercontent.com/106195/89921933-54130280-dc39-11ea-8a31-19f5caa9b24c.png)

lwouis:
Here are the things I tried to get the app icon to work on Sierra:

* Generate the `.icns` with `iconutil` to test if this wasn't a regression of #169 
* Use `CFBundleIconFile` with the full name with extension `app-icon.icns`
* Use `CFBundleIconFile` with the default name `AppIcon`
* Use `CFBundleIconName` instead or in addition to `CFBundleIconFile`

I don't get it. I downloaded AppCleaner and Rectangle to compare how other apps do it, and then do literally the same way, yet their icons display.

lwouis:
I can confirm that the layout issue in the Preferences > Controls is still there in 6.7.0 on Sierra.

lwouis:
Still happening on 6.21.2

lwouis:
Closing this as subsumed by #351

Issue #351, opened 2020-05-22
Simplifying the preferences UI


lwouis:
I was responding on #337, and realized it's been a theme recently, and I feel the need to take some time to steer this project. Here is the situation as I see it. This is a copy/paste of my answer of the ticket I just mentioned:

AltTab is a port of Windows 10. As such, it shows windows in chronological order. While I believe it to be the best general approach for productivity, I'm open to introducing other workflows to the app. People have suggested direct access for instance (#210).

The issue however with adding more workflows into the same tool, is the explosion of preferences. The tool already has a large number of preferences because I find myself wanting to let people have it their way, and be empowered. The downside of that is that the tool is becoming unwelcoming/overwhelming, especially to new and more casual users.

Let's imagine we add a new preference to group the windows by app. That's yet one more preference. Now in your example (Pic 5), you say that the windows for each app could be ordered based on the screen they are on. Some people may prefer that they be in chronological order, instead of based on which screen. Actually if you have multiple windows of the same app on the same screen, you need to sort them anyway, and chronological seems the way to go. My point being that we are already talking about adding 2 new preferences here.

Another issue is that I decided our of principles to have 0 telemetry in the app, website, etc. This means that I have 0 statistics on users preferences, usage pattern, etc, outside of the active people like you who take the time to discuss on the Issues here. I think everyone would agree that adding a preference to accommodate 1 person is not the best idea. Preferences should be useful to a bunch of people. However as I said there is no way to know even if current preferences in the app are of any use.

Really, I'm not sure how to pilot this project forwards without data. I all falls down to my judgment in the end since I'm the main contributor. It's not necessarily a bad thing btw to not follow data. Data is why fast food in the US is so big. They optimize demand and human feedback for caloric-rich food, and now they have a terrible obesity problem. I rather like the idea of not always optimizing for real-world usage as people's habits may not be the direction we should reinforce.

I'm inviting everyone to feel welcome to share here their vision on where the project should go, in terms of features, flexibility, UX of the preferences, richness vs complexity, etc.

borna761:
I really appreciate you taking the time to think this through before just going ahead and implementing all the suggestions and ideas that people come up with.

My suggestion would be to perhaps have a toggle of a simple/advanced mode? The simple mode could have the current functionality (or perhaps even some of it removed to make it even simpler?), and the advanced mode could have the extra functionality that has been asked for.

[...]

PR #1223, opened 2021-11-12
Add spacing between preference tabs

zacharee:
This is a minor cosmetic change to make the preference tabs look a little nicer. 

I also wanted to add some padding to the toolbar items themselves, but I can't figure that out (and it seems like macOS should be doing it anyway).

image

### DIFF
--- a/docs/Contributors.md
+++ b/docs/Contributors.md
@@ -60,6 +60,7 @@ They helped [localize the app](https://poeditor.com/join/project/8AOEZ0eAZE):
 * Julian Nowaczyk
 * Kagurazaka Tsuki
 * kal
+* kant
 * Kevinsevinche
 * Klara
 * Kushnee5
--- a/resources/l10n/es.lproj/Localizable.strings
+++ b/resources/l10n/es.lproj/Localizable.strings
@@ -96,7 +96,7 @@
 "Close window" = "Cerrar ventana";
 
 /*No comment provided by engineer.*/
-"Conflicting shortcut" = "Conflicting shortcut";
+"Conflicting shortcut" = "Atajo conflictivo";
 
 /*No comment provided by engineer.*/
 "Controls" = "Controles";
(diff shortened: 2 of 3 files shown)


lwouis:
I must confess, I'm not immediately convinced that this is an improvement. It feels like the design with extra margin is not standard. It's recommended nowhere on [the HIG](https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/toolbars/). It makes the app look different from other apps, and I don't see clearly the benefits. It also makes the code implementation have more code, and it could introduce layout regressions. We already have some layout issues:

* https://github.com/lwouis/alt-tab-macos/issues/522
* https://github.com/lwouis/alt-tab-macos/issues/1062

Be careful as well that usually you should let standard components handle themselves, so they can adapt to the various relooking of macOS versions. By that I mean that your change may look broken in another version of macOS where the toolbars look fairly different.

Overall, I feel sorry but I don't see the benefits clearly, so I don't think I could merge this change

zacharee:
It's similar to the Xcode layout, although it looks like Xcode has fixed spacing between tabs and flexible spacing on the edges.

image

lwouis:
I'm not sure. There is a lot of testing to do on the various macOS versions. Like how about this line here: https://github.com/lwouis/alt-tab-macos/blob/9847038f26cb9098d2af6ec0b31f55ca94333d99/src/ui/preferences-window/PreferencesWindow.swift#L32

I have never properly tested AltTab on macOS 11 or 12 as I'm still on 10.15, so there are a lot I'm not fully aware of, like the new design direction of the OSes, compatibility, maybe changes in the way we should call the APIs, maybe new dark mode stuff, etc.

zacharee:
Maybe it would make sense to redo the toolbar entirely.

lwouis:
> Maybe it would make sense to redo the toolbar entirely.

What do you mean exactly?

zacharee:
> > Maybe it would make sense to redo the toolbar entirely.
> 
> What do you mean exactly? Do you mean rework the preferences overall, as in the scope of #351?

It could probably be part of that.

lwouis:
It's been a year without activity. I'm closing the PR 🙇

Issue #1265, opened 2021-11-30
Make AppCenter integration opt-in


Rjevski:
**Describe the bug**

The application contacts `in.appcenter.ms` on first run. While I haven't examined the exact data transferred, it's most likely checking in and registering a persistent ID that it'll use for future telemetry/crash reports. This inherently includes the user's IP address and if a persistent ID is indeed used then would allow AppCenter to have a log of the user's IPs. This should be opt-in and having it phone home immediately (without any warning nor disclosure) may be a breach of the GDPR.

**Screenshots / video**

<img width="566" alt="image" src="https://user-images.githubusercontent.com/45625150/144025766-304d3840-9dc2-4c6e-89b4-9e3417366b99.png">

**Steps to reproduce the bug**

1. Monitor network traffic
2. Open alt-tab-macos
3. See spyware phoning home

Rjevski:
I see that there is a setting that's available once the app is configured and accessibility permissions are granted. It does do the right thing and defaults to "ask whether to send crash reports". The AppCenter registration should be delayed until the first crash is encountered and the user explicitly opts-in.

lwouis:
*On the form of your message*

I didn't know that AppCenter was making network calls on setup. I don't know what is being exchanged. As you've seen, I've designed AltTab to be as unintrusive as possible. I don't appreciate you attributing malice on my part. I've made a strong stance with this app to have 0 telemetry, use as little CPU/memory/disk/battery/etc as possible. I've spent days optimizing images so the bundle size is small, using observers instead of polling so the battery usage is as low as possible, dealing with hundreds of poorly written tickets from people who don't know how to fill a bug report, which I could have approached by having telemetry but didn't, etc.

> breach of the GDPR

This is FOSS I do on my weekends. This is not a commercial endeavour. I benefit in no way from this. I've been refusing donations for years. I'm not a lawyer, but my opinion is that there would never be any ground for litigation here. It's like going to your local homeless shelter and criticizing the volunteers for having the "free food" sign show a picture of a burger. "The burger on the sign is not exactly the food served; it's false advertising". The allegation is quite misplaced in that context.

> See spyware phoning home

[...]

PR #1266, opened 2021-11-30
Make AppCenter truly opt-in

Rjevski:
This PR addresses my earlier issue where AppCenter was noticed to be phoning home despite the crash reporting setting defaulting to "ask". This uses a [new feature](https://docs.microsoft.com/en-us/appcenter/sdk/other-apis/ios#disallow-network-requests) in the AppCenter SDK 4.3.0, so as a result the pod has been updated as well.

I haven't yet tried to actually build this (need to get Apple to sort out my developer account situation - my personal Apple ID seems to be associated with an expired _corporate_ dev account) so leaving in draft for now.


Podfile
```diff
@@ -6,5 +6,5 @@ target 'alt-tab-macos' do
   pod 'Sparkle', :podspec => 'https://raw.githubusercontent.com/lwouis/Sparkle/fix-iframe-popup-1.24/Sparkle.podspec'
   pod 'ShortcutRecorder', :git => 'https://github.com/Kentzo/ShortcutRecorder.git', :commit => 'f718220854575bad86fa89dd2bb3decd0b14b2b0'
   pod 'SwiftyMarkdown', '1.1.0'
-  pod 'AppCenter/Crashes', '4.1.0'
+  pod 'AppCenter/Crashes', '4.3.0'
```
Rjevski, line 9: Updated the version.

src/logic/AppCenterCrashes.swift
```diff
@@ -4,41 +4,26 @@ import AppCenterCrashes
 
 class AppCenterCrash: NSObject, CrashesDelegate {
```
Rjevski, line 5: Here are the key changes.

src/logic/AppCenterCrashes.swift
```diff
@@ -4,41 +4,26 @@ import AppCenterCrashes
...
     static let secret = Bundle.main.object(forInfoDictionaryKey: "AppCenterSecret") as! String
-
-    override init() {
-        super.init()
+    // we need to call this again if the setting changes so make it a function
+    func setDefaultNetworkRequestsPolicy(override: Bool = false) {
+        // this will default to false thus not sending anything until opted-in
+        // override is to override temporarily on a case-by-case basis
+        AppCenter.networkRequestsAllowed = override || Preferences.crashPolicy == .always
+    }
```
Rjevski, line 12: We may need to call this twice so made this a function. If the user's preference is to "ask", this will restrict network communications by default, so we need to call it again with the override flag when the user clicks "send" in the modal following a crash to temporarily allow network requests.

src/logic/AppCenterCrashes.swift
```diff
@@ -4,41 +4,26 @@ import AppCenterCrashes
...
-                    buttons[id].state = .on
-                }
-                Preferences.set("crashPolicy", String(id))
-                BackgroundWork.crashReportsQueue.async { Crashes.notify(with: userChoice == .alertFirstButtonReturn ? .send : .dontSend) }
-            } else {
-                BackgroundWork.crashReportsQueue.async { Crashes.notify(with: Preferences.crashPolicy == .always ? .send : .dontSend) }
-            }
-            return true
-        }
+        Crashes.userConfirmationHandler = self.confirmationHandler
```
Rjevski, line 24: Moved this to a dedicated function on the class.

src/logic/AppCenterCrashes.swift
```diff
@@ -48,7 +33,50 @@ class AppCenterCrash: NSObject, CrashesDelegate {
...
+            
+            let id = self.crashButtonIdToUpdate(userChoice, checkbox)
+            if let buttons = PoliciesTab.crashButtons, buttons.count > id {
+                buttons[id].state = .on
+            }
+            
+            Preferences.set("crashPolicy", String(id))
+            let shouldSend = userChoice == .alertFirstButtonReturn
+            
+            self.setDefaultNetworkRequestsPolicy(override: shouldSend)
```
Rjevski, line 65: In this case, we temporarily (for this session) allow network access.

Rjevski:
This should address #1265.

lwouis:
I have cherry-picked and edited the commit a bit. It's mostly details except for [this change here](https://github.com/lwouis/alt-tab-macos/pull/1269/commits/6fad1296c2660c5581e326df805d2e3136547a4f?authenticity_token=slYW11ciAVgdQPB6K9Y7XMaXeXzuDVfbBsIVbwBwWoqFNBwZKpHDX6oQTYZiE6BlUY1ZwmWbx6AmCbgGrjNpoQ%3D%3D&file-filters%5B%5D=.swift#diff-349f196bf3ae688a3b262c81395c5d01ac218d729738566a0d8fe6c40db669b0R36). I basically surrounded the line which sends the report with `network ON > send report > network OFF`.

The idea is that we don't want AppCenter to network after a report has been sent. You haven't noticed this pattern in your original ticket, but it's possible that it also networks on a schedule, or who knows. We want network calls *only* when necessary to send the report, so we shut networking off before and after.

Beyond the code, it took me a while to test, but I think everything works. That is:

* No networking outside of sending report
* Critically: reports are being sent successfully

I was worried the reports may have some issues as this scenario is probably used by very few users given that the flag is very recent (only on the latest version of AppCenter) and that there is 0 ticket/discussion around the flag on their github repo. In addition, they have coded the flag setter in such a way that it actually logs an error. In the setter, they check a `canBeUsed` flag, which is only set after `.start` has been called: https://github.com/microsoft/appcenter-sdk-apple/blob/fa8c7d353a4f7819763816a026666320f7315d45/AppCenter/AppCenter/MSACAppCenter.m#L582. However we want to set it before, to prevent `.start` from triggering network calls, so it's a bit of a catch-22. Anyway, it seems to work despite the error being logged. I hope there are no unexpected side-effect though :/

I'll close this PR as I've subsumed the work into the `develop` branch. Of course you'll remain the commit author and will be credited on the website and in the app 👍

PR #1350, opened 2022-01-29
feat: allow cursor follow focus behavior

GrzegorzKazana:
Addresses https://github.com/lwouis/alt-tab-macos/issues/1087

Adds user preference that makes the cursor follow the focus after _alt-tabbing_. In order to not confuse the users, this option is disabled by default.

Main motivation was integration with other shortcuts (like `⌃ + ←` or `⌃ + → ` in _mission control_) which work in context of the display on which the cursor is (regardless of focus). 

The solution also works across multiple displays, but I didn't know how to screen capture that :sweat_smile: 

image
Screen Recording 2022-01-29 at 18 10 39

### DIFF
--- a/src/ui/App.swift
+++ b/src/ui/App.swift
@@ -189,10 +189,26 @@ class App: AppCenterApplication, NSApplicationDelegate {
         KeyRepeatTimer.toggleRepeatingKeyPreviousWindow()
     }
 
-    func focusSelectedWindow(_ window: Window?) {
+    func focusSelectedWindow(_ selectedWindow: Window?) {
         hideUi()
         guard !CGWindow.isMissionControlActive() else { return }
-        window?.focus()
+        guard let window = selectedWindow else { return }
+
+        window.focus()
+        if Preferences.cursorFollowFocusEnabled {
+            moveCursorToSelectedWindow(window)
+        }
+    }
+
+    func moveCursorToSelectedWindow(_ window: Window) {
+        guard let position = window.position, let size = window.size else { return }
+
+        let point = CGPoint(
+            x: position.x + size.width / 2,
+            y: position.y + size.height / 2
+        )
+
+        CGWarpMouseCursorPosition(point)
     }
 
     func reopenUi() {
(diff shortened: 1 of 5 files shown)


src/ui/App.swift
```diff
@@ -189,10 +189,26 @@ class App: AppCenterApplication, NSApplicationDelegate {
         KeyRepeatTimer.toggleRepeatingKeyPreviousWindow()
     }
 
-    func focusSelectedWindow(_ window: Window?) {
+    func focusSelectedWindow(_ selectedWindow: Window?) {
         hideUi()
         guard !CGWindow.isMissionControlActive() else { return }
-        window?.focus()
+        guard let window = selectedWindow else { return }
```
lwouis, line 195: Actually, this guard could be moved above `CGWindow.isMissionControlActive`. It's a bit expansive to check if Mission Control is active, so better first check if the window exists as that check is cheap.

It's not directly related to the PR, but the PR touching that area of code could be an opportunity to improve the code/performance here 👍

lwouis:
Thank you for sharing this PR! It's very neat and was easy to review!

The implementation is straightforward and seems robust. That being said, there are always corner-cases on people's computer like people having multiple cursors or who knows what, so I'm suspecting possible issues with `CGWarpMouseCursorPosition`. That being said, it's an opt-in preference, so worst case is that the user turns it off.

I was happily surprised to see that it works with minimized and hidden windows for instance. `window.position` returns the future position whereas I would expect it to return `nil` since the window is not onscreen yet. These are example of edge cases that can be surprising and introduce bugs.

It looks good to me overall, so I'll cherry-pick the commit and ship a release soon. You'll be credited in the app and on the website for your contribution 👍

lwouis:
Closing as I'll include this PR in another PR which will trigger a release via CD.

src/ui/App.swift
lwouis: I'll make this little improvement as an edit to the commit, and I'll commit it under your username

GrzegorzKazana:
Cool 🚀 
Thanks for constructive feedback and taking care of mentioned adjustments.

LiamKearn:
Issue fixed by restart. 😄

lwouis:
@GrzegorzKazana FYI, there is a bug with the implementation. See https://github.com/lwouis/alt-tab-macos/issues/1087#issuecomment-1115952528

Issue #1249, opened 2021-11-24
JetBrains IntelliJ IDEA and other IDEs disappearing from alttab


dvcrn:
**Describe the bug**

Continuation from https://github.com/lwouis/alt-tab-macos/issues/824

I have repeatedly the issue that jetbrains stuff just vanishes from the alt-tab after longer working sessions. Not doing any fullscreen settings here and keep it in a normal window all the time. 

~Only after restarting IDEA, it shows back up. Restarting alt-tab does not help.~ Restarting either alt-tab or IDEA fixes this

IDEA and other apps are installed through jetbrains toolbox

**Screenshots / video**

<img width="686" alt="Screen Shot 2021-11-24 at 11 09 23" src="https://user-images.githubusercontent.com/688326/143158693-e9504a09-7822-4d19-979d-abf141381601.png">

<img width="686" alt="Screen Shot 2021-11-24 at 11 09 24" src="https://user-images.githubusercontent.com/688326/143158671-eb7b608d-d032-4141-a8e1-ac0537fae11c.png">

**Steps to reproduce the bug**

I'm not actually sure. I tested this with CLion, IDEA Ultimate and Goland, and they all disappear after a longer working session (a couple hours)

lwouis:
> Restarting alt-tab does not help.

This is interesting/unexpected. usually issues with windows missing from AltTab are because AltTab may have missed an event or there was a transition. Here it doesn't show after a restart.

It would seem to indicate that the windows not shown on AltTab may be considered "not real windows". Could you please show screenshots/videos of exactly which windows are not showing but you think should show. The best would be a video where you show the app windows, AltTab thumbnails, and narrate what you think is wrong

Dimezis:
I have the same problem both with IntelliJ IDEA and Android Studio (basically the same thing).
Except restarting alt-tab always helps in my case.
Basically, if the IDE was started after alt-tab, it's never shown. If it's started before alt-tab, it's being detected just fine.

dvcrn:
Thanks for chiming in! I just tried it again and yu are right, restarting alt-tab does help. Sorry for misreporting this. 

JetBrains stuff is getting sorted at the end of the alt-tab switcher so I didn't notice it was in the list. So yeah I confirm what @Dimezis said, restarting alt-tab does fix this

lwouis:
Can one of you please share a video so i can see which windows are having this issue?

[...]

Issue #1079, opened 2021-08-19
Sometimes firefox is hidden in tab switcher


nrjais:
**Describe the bug**

Sometimes(Faced it 3-4 time over a span of 2 months) firefox window is not shown in the tab switcher and the only way to get it back is to restart alt-tab

**Screenshots / video**
<img width="1680" alt="Screenshot 2021-08-19 at 12 55 28 PM" src="https://user-images.githubusercontent.com/20153307/130026207-3ca99efa-a038-4e03-9972-b64d5b89b50a.png">

**Steps to reproduce the bug**

I haven't been able to reproduce this, It randomly happens, but if you need anything in the current state I can provide, But once I restart alt-tab the issue will go away and re-appear after 2-3 weeks

lwouis:
Is there any chance that this window on your screenshot started fullscreen, or was set fullscreen at some point in time? There are known issues with fullscreen windows, so i'm wondering if this is within that scope or if it's another root cause

nrjais:
No, The Zoom window is full-screen. It has always happened with Firefox only, and it randomly happens, but I think whenever this has occurred a full-screen zoom window was open, but I don't remember if it is always the case.

lwouis:
I need steps to reproduce if we want any chance to fix this 😅

Which version of macOS are you on?

nrjais:
Mojave
I will be changing my laptop after two days, Then if it happens again will try to reproduce.
It is just very rare that I can't get the steps to reach to this state knowingly.

fredless:
Volunteering that I'm seeing similar behavior with Microsoft Edge, using multiple desktops and Show Windows from Visible Desktops selected for the shortcut.  Sometimes none of the Edge windows on the active desktop are shown... after manually clicking through the open windows, and then trying again, they show up in the switcher. Happening frequently, but not precisely sure how to reproduce reliably yet.  When I figure it out I will open a new issue. Big Sure 11.5.2 and Edge v95

lwouis:
Should be fixed in v6.32.0

Issue #1179, opened 2021-10-19
I'm looking for someone to take over this project 🤝 


lwouis:
Hi everyone!

## The project so far

I've been working on AltTab for more than 5 years now. This project has proven popular:

<img width="3664" height="2616" alt="Image" src="https://github.com/user-attachments/assets/ad22ec50-2d73-4b57-a061-eda23fc3fbbc" />

It's hard to put an exact number on the active users count, because the app has no user tracking (very much on purpose). But I imagine it's in the thousands. I based that on crash reports (on AppCenter) and [releases download stats](https://tooomm.github.io/github-release-stats/?username=lwouis&repository=alt-tab-macos).

## The future

I have decided to allocate less time to this project. The project is quite easy to jump into. I've taken time to [document steps to understand the codebase, the tools, etc](https://alt-tab-macos.netlify.app/contributing). A [lot of people](https://alt-tab-macos.netlify.app/thanks) have contributed localizations over time, but [few people](https://alt-tab-macos.netlify.app/thanks) have contributed significant code changes.

I'm looking for someone interested in taking over this project and become the main maintainer. We can discuss the hand-over. I can walk you through everything, from the codebase, to different tools and services (e.g. AppCenter, Travis, Homebrew, CI, Apple Certificates, etc).

### If you're interested, please [contact me](mailto:lwouis@gmail.com)

Cheers~

fennecs:
thanks!

dresl:
Thank you for doing this. It's awesome!

prasskaly:
Best app for macos. Can't go back now. 
Thank you for this.

milomylo:
nice project

XInTheDark:
Thanks for all the effort you have put into this free app ❤️

siempay:
Did you get any one to maintain this project ?

lwouis:
@siempay no-one has talked to me about it yet, unfortunately.

siempay:
If I get you to walk me throu and explain to me the project and what are you expecting from the maintainer to be doing for the future of the project, I would be interested

imikushin:
It's a great product! 

I'd be glad to buy it on the AppStore and even throw in a yearly subscription to pay for continued maintenance (version upgrades) of this app. Just an idea.

lwouis:
[...]

PR #1437, opened 2022-03-17
fix issue #1249 #1079 #1392

metacodes:
I am able to reproduce the issue by reinstall JetBrains AppCode. The first time I open AppCode after reinstall it, AltTab never show AppCode window. I debug the code on my Mac, I find that AltTab can get the event of app launched but never get the event of finishedLaunching. So I just remove the code runningApplication.isFinishedLaunching. After that, AltTab work well and the issue is fixed.

The following picture is the debug log when I reproduce this issue:
image

### DIFF
--- a/src/logic/Application.swift
+++ b/src/logic/Application.swift
@@ -72,7 +72,7 @@ class Application: NSObject {
     }
 
     func addAndObserveWindows() {
-        if runningApplication.isFinishedLaunching && runningApplication.activationPolicy != .prohibited && axUiElement == nil {
+        if runningApplication.activationPolicy != .prohibited && axUiElement == nil {
             axUiElement = AXUIElementCreateApplication(pid)
             AXObserverCreate(pid, axObserverCallback, &axObserver)
             debugPrint("Adding app", pid ?? "nil", runningApplication.bundleIdentifier ?? "nil")


lwouis:
@metacodes thanks for sharing this

The check is there for a reason: so that we know that the app is finished launching. This is the signal for us to ask for its windows and subscribe to its events.

It's likely that removing this check would break things. Could you please look at the git history for examples of issues the check was mitigating?

I advice you test with a variety of apps like Gimps for example which is slow to launch. See if your change has not broken things there.

metacodes:
@lwouis That makes sense.

I have look at the git history and I found that you added these code to solve some background processes. But in the function observeNewWindows there also has runningApplication.isFinishedLaunching, so I think maybe remove the code runningApplication.isFinishedLaunching in the function addAndObserveWindows is safe. We just need to run observeEvents() to work around the issue they never send the finishedLaunching event.
image

I try to test with a variety of apps. You said Gimps. Do you mean this app(https://www.gimp.org)? I have already test it, it seems work fine. I will test this code for a few days to see if the change has not broken things. I will report the result after that.

lwouis:
It's possible that checking for `isFinishedLaunching` is not necessary before subscribing to the app events, yes. This [comment I left here](https://github.com/lwouis/alt-tab-macos/pull/1437/files#diff-45acb5e8fb8c3382ebc34954179b288dd6f23b96745d43c3c6c519f729ac2527L160) actually says that we don't even trust this boolean flag at the end of the day, and rely on a subscription to actually complete to consider the app "launched". That's why we have this ridiculous boolean flag `isReallyFinishedLaunching`.

So yes, maybe your removal of the check in that specific spot is safe. However, better test it carefully, as it could break the app for everyone.

Regarding testing, yes I was referring to [Gimp](https://www.gimp.org/), sorry about the typo. Furthermore, I looked into past tickets, and here are some apps I think are worth testing with your change:

* [Bear.app](https://bear.app/)
* [Octave.app](https://octave-app.org/)
* [Protege](https://protege.stanford.edu/products.php#desktop-protege)
* Books.app (available by default on macOS)

metacodes:
@lwouis Currently, I have tested Bear.app, Protege, Books.app. Only Bear.app has some problems, sometimes AltTab still can not show Bear's window. But AltTab received the event isFinishedLaunching. The problem is the axWindows_.count gets 0. When I add sleep(3) before that, axWindows_.count will get the right count of windows and AltTab shows Bear's window. I guess there are some problems with the function retryAxCallUntilTimeout, but I don't know how to solve this.
Here is the debug code.
image

Could you give me some advice?

metacodes:
@lwouis I have added some codes to fix Bear.app's issue. Its scenario is different from other apps(Octave.app, Protege.app, Books.app, Firefox.app, JetBrains' app). You can find more details about this issue in my commit comments. For now, these apps(Gimp.app, Bear.app, Octave.app, Protege.app, Books.app, Firefox.app, JetBrains' app) work well after testing. And it seems no impact to other apps which are work well.
image

metacodes:
#1439 I have also tested [Sublime Merge.app](https://www.sublimemerge.com) . Everything works well.

metacodes:
After a week of testing, everything is fine. As Apple says that some applications do not post this notification and so are never reported as finished launching, I think it's reasonable to remove the check of isFinishedLaunching before we create a new observer that can receive notifications from the specified application.
image

lwouis:
I'll try and test/merge this as soon as possible. I'm struggling to find time these days (see https://github.com/lwouis/alt-tab-macos/issues/1179). Please forgive the delays. Hopefully i can finish this soon

src/logic/Application.swift
```diff
@@ -117,6 +117,11 @@ class Application: NSObject {
                     if group == nil && !self.wasLaunchedBeforeAltTab && CGSSpaceGetType(cgsMainConnectionId, Spaces.currentSpaceId) == .fullscreen {
                         throw AxError.runtimeError
                     }
+                    // workaround: sometimes some apps are launched and are actived but OS returns the windows.count == 0. we wait for a moment
+                    // for example: Bear.app
+                    if group == nil && !self.wasLaunchedBeforeAltTab && self.runningApplication.isActive {
```
lwouis, line 122: I'm wondering if we need `self.runningApplication.isActive` here. An app could be launching in the background, and still benefit from these retries, no?

The way I think of it is: an app may launch but not have its windows exposed instantly. In that case we want to retry until the 5s timeout.

Note that it may be bad to waste CPU/battery on retrying for 5s, but the case of an app that launches but has no window to show seems very unlikely. So in most cases, I expect the retries to be a good thing, and make the windows visible within a short delay. Whereas the cases where the app actually has no window to show and waste 5s of CPU/battery should be rare.

metacodes: I agree with you. I have also done some tests base on your changes. But there are two problem. 

Firstly, the variable wasLaunchedBeforeAltTab is not initialized correctly when AltTab was first launched. The value should be true for all of the apps launched before AltTab. But I found that the value always false. That cause AltTab throws lots of AxError.runtimeError when it was launching and the CPU usage was very high(20%~50%). I will add some code to fix this issue.
image

Secondly, there are some apps that have menu bar only and launched without window. In my case, it's a screenshot app called iShot.app. The app launches but has no window. When I use it to capture the screen, the AltTab throws lots of AxError.runtimeError.
image

Base on that two scenario mentioned above, I think we can't simplify both if with a single one although it's very unlikely there are really no window. We need to make sure that the app really has window, so that the retries will be a good thing.

metacodes: I have already fixed the issue that the variable wasLaunchedBeforeAltTab is not initialized correctly.

metacodes: You are right. I accept the retry option. In my mind, I just want to reduce the case we retry so that we can reduce CPU usage. But if we add isActive check here, we may miss some windows as some scenarios that we did not expect.
So I accept your changes here.
Do I need to change the code? Or you can change and merge it directly?

metacodes: > We could keep the status quo here to be honest, and these apps fix themselves, instead of penalizing every alttab users with more resource usage.

Frankly, I came to modify the code because the window loss problem has been bothering me for a long time, and I tried to find other alternative applications, but nothing met my needs, so I finally took the plunge and learned Swift and try to solve the problem. 

I don't think these problematic apps will be fixed in future versions. Because for them, they probably won't be aware of these problems without AltTab users giving them feedback, and even if they do, they'll probably think it's not a functional problem and ignore it. For AltTab users, they don't think it's a problem with other apps, they probably think it's a functional bug in AltTab.

So personally, I'd like to see a compromise solution to solve these app bugs, maybe even OS bugs, although these compromise solutions will have some maintenance costs and even negative effects in the future. It depends on how we balance the two. But for now we can control the scope of this negative impact, such as the processing of specific applications, although these special codes are annoying in terms of engineering structure and code aesthetics, and these codes will also incur some maintenance costs in the future.

lwouis:
I've embedded these commits in the next release. I committed them slightly altered, but with your author name/email, so you're properly credited. You will also be listed as a contributor on the website and in the app 👍  Closing this PR

metacodes:
Cool 🚀
Thanks for constructive feedback and taking care of mentioned adjustments.

lwouis:
@metacodes it seems that my worries about CPU usage were founded :/ I woke up to these tickets, after yesterday new release:

* https://github.com/lwouis/alt-tab-macos/issues/1481
* https://github.com/lwouis/alt-tab-macos/issues/1477
* https://github.com/lwouis/alt-tab-macos/issues/1476

metacodes:
@lwouis sorry about that.😭 I will try to solve this issue as soon as possible.

lwouis:
@metacodes no worries! This is volunteer work. We don't have a QA team to avoid these scenarios. It's unfortunate, but it is expected given the limited resources we have. So no need to feel guilty or anything.

Now if you find a mitigation for this, it's a win for everyone ;)

metacodes:
@lwouis I found that the high cpu usage was due to the fact that we kept trying to subscribe to notifications for certain processes, but they were background programs with no windows. (e.g. com.apple.universalcontrol, org.mozilla.plugincontainer)
image

For each such process, AltTab would spend 120s to retry. The whole process is actually doing useless work. But for those programs that take a long time to load, this retry is necessary again. I think it is not necessary to keep retrying all the time, we can let the program take a break and retry again, although in rare cases there may be a slight delay in the creation of the window. 
image
image

Retries don't happen very often, and even in the worst case, there is only a 1s delay in displaying the window, which I think is acceptable.
`
func retryAxCallUntilTimeout_(_ group: DispatchGroup?, _ timeoutInSeconds: Double, _ fn: @escaping (	) throws -> Void, _ startTime: DispatchTime = DispatchTime.now()) {
    do {
        try fn()
        group?.leave()
    } catch {
        Thread.sleep(forTimeInterval: 1) // sleep for 1s to reduce CPU usage
        let timePassedInSeconds = Double(DispatchTime.now().uptimeNanoseconds - startTime.uptimeNanoseconds) / 1_000_000_000
        if timePassedInSeconds < timeoutInSeconds {
            BackgroundWork.axCallsQueue.asyncAfter(deadline: .now() + .milliseconds(10)) {
                retryAxCallUntilTimeout_(group, timeoutInSeconds, fn, startTime)
            }
        }
    }
}
`

What do you think?

lwouis:
Increasing the time between retries mitigates the issue, but is not really addressing the issue directly.

As i understand you, it seems that `com.apple.universalcontrol` never succeeds in subscribing. Is there some metadata/flag on that process that we could use to tell that it will never succeed to subscribe? Maybe the error code from the subscription call is an indication? Anything to be able to discriminate these processes that can never succeed?

metacodes:
> Maybe the error code from the subscription call is an indication?
@lwouis I've considered it, but currently these processes return the value `cannotComplete`, which is the same as those that have windows but load very slowly. I will continue to try to find some other methods based on your suggestion.

metacodes:
@lwouis I have consulted other window management tools and found some useful methods. I referenced the codes and submitted a new PR(#1484).

- Amethyst

https://github.com/ianyh/Amethyst/blob/cf7c35553acdac09fdb71f3eb86e3a5d7263f040/Amethyst/Managers/WindowManager.swift#L285

https://github.com/ianyh/Amethyst/blob/cf7c35553acdac09fdb71f3eb86e3a5d7263f040/Amethyst/Categories/NSRunningApplication%2BManageable.swift#L40

https://github.com/ianyh/Silica/blob/2e9f8aa635d8162e17a7cec9babd11f07cc7d4dc/Silica/NSRunningApplication%2BSilica.m#L10

- yabai

This discussion https://github.com/koekeishiya/yabai/issues/439#issuecomment-597716114

metacodes:
Some `LSUIElement` app has windows (e.g. menu bar app). For those apps' window will disappear. Further research is needed.

metacodes:
> Some `LSUIElement` app has windows (e.g. menu bar app). For those apps' window will disappear. Further research is needed.

I found it interesting that Apple's default application switcher doesn't show these windows either.

Issue #1481, opened 2022-04-09
High CPU usage on 6.33.0


litori:
AltTab 6.33.0 causes high cpu usage on AltTab and launchd process.  Nonstop for 1 hour now.  Killing AltTab gets rid of launchd cpu usage.

metacodes:
@litori Could you share what apps you were using when this issue happened? I'm trying to reproduce this issue.

litori:
I closed all my apps and just have activity monitor on and launchd, logd, and alttab will be the top 3 cpu usage hovering around 20 for launchd, 5 for logd, and 8 for alttab.  Once I open discord and firefox without any websites, launchd will be at 55, alttab at 30, and logd at 27.

panpietrek:
Same here with v. 6.33.0
Apps open are Firefox (v99), Activity Monitor and Signal Desktop 5.38. Activity Monitor show AltTab munching ~25% CPU.

panpietrek:
launchd and logd's CPU usage goes up as well, presumably due to this spamming launchd.log file:

<img width="1345" alt="obraz" src="https://user-images.githubusercontent.com/13390324/162566455-fd1a49fe-9c2b-4ceb-b568-1483df1567ba.png">

litori:
The fact that it uses so much cpu without anything open besides activity monitor is already a problem.  I've binned 6.33.0 for now as it is not usable.

metacodes:
@litori @panpietrek I have located the problem and am working on it. See #1437 . You can use 6.31.0 before this issue is fixed.

metacodes:
#1484

burdo3417:
Same problem. Also just firefox opened.
Macbook Pro M1 Pro
MacOS 12.3.1

otandka:
For easier searching by log message keywords, here are the messages from "log show -last 1m" that are repeated every 150usec or so.
```
2022-04-14 10:30:57.116774-0700 0x635391   Default     0x0                  1      0    launchd: [gui/501 [100006]:] failed lookup: name = com.apple.axserver, handle = 92824, flags = 0x3, requestor = AltTab[88006], error = 3: No such process
2022-04-14 10:30:57.116861-0700 0x635391   Default     0x0                  1      0    launchd: [gui/501 [100006]:] could not find active service for lookup: pid = 84636, name = com.apple.axserver
[...]

Issue #1324, opened 2022-01-13
Starting with macOS 12.2, AltTab only grabs windows in active spaces


skyzyx:
**Describe the bug**

* I work with 2 monitors
* I work almost exclusively in full-screen mode

I updated to macOS 12.2 Beta (21D5039d) this morning, and since then, AltTab only grabs the front-most full-screen app window from each of my displays — ignoring the full-screen windows that are not front-most.

I know that the full-screen support is buggy because this app has to use private APIs, etc., but I wanted to bring this situation to the attention of the developers as a known issue.

Usually when AltTab is missing windows, I can quit and re-launch the app to get it to pick up the current state of my environment. However, even this action no longer works as of this morning's macOS beta update.

**Screenshots / video**

http://s3.ryanparman.com.s3.amazonaws.com/alttab-bug-1.mp4

dnivi3:
Seeing the same on macOS 12.2 (21D49), the production version of macOS 12.2.

kaatt:
likely related to https://github.com/kasper/phoenix/issues/289 and https://github.com/ianyh/Amethyst/issues/1192

dnivi3:
Looks like Amethyst has fixed this in https://github.com/ianyh/Amethyst/pull/1184

lwouis:
@dnivi3 the code change you linked to is this:

![image](https://user-images.githubusercontent.com/106195/152257355-7f24877f-52f5-487a-90b7-b8a7ee716480.png)

Changing the order `remove -> add` to `add -> remove`. AltTab already does `add -> remove`:

![image](https://user-images.githubusercontent.com/106195/152257430-d32f87c2-80c2-461a-8f47-9caf40db6a93.png)

dnivi3:
@lwouis yeah, I realised my layman non-technical understanding of this isn't helpful. Is there any I can help or contribute towards this getting fixed?

lwouis:
@dnivi3 i'm afraid only code contributions would help. Please see https://github.com/lwouis/alt-tab-macos/issues/1179

lwouis:
Related: #1351

stevetodd:
Just to add the datapoint, the issue persists in the newly released macOS 12.3.

radaczynski:
[...]

PR #1484, opened 2022-04-10
fix high CPU usage #1481

metacodes:
See detail discussion in #1437 .


src/logic/Application.swift
```diff
@@ -158,20 +158,23 @@ class Application: NSObject {
...
-                try self.axUiElement!.subscribeToNotification(axObserver, notification, {
-                    DispatchQueue.main.async { [weak self] in
-                        guard let self = self else { return }
-                        // some apps have `isFinishedLaunching == true` but are actually not finished, and will return .cannotComplete
-                        // we consider them ready when the first subscription succeeds, and list their windows again at that point
-                        if !self.isReallyFinishedLaunching {
-                            self.isReallyFinishedLaunching = true
-                            self.observeNewWindows()
+        // we only need to subscribe to those apps which will have windows
+        if AXUIElement.isManageable(runningApplication) {
```
lwouis, line 162: This check is a bit late. Better not even create the `Application` instance, if we think it will not spawn any window. There is already such a method to filter apps here: https://github.com/lwouis/alt-tab-macos/blob/02590579df936d6a2124707e12d76865636fe23f/src/logic/Applications.swift#L104

You could add your new filtering logic there.

metacodes: Thank you for the suggestion. I found a compensation method to solve this problem. The initial technical attempts have been made, and I'll implement the code for this later. I'm a bit busy these two weeks.

src/api-wrappers/AXUIElement.swift
```diff
@@ -247,6 +247,80 @@ extension AXUIElement {
...
+        "com.apple.TextInputSwitcher"
+    ])
+
+    /**
+     *  Returns a Bool indicating whether or not the application will have windows.
+     *
+     *  @return true if the application will have windows and false otherwise.
+     */
+    static func isManageable(_ runningApp: NSRunningApplication) -> Bool {
+        guard let bundleIdentifier = runningApp.bundleIdentifier else {
```
lwouis, line 278: we have cases like the android emulator where the app has no `bundleIdentifier` but is still app with windows (see https://github.com/lwouis/alt-tab-macos/blob/02590579df936d6a2124707e12d76865636fe23f/src/logic/Applications.swift#L127)

src/api-wrappers/AXUIElement.swift
```diff
@@ -247,6 +247,80 @@ extension AXUIElement {
...
+        "com.apple.PowerChime",
+        "com.apple.WebKit.Networking",
+        "com.apple.WebKit.WebContent",
+        "com.apple.WebKit.GPU",
+        "com.apple.FollowUpUI",
+        "com.apple.controlcenter",
+        "com.apple.SoftwareUpdateNotificationManager",
+        "com.apple.TextInputMenuAgent",
+        "com.apple.TextInputSwitcher"
+    ])
```
lwouis, line 270: I don't think it's a good idea to hardcode specific `bundleIdentifier`s. If these start creating windows in further macOS releases, then AltTab will have to be updated to show them. Also how did you build that list? Maybe it's missing processes from previous macOS versions you don't use, or maybe (actually it's certain) different processes will run on your mac depending on your System Preferences, if you have a network printer, external monitor, keyboard, if your phone does AirPlay with the mac, etc.

I think we should find more flags or properties of apps that give strong hints that they will not spawn windows. Also retrying less often than 10ms is already a large workaround the CPU issue

src/api-wrappers/AXUIElement.swift
```diff
@@ -21,7 +21,7 @@ func retryAxCallUntilTimeout_(_ group: DispatchGroup?, _ timeoutInSeconds: Doubl
     } catch {
         let timePassedInSeconds = Double(DispatchTime.now().uptimeNanoseconds - startTime.uptimeNanoseconds) / 1_000_000_000
         if timePassedInSeconds < timeoutInSeconds {
-            BackgroundWork.axCallsQueue.asyncAfter(deadline: .now() + .milliseconds(10)) {
+            BackgroundWork.axCallsQueue.asyncAfter(deadline: .now() + .milliseconds(1000)) {
```
lwouis, line 24: I think a value of `250ms` may be a better comprise: 10ms is very aggressive for a responsive UX but at the cost of CPU. 1s is easy on the CPU, but a bit laggy. I think `250ms` is similar to human delay in processing changes on screen. So maybe a better compromise. See https://humanbenchmark.com/tests/reactiontime

metacodes:
@lwouis I use new method to aovid subcribing to applications withou windows. See details in the code comments.

src/logic/Applications.swift
```diff
@@ -104,7 +105,30 @@ class Applications {
     private static func isActualApplication(_ app: NSRunningApplication) -> Bool {
         // an app can start with .activationPolicy == .prohibited, then transition to != .prohibited later
         // an app can be both activationPolicy == .accessory and XPC (e.g. com.apple.dock.etci)
-        return (isNotXpc(app) || isAndroidEmulator(app)) && !app.processIdentifier.isZombie()
+        return (isNotXpc(app) || isAndroidEmulator(app)) && !app.processIdentifier.isZombie() && isAnWindowApplication(app)
```
lwouis, line 108: If your comment below is correct (that apps that trigger a `didActivateApplicationNotification` event must have windows), then all these checks are no longer useful for these apps, no?

I imagine that they would only be useful for the `wasLaunchedBeforeAltTab == true` apps. So maybe we should add that boolean check first (left-side), to avoid the more expansive checks in the case where it's `wasLaunchedBeforeAltTab == false`, no?

src/logic/events/WorkspaceEvents.swift
```diff
@@ -14,11 +14,30 @@ class WorkspaceEvents {
...
+    }
+
+    // We add apps when we receive a didActivateApplicationNotification notification, not when we receive an apps launched, because any app will have an apps launched notification.
+    // But we are only interested in apps that have windows. We think that since an app can be activated, it must have a window, and subscribing to its window event makes sense and is likely to work, even if it requires multiple retries to subscribe.
+    // I'm not very sure if there is an edge case, but so far testing down the line has not revealed it.
+    // When we receive the didActivateApplicationNotification notification, NSRunningApplication.isActive=true, even if the app is not the frontmost window anymore.
+    // If we go to add the application when we receive the message of apps launched, at this time NSRunningApplication.isActive may be false, and try axUiElement.windows() may also throw an exception.
+    // For those background applications, we don't receive notifications of didActivateApplicationNotification until they have their own window. For example, those menu bar applications.
```
lwouis, line 33: If this is true, then your approach here to listen to `.didActivateApplicationNotification` is genius! Did you think or it yourself or was it used by other projects?

Also, if I summarize this change, it seems you found an API that gives us properly filtered apps (only apps with windows), but that's only for apps launched *after* AltTab is started. For apps launched before AltTab is started, we still have to filter out processes with `activationPolicy`, `XPC`, etc. But luckily, apps launched before AltTab should not throw an exception on `.windows()`.

And then you mention the edge-case of an app launching at the same time as AltTab, which would be annoying. This is actually a very important edge-case, unfortunately. That's what happens when you log into macOS. All the startup apps launch at once, concurrently. This means AltTab is launched, and all sorts of app at the very same time.

I guess for that scenario, `didActivateApplicationNotification` will not help us, and we will need to rely on the other existing checks (XPC, etc), and with the new `250ms` delay, it may be an acceptable situation?

Am I understanding correctly the whole approach?

metacodes: > Did you think or it yourself or was it used by other projects?

At first I was thinking that when we open an application, as long as they show a window, then a change in window switching is bound to happen, and if we can find a way to observe this change, we will be able to tell that the program is equipped with a window. Then I used some window management software, such as automatic window layout APP, like AppGrid, Rectangle, and found that they can observe this change very well, so I went to study their code, and then found in Rectangle's code that it subscribes to `.didActivateApplicationNotification` notifications. I was curious, so I wrote some test code to see the effect. Turns out it worked and was able to solve half of the problem, the other half still had to be figured out.

> And then you mention the edge-case of an app launching at the same time as AltTab, which would be annoying. This is actually a very important edge-case, unfortunately. That's what happens when you log into macOS. All the startup apps launch at once, concurrently. This means AltTab is launched, and all sorts of app at the very same time.

That's my concern, and I still need to do some testing on this situation.

> I guess for that scenario, didActivateApplicationNotification will not help us, and we will need to rely on the other existing checks (XPC, etc), and with the new 250ms delay, it may be an acceptable situation?

This edge case only happens when `wasLaunchedBeforeAltTab=true`, at which point we go to the `axUiElement.windows()` test. If it happens that the program takes a long time to load, then `axUiElement.windows()` will most likely throw an exception, and at that point we do need to detect it by other means (e.g. XPC, etc), otherwise we will lose the window. Of course, reopening the program that lost the window, AltTab will capture its window without any problem.

metacodes: For that edge-case, we probably don't need to restart the applications that are missing windows, we can do a check when we get the `.didActivateApplicationNotification` notification, and if the application is not in the AltTab list of existing windows, then we add it.

metacodes:
@lwouis I have submitted a new commit for fixing #1324 and some minor improvements to full-screen thumbnails.

src/logic/Applications.swift
```diff
@@ -115,7 +124,13 @@ class Applications {
...
             // I'm not quite sure if this happens, but even if it does, then after restarting this application, AltTab captures its window without any problem. I think this happens rarely.
             let allWindows = CGWindow.windows(.optionAll)
-            guard let winApp = (allWindows.first { app.processIdentifier == $0.ownerPID() && $0.isNotMenubarOrOthers() && $0.id() != nil && $0.title() != nil}) else {
+            guard let winApp = (allWindows.first { app.processIdentifier == $0.ownerPID()
+                    && $0.isNotMenubarOrOthers()
+                    && $0.id() != nil
+                    && $0.bounds() != nil
+                    && CGRect(dictionaryRepresentation: $0.bounds()!)!.width > 0
+                    && CGRect(dictionaryRepresentation: $0.bounds()!)!.height > 0
+            }) else {
```
lwouis, line 133: As I said in the other ticket, you're re-implementing a mirror to https://github.com/lwouis/alt-tab-macos/blob/b668928c89dd256b39084d47b523ea882ebf6833/src/api-wrappers/AXUIElement.swift#L76 here. This will not be good enough. AltTab will thus show windows it shouldn't show.

src/api-wrappers/PrivateApis.swift
```diff
@@ -122,10 +122,15 @@ func CGSCopyWindowsWithOptionsAndTags(_ cid: CGSConnectionID, _ owner: UInt32, _
...
 
 // adds the provided windows to the provided spaces
-// * macOS 10.10+
+// * macOS 10.10-12.2
 @_silgen_name("CGSAddWindowsToSpaces")
 func CGSAddWindowsToSpaces(_ cid: CGSConnectionID, _ windows: NSArray, _ spaces: NSArray) -> Void
 
+// Move the given windows (CGWindowIDs) to the given space (CGSSpaceID)
+// * macOS 10.10+
+@_silgen_name("CGSMoveWindowsToManagedSpace")
```
lwouis, line 131: Does this work exactly like `CGSAddWindowsToSpaces`? I remember testing almost every API from https://github.com/phracker/MacOSX-SDKs/blob/10dd4868459aed5c4e6a0f8c9db51e20a5677a6b/MacOSX10.10.sdk/System/Library/Frameworks/CoreGraphics.framework/Versions/A/CoreGraphics.tbd which had the word "space" or "window" in it. I'm surprised that I missed this one. Are you sure it works in every scenario (e.g. multiple monitors, minimized window, hidden app, etc)?

src/api-wrappers/PrivateApis.swift
```diff
@@ -122,10 +122,15 @@ func CGSCopyWindowsWithOptionsAndTags(_ cid: CGSConnectionID, _ owner: UInt32, _
 func CGSManagedDisplayGetCurrentSpace(_ cid: CGSConnectionID, _ displayUuid: CFString) -> CGSSpaceID
 
 // adds the provided windows to the provided spaces
-// * macOS 10.10+
+// * macOS 10.10-12.2
```
lwouis, line 125: RIP 😄

src/api-wrappers/CGWindowID.swift
```diff
@@ -19,11 +19,63 @@ extension CGWindowID {
         return CGSCopySpacesForWindows(cgsMainConnectionId, CGSSpaceMask.all.rawValue, [self] as CFArray) as! [CGSSpaceID]
     }
 
-    func screenshot() -> CGImage? {
-        // CGSHWCaptureWindowList
+    // fullscreen has multiple windows
```
lwouis, line 22: I don't understand this. I just fullscreen'ed Notes.app, and AltTab shows the same thing I see.

Could you share screenshots to illustrate what you mean by "toolbar window and a main window", and generally what problem this new method fixes?

src/logic/Window.swift
```diff
@@ -83,13 +121,13 @@ class Window {
...
     func refreshThumbnail() {
-        guard let cgImage = cgWindowId.screenshot() else { return }
+        guard let cgImage = cgWindowId.screenshot(self) else { return }
         thumbnail = NSImage(cgImage: cgImage, size: NSSize(width: cgImage.width, height: cgImage.height))
         thumbnailFullSize = thumbnail!.size
     }
 
     func close() {
-        if isWindowlessApp { return }
+        if isWindowlessApp || self.axUiElement == nil { return }
```
lwouis, line 130: Does this mean we can't close a window if we haven't visited its Space? That not ideal...

src/logic/Window.swift
```diff
@@ -102,7 +140,7 @@ class Window {
     }
 
     func minDemin() {
-        if isWindowlessApp { return }
+        if isWindowlessApp || self.axUiElement == nil { return }
```
lwouis, line 143: See comment above

src/logic/Window.swift
```diff
@@ -119,7 +157,7 @@ class Window {
     }
 
     func toggleFullscreen() {
-        if isWindowlessApp { return }
+        if isWindowlessApp || self.axUiElement == nil { return }
```
lwouis, line 160: See comment above

src/logic/Window.swift
```diff
@@ -145,13 +183,43 @@ class Window {
...
+                    BackgroundWork.accessibilityCommandsQueue.asyncWithCap { [weak self] in
+                        guard let self = self else { return }
+                        var psn = ProcessSerialNumber()
+                        GetProcessForPID(self.application.pid, &psn)
+                        _SLPSSetFrontProcessWithOptions(&psn, self.cgWindowId, .userGenerated)
+                        self.makeKeyWindow(psn)
+                        self.axUiElement.focusWindow()
+                    }
+                }
+            } catch {
```
lwouis, line 220: I don't think exceptions can bubble up to here. `retryAxCallUntilTimeout` happens on another thread, and absorbs exceptions within

src/api-wrappers/CGWindowID.swift
metacodes: Before adding the `fullScreenshot` method
image

After adding the `fullScreenshot` method
image

Maybe it depends on machine model. I'm using MBP 16' with Apple Silicon.

src/logic/Applications.swift
```diff
@@ -32,10 +40,11 @@ class Applications {
             let windowsOnOtherSpaces = Spaces.windowsInSpaces(otherSpaces)
             let windowsOnlyOnOtherSpaces = Array(Set(windowsOnOtherSpaces).subtracting(windowsOnCurrentSpace))
             if windowsOnlyOnOtherSpaces.count > 0 {
-                // on initial launch, we use private APIs to bring windows from other spaces into the current space, observe them, then remove them from the current space
-                CGSAddWindowsToSpaces(cgsMainConnectionId, windowsOnlyOnOtherSpaces as NSArray, [Spaces.currentSpaceId])
-                Applications.observeNewWindowsBlocking()
-                CGSRemoveWindowsFromSpaces(cgsMainConnectionId, windowsOnlyOnOtherSpaces as NSArray, [Spaces.currentSpaceId])
+                // Currently we add those window in other space without AXUIElement init
+                // We don't need to get the AXUIElement until we focus these windows.
```
lwouis, line 44: If you're correct that `CGSMoveWindowsToManagedSpace` is a perfect replacement to `CGSAddWindowsToSpaces`, then we could at least fix the 12.3 issue with `CGSAddWindowsToSpaces` not working, but replacing the call to `CGSMoveWindowsToManagedSpace`. That would fix the issue. Then, separately, we can have the conversation about the invisible window trick, and how to move forward there

src/api-wrappers/CGWindowID.swift
lwouis: Oh, it does it on my machine too (still on Catalina). Very cool then! I wonder how many apps really benefit from that nuance though. It's a cool trick, but in the case of Notes.app, it's not very necessary. Maybe some other apps have an important UI element that would get lost though. I worry a bit about the complexity/performance/potential-failure-points of created the composition vs showing the simple image, though. I think it would probably be worth it to first find an app that truly benefits from this refinement before putting it in. What do you think? Especially given that in order to do this trick, we have to rely on CGS API, which has all the problems I mentioned in other comments

src/logic/Applications.swift
metacodes: I add these codes just to filter those apps which have no windows(e.g. com.apple.universalcontroller). This is for those apps launched before AltTab. At that time, we haven't go to `AXUIElement.isActualWindow`. The code is just there to allow us to avoid observing to applications we are not interested in and to avoid wasting CPU time on `subscribeToNotification` retries.

src/api-wrappers/PrivateApis.swift
metacodes: This api will move the window from the original space to another space. It means the window will disappear from the original space and show on another space. I did not test the scenarios you mentioned. I just use this api to move the invisible window to the space we want to. I will do some tests. By the way, I can not do the tests for multiple monitors, because I only have one monitor.😂

src/api-wrappers/CGWindowID.swift
metacodes: Indeed, I also took into account the performance issues. In fact, this small issue has little impact, I just made it a little bit optimized as free time. Reminders.app, Hopper Disassembler.app, Xcode.app, Microsoft Word.app, Calendar.app also have this problem as I known.

metacodes: Initially I tried to use CGWindowListCreateImageFromArray, but it didn't work at all. Maybe I'm not using it the right way.

src/logic/Window.swift
metacodes: It's a pity that HyperSwitch also can't do that either.

lwouis: I just tested and it can.

I did this to test:

* Opened a new window
* Send it to another Space
* Launch HyperSwitch
* Press alt-tab
* Notice that HS shows the window
* Hover the mouse over the window in HS
* A close button appears, press it
* The window on the other Space is closer, even though I never navigated to that Space

How can HS close the window without the AX reference?

lwouis:
FYI, I just pushed a release with these 2 commits: https://github.com/lwouis/alt-tab-macos/commit/0569ed0f8603ca0611c9d6bc0f00046f48c4415e https://github.com/lwouis/alt-tab-macos/commit/cdb461ab6820b21146ee16fa85d197b846e1b276

I thought they may mitigate the high CPU issues people are facing at the moment. I think this buys more time for this pull-request to become more robust and maybe deliver breakthroughs

src/logic/Window.swift
metacodes: On macOS 12.3.1, I did the steps same to you, but HyperSwitch just took me to that space and didn't close the window. Maybe the private APIs they used also broken.

src/logic/Applications.swift
lwouis: Please see my follow-up here: https://github.com/lwouis/alt-tab-macos/issues/1324#issuecomment-1120156153

lwouis:
@metacodes I'll close this PR now since I merged part of it. If you're still interested in improving perf, could you please open a new PR with your new improvements?

metacodes:
> @metacodes I'll close this PR now since I merged part of it. If you're still interested in improving perf, could you please open a new PR with your new improvements?

OK👌

PR #1566, opened 2022-05-11
Improve the Chinese translation of "Miscellaneous"

ifsheldon:
The previous translation of "Miscellaneous" to Chinese takes the meaning when the word is used as an adj. The new translation takes the meaning when the word is used as a none. "杂项", when back-translated to English, is "Miscellaneous items", which can be derived as "Miscellaneous settings" based on the context. It is a more suitable translation in my opinion.

BTW: Loving this app! THANKS a lot to every contributor.

### DIFF
--- a/resources/l10n/zh-CN.lproj/Localizable.strings
+++ b/resources/l10n/zh-CN.lproj/Localizable.strings
@@ -192,7 +192,7 @@
 "Minimized windows:" = "最小化窗口：";
 
 /*No comment provided by engineer.*/
-"Miscellaneous:" = "混杂的：";
+"Miscellaneous:" = "杂项：";
 
 /*No comment provided by engineer.*/
 "Mouse hover" = "鼠标悬停";


lwouis:
@ifsheldon thank you a lot for contributing!

To change localization, please [go here](https://poeditor.com/join/project/8AOEZ0eAZE), and you will have a nice UI to update the text.

poeditor.com is our source-of-truth for localized strings, so I can't merge this PR. The change has to be done there in the editor, and then there are scripts which will bring the new values into the git project automatically.

An added benefit of updating through poeditor.com is that your name will be mentioned in the list of contributors, both on the website and in the app itself 👍

ifsheldon:
OK, I have changed the simplified Chinese translation of "Miscellaneous" and added traditional Chinese translation of it. I'm not familiar with poeditor.com, but it seems I have submitted the changes. Can you please check if everything is alright? Thanks!

lwouis:
It's correctly updated. It will ship in the app next release 👍 

Thank you~

PR #1660, opened 2022-06-05
docs: update homebrew instructions for newer versions

s4na:
Since `brew cask install` has been disabled, you must use `brew install --cask`.

reference: https://github.com/ansible-collections/community.general/issues/1524

---

Best regards,
s4na

### DIFF
--- a/docs/Introduction.md
+++ b/docs/Introduction.md
@@ -27,7 +27,8 @@ permalink: /
 
 Alternatively, you can use [homebrew](https://brew.sh/):
 
-* Homebrew 2.5 or above: `brew install alt-tab`
+* Homebrew 2.7 or above: `brew install --cask alt-tab`
+* Homebrew 2.6 or below and 2.5 or above: `brew install alt-tab`
 * Homebrew 2.4 or below: `brew cask install alt-tab`
 
 ## Compatibility


docs/Introduction.md
```diff
@@ -27,7 +27,8 @@ permalink: /
 
 Alternatively, you can use [homebrew](https://brew.sh/):
 
-* Homebrew 2.5 or above: `brew install alt-tab`
+* Homebrew 2.7 or above: `brew install --cask alt-tab`
```
lwouis, line 30: This is pretty ridiculous 😅 . They have removed a previously working command *again*? Is there a better way than listing more and more idioms here? It looks crazy to me that we would need to display a table with variations because homebrew keeps changing their mind on syntax, and stops supporting wildly used command syntaxes. There must be a better way, no?

docs/Introduction.md
```diff
@@ -79,4 +80,4 @@ There are also related apps which don’t really overlap in functionality, but t
 
 ## License
 
-AltTab is under the [GPL-3.0 license](https://github.com/lwouis/alt-tab-macos/blob/master/LICENCE.md). 
+AltTab is under the [GPL-3.0 license](https://github.com/lwouis/alt-tab-macos/blob/master/LICENCE.md).
```
lwouis, line 83: Could you please remove that change from the commit? Thank you~

lwouis:
Thank you for this PR @s4na! I have left some comments. I need to dig deeper I guess because this situation of having to add so many flavours of homebrew looks crazy to me.

docs/Introduction.md
lwouis: Ok I dug deeper, and [this discussion sums up the situation](https://github.com/orgs/Homebrew/discussions/340).

The homebrew maintainers simply don't care about breaking their own commands for people. They did it here 20 days after deprecating the command, which is shocking for a popular system to break all its users with 20 days of warning.

They have a pretty bad culture I think. [They think it's all about them](https://docs.brew.sh/Maintainers-Avoiding-Burnout#3-prioritise-maintainers-over-users) because without them, there is no homebrew. So users come second. I disagree that it's such a binary choice, and I think they should foster a culture of stable interfaces, like the Linux kernel. Stability more valuable, the more popular software is. So at homebrew scale, they should treat it probably **more** seriously than new features or even bug fixes, because people may build workarounds around bugs, and rely on those. It's like a big boat that has to move slowly now.

Anyway, let's add yet another line in this readme. If they do another breaking move like this, I'll simply stop playing cat-and-mouse, and link to [the cask page](https://formulae.brew.sh/cask/alt-tab), and let homebrew users complain to homebrew when the command doesn't work on their older version of homebrew. Why don't they even mention the various instructions possible on their own page I just linked? It seems so careless to me.

For now, we are doing their job of helping their users have a good experience, regardless of the version of homebrew they use. But I don't want this section of the readme to become it's own little book.

lwouis: On a technical side, I think we should move to a table now that there are 3 options. I'll push a commit with it, so no need to do it yourself. I'll commit with your name/email so you'll get credited for it, since this whole PR was your work. Thanks again!

docs/Introduction.md
s4na: Thank you.
Removed -> https://github.com/lwouis/alt-tab-macos/pull/1660/commits/253a73cb896aff2d36de338a23fb5f585b67eb1b

lwouis:
I have bundled this change with the next release. Closing this PR, but don't worry your change will be in the next release, with your name on the commit, and you'll be credited for the work in the app itself, and on the website as well.

PR #1666, opened 2022-06-07
feat: Big Sur app icon

foxster-mp4:


lwouis:
Hi @foxster-mp4,

Could you please elaborate on why you are sharing this PR? Something is wrong with the app's icon on Big Sur? I'm not aware of that, so I would like to understand the issue you found, and how you addressed it. As it stands, there is only a title mentioning the icon, and then it was updated. Why? How?

Thank you,

Louis

foxster-mp4:
Hey there! Sorry I'm quite new to all of this pull request stuffs and I should have been more elaborate. From macOS 11.0 (Big Sur), Apple switched to using macOS app icons that are more iOS-like, which are squared with rounded corners. I noticed that AltTab is still using a slanted looking app icon (similar to ones from OS X), and although it's looking absolutely great and fine, it doesn't quite fit with Apple's current design guidelines, which as I mentioned uses app icons that are squared with rounded corners (see attached image). To address this, I made a blank, white, rounded corner square and dropped the current AltTab icon inside, and that's the only file I changed. I'm not a professional graphics designer so that's as far as I took it, just to demonstrate the idea. I know it looks really simple so you can find better alternatives to the AltTab icon (that matches the design guidelines) [here](https://macosicons.com/#/alttab). Let me know what you think!
poc

lwouis:
For artistic references:

Here's is what app icons look like on macOS 12.4:

image

Out of all the [Apple docs about app icons](https://developer.apple.com/design/human-interface-guidelines/foundations/app-icons), I found this to be actually helpful to understand their icon template and ideology:

image

Other community-created icons on [macosicons.com](https://macosicons.com/#/):

image

lwouis:
@fturcheti would you be interested to follow-up on your original master-piece? I was looking at your designs from 2 years ago, and they still look really cool:

Experiments | Refinements
-|-
image | image

I'm thinking that rather than wrapping the current icon inside a rounded-corner square, we could embrace the new guidelines and do something more interesting.

fturcheti:
Hey @lwouis! I finally got some free time to work on a new version of the icon. I tried to follow the same idea but adapting it this new icon template. 

Icon-1024

Screen Shot 2022-08-23 at 15 18 30

lwouis:
@fturcheti this looks great! Thank you for working on this! Do you think you could do a variation of this which uses the color scheme of the current icon? I'm thinking that it would make it more recognizable for existing users that it's the same app in the new macOS environment.

fturcheti:
@lwouis I tried to follow the color scheme from the first icon, but I liked these "hotter" hues. Anyway, here's a variation with colors more similar to the "old one":

Icon v2-1024

Screen Shot 2022-08-23 at 18 02 50

Screen Shot 2022-08-23 at 18 08 40

lwouis:
@fturcheti these are really nice, thank you!

PR #1685, opened 2022-06-14
Fixed and improved German translation

LostInCompilation:
Improved German translation.


LostInCompilation:
Oh well, just found the comment about POEditor. I will edit the translation there.

PR #1686, opened 2022-06-14
Added contribution info to README

LostInCompilation:
Maybe a good idea to add this small hint, since a Pull Request for translation happened at least twice now (including my error, lol)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,7 @@
 **AltTab** brings the power of Windows alt-tab to macOS
 
 [Find out more on the official website](https://alt-tab-macos.netlify.app/)
+
+#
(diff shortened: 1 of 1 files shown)


lwouis:
I would like to keep all project info on the website. The readme just links there. Otherwise, by that logic, i should bring other info here on github like compatibility, screenshots, etc, to avoid people missing it on the website. But then the readme and the website would end up being the same content just at 2 different URLs.

Sorry you had to redo the work in poeditor.com. I hope you understand the direction i'm taking

LostInCompilation:
Sure no problem. It was just a suggestion :)

Issue #1241, opened 2021-11-20
Migrate from TravisCI to Github Actions


lwouis:
**Is your feature suggestion related to a problem? Please describe.**
Currently CI/CD is done on TravisCI. They sponsor OSS with 500 minutes/month of build on a mac. I just checked out Github Actions pricing and their free tier have 2000 minutes/month.

Also TravisCI has me open a ticket to ask for credits every time I run out. It's not fun, and it takes them weeks sometimes to re-credit the project, which freezes any release we would have ready to ship.

For these reasons, it may be worth the effort to move to Github Actions.

We could also look at other providers like Azure Pipelines, which I've used successfully in the past as well.

**Describe the solution you'd like**
Not having to worry about build minutes.

**Describe alternatives you've considered**
We can stay on TravisCI. It works well and was hard/long to setup.

lwouis:
Github Actions actually has a [10x minutes multiplier for macOS machines](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers). This means that we would only get 200min/month with them.

lwouis:
Migrated today 🎉

PR #1739, opened 2022-06-26
add GitHub Actions workflow for build

ryenus:
setup a GitHub Actions workflow to run the build

- fix compile error: value of optional type must be unwrapped
- add GitHub Actions workflow for build

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,48 @@
+name: Build
+
+on: [push, pull_request]
+
+permissions: read-all
+
+env:
+  XCODE_WORKSPACE: "alt-tab-macos.xcworkspace"
+  XCODE_SCHEME: "Debug"
+  CODE_SIGN_IDENTITY: "-"
+  BUILD_DIR: "build"
+  XCODE_ARCHIVE: "AltTab.xcarchive"
+  APP_NAME: "AltTab.app"
+  EXPORT_OPTIONS_PLIST: "exportOptions.plist"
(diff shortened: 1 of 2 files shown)


lwouis:
Hi @ryenus,

Could you share a bit more context as to what your goal is with this PR? Are you trying to implement https://github.com/lwouis/alt-tab-macos/issues/1241?

The [CI/CD flow we currently have on TravisCI](https://github.com/lwouis/alt-tab-macos/blob/master/.travis.yml) is pretty complex/rich. It does many different things, and is the only example of CD for a macOS app I've seen in the wild actually. It builds the app, signs it, notarizes it, uploads the build to Github, uploads build symbols to AppCenter Crashlytics, auto-generate a semver, auto-generate release notes based on commit messages, updates the appcast file for Sparkle updates, updates the Netlify website, makes sure translators on poeditor.com and contributors on github are credited both in the app and on the website, etc.

I looked at the github workflow in this PR, and it seems to do something quite different. Building, making a dmg, almost none of the steps I listed above, etc. So I'm not sure what your intention is with this PR.

Furthermore, regarding Travis vs Github, these days, the Travis staff replies quickly to my requests when I ask for a credit refill. So since we have a working setup there, there is no incentive to do the complex migration to Github.

Anyway please let me understand better your intentions here. Maybe it will become clearer then 👍

ryenus:
@lwouis, I was mainly looking around for a CI build which runs on latest macOS and got this one using GitHub Actions. If this happens to address #1241 then I'd be happy it can be taken for that :-)

Issue #1852, opened 2022-08-09
AltTab hangs when launched from the Applications folder on Ventura beta 5


zacharee:
**Describe the bug**

I updated to Ventura beta 5 today on my M1 Pro MBP and noticed that AltTab was hanging on launch.

After playing around with building my own version, I noticed that when run from the Applications folder it hangs. Otherwise (with the debug version that doesn't move itself to `/Applications`) it runs fine.

**Steps to reproduce the bug**

1. Update to Ventura beta 5 on an M1 Mac.
2. Try opening AltTab from the Applications folder and notice it hangs.
3. Try opening a debug version of AltTab from the Applications folder and notice it hangs.
4. Try opening a debug version of AltTab from elsewhere and notice it has no issues.

Note: I built my debug version with Xcode 14.0 beta 5.

lwouis:
Could you better describe what you mean by "hanging"? Does it spins with high CPU?

If you can build it locally, could you then launch it in debug mode locally? You could then pause the debug and see where the execution is spending cycles or where it's blocking.

Alternatively, a spin dump from Activity Monitor would help diagnose the issue.

zacharee:
> Could you better describe what you mean by "hanging"? Does it spins with high CPU?
> 
> If you can build it locally, could you then launch it in debug mode locally? You could then pause the debug and see where the execution is spending cycles or where it's blocking.
> 
> Alternatively, a spin dump from Activity Monitor would help diagnose the issue.

As soon as it launches, it's marked as "not responding" in Activity Monitor. The menu bar icon consistently appears and the settings window occasionally appears but will have a beach ball of death. CPU usage is low.

The problem is that it only hangs if the `.app` is in `/Applications`. If I do a direct launch from AppCode there are no issues.

I can't seem to launch the debug executable inside `/Applications` through the terminal to get a log, either. It doesn't see it has screen recording permissions.

Here's a spindump:

[Spindump.txt](https://github.com/lwouis/alt-tab-macos/files/9294180/Spindump.txt)

zacharee:
Even weirder: now the debug app only successfully launches if it's run through AppCode. Double-clicking in Finder causes it to hang no matter where the file is located.

lwouis:
> I can't seem to launch the debug executable inside /Applications through the terminal to get a log, either. It doesn't see it has screen recording permissions.

I think it will work if you give the permissions to Terminal.app. Could you try that?

> Here's a spindump

[...]

PR #1865, opened 2022-08-12
fix: workaround for not responding on Ventura

zacharee:
Thanks to the discussion in https://github.com/lwouis/alt-tab-macos/issues/1852 and @Write's finding of self-observing being the issue, I narrowed it down further to some issue with `retryAxCallUntilTimeout()`. Something isn't happy about being put into a background queue and then back into the main queue when it's the AltTab application.

The "fix" here is to check if the current application is AltTab and subscribe to events directly instead of through `retryAxCallUntilTimeout()`.

A better long-term fix is probably to look into all the async stuff and see if it can be simplified, but this at least gets the app functional on Ventura again.

### DIFF
--- a/src/logic/Preferences.swift
+++ b/src/logic/Preferences.swift
@@ -242,12 +242,12 @@ class Preferences {
     @available(OSX, deprecated: 10.11)
     private static func migrateLoginItem() {
         do {
-            let loginItems = LSSharedFileListCreate(nil, kLSSharedFileListSessionLoginItems.takeRetainedValue(), nil).takeRetainedValue()
-            let loginItemsSnapshot = LSSharedFileListCopySnapshot(loginItems, nil).takeRetainedValue() as! [LSSharedFileListItem]
+            let loginItems = LSSharedFileListCreate(nil, kLSSharedFileListSessionLoginItems.takeRetainedValue(), nil)!.takeRetainedValue()
+            let loginItemsSnapshot = LSSharedFileListCopySnapshot(loginItems, nil)!.takeRetainedValue() as! [LSSharedFileListItem]
             let itemName = Bundle.main.bundleURL.lastPathComponent as CFString
             let itemUrl = URL(fileURLWithPath: Bundle.main.bundlePath) as CFURL
             loginItemsSnapshot.forEach {
-                if (LSSharedFileListItemCopyDisplayName($0)?.takeRetainedValue() == itemName) ||
+                if (LSSharedFileListItemCopyDisplayName($0).takeRetainedValue() == itemName) ||
                        (LSSharedFileListItemCopyResolvedURL($0, 0, nil)?.takeRetainedValue() == itemUrl) {
                     LSSharedFileListItemRemove(loginItems, $0)
                 }
--- a/src/logic/Application.swift
+++ b/src/logic/Application.swift
@@ -15,10 +15,12 @@ class Application: NSObject {
     var pid: pid_t!
     var focusedWindow: Window? = nil
     var alreadyRequestedToQuit = false
+    var isCurrent: Bool!
 
-    init(_ runningApplication: NSRunningApplication) {
+    init(_ runningApplication: NSRunningApplication, isCurrent: Bool) {
         self.runningApplication = runningApplication
         pid = runningApplication.processIdentifier
+        self.isCurrent = isCurrent
         super.init()
         isHidden = runningApplication.isHidden
         hasBeenActiveOnce = runningApplication.isActive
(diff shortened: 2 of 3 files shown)


Write:
Wow, awesome that someone could do that ! Thank a lot !

src/logic/Preferences.swift
```diff
@@ -242,12 +242,12 @@ class Preferences {
...
         do {
-            let loginItems = LSSharedFileListCreate(nil, kLSSharedFileListSessionLoginItems.takeRetainedValue(), nil).takeRetainedValue()
-            let loginItemsSnapshot = LSSharedFileListCopySnapshot(loginItems, nil).takeRetainedValue() as! [LSSharedFileListItem]
+            let loginItems = LSSharedFileListCreate(nil, kLSSharedFileListSessionLoginItems.takeRetainedValue(), nil)!.takeRetainedValue()
+            let loginItemsSnapshot = LSSharedFileListCopySnapshot(loginItems, nil)!.takeRetainedValue() as! [LSSharedFileListItem]
             let itemName = Bundle.main.bundleURL.lastPathComponent as CFString
             let itemUrl = URL(fileURLWithPath: Bundle.main.bundlePath) as CFURL
             loginItemsSnapshot.forEach {
-                if (LSSharedFileListItemCopyDisplayName($0)?.takeRetainedValue() == itemName) ||
+                if (LSSharedFileListItemCopyDisplayName($0).takeRetainedValue() == itemName) ||
```
zacharee, line 250: I had to do some unwrapping here to get AltTab to compile. I'm not sure if my lint rules are different or if Xcode 14 has new nullability indicators.

lwouis: I think this is neither. I think it's the following: the SDK for more recent versions of macOS has updated these public APIs signature. And Apple has this terrible binding of certain macOS versions > certain XCode versions > certain SDK versions. So you're on Ventura, so you have to use a more recent XCode which has to use a more recent SDK, so you get the compiling errors.

Meanwhile i'm on Catalina still, and so is the CI machine. So both build fine.

lwouis:
As i mentioned in the issue, I would like to wait out the beta so that we see what the situation is ounce things have settled.

This is interesting and a nice thing to have, but for a local build enthusiasts of the beta can share amongst enthusiasts. I can't really release a new public release with this workaround.

Let's keep this PR open for people to build it if they are interested, and to pick up from here later when the beta is over.

zacharee:
Just to make it easier on people, here's a debug compile of the app. 

You'll need to run `sudo xattr -cr /path/to/AltTab.app` in a terminal before you can open it.

[AltTab.app.zip](https://github.com/lwouis/alt-tab-macos/files/9334145/AltTab.app.zip)

simyanliang:
> Just to make it easier on people, here's a debug compile of the app.
> 
> You'll need to run `sudo xattr -cr /path/to/AltTab.app` in a terminal before you can open it.
> 
> [AltTab.app.zip](https://github.com/lwouis/alt-tab-macos/files/9334145/AltTab.app.zip)

Ran this in Ventura beta 5, from the applications folder and it seems to be working. The only thing I needed to do was to remove the instances of AltTab from the accessibility and screen recording pages in settings and open AltTab again. Otherwise it didn't detect that it had access.

lwouis:
@zacharee here's a [build of this PR, signed with my certificate](https://github.com/lwouis/alt-tab-macos/files/9334253/AltTab.app.zip). This means it's the same signature as the public releases. I think it should avoid people to:

* Have to use `xattr`
* Have to re-set the permissions for the app

bradhs:
> 

This worked for me, though I had to allow it to run.

PKizzle:
I have noticed that the settings window opens every time AltTab is started. Is there a way to prevent this from happening?

lwouis:
@PKizzle are you using my build above? It seems you get the behavior of a Debut build. My build above should be Release and not do that.

PKizzle:
Oh yes, I was using the Debut build. I just tried out your signed version and it seems to work fine. Thank you very much!

100472175:
I'm running beta 6 and its working for me, using the zip provided by:#

> @zacharee here's a [build of this PR, signed with my certificate](https://github.com/lwouis/alt-tab-macos/files/9334253/AltTab.app.zip). This means it's the same signature as the public releases. I think it should avoid people to:
> 
> * Have to use `xattr`
> * Have to re-set the permissions for the app

Thanks a lot :)

lwouis:
Closing this PR since the [issue in the beta](https://github.com/lwouis/alt-tab-macos/issues/1852) was reported fixed by Apple 👍

Issue #1466, opened 2022-04-05
File copy progress windows are not picked up


adarshvadwitiya:
**Describe the bug**

AltTab doesn't pick up file copy progress windows

**Screenshots**
This is what my preferences look like:
![SCR-20220405-na7](https://user-images.githubusercontent.com/30106062/161742473-0480abd5-74fc-4a02-9db1-b80b13e7e61c.png)

![SCR-20220405-naf](https://user-images.githubusercontent.com/30106062/161742474-f94f2242-d41e-464b-9e20-a84063b4a3e3.png)

**Steps to reproduce the bug**

1. Have a file copy progress window open
2. Click CMD+Tab
3. All other windows show up except the file copy progress window

lwouis:
Are these really windows though? AltTab has to filter out things like dialogs and HUDs. It may be that Finder declares those as such and not as windows, in which case AltTab would rightfully filter them out.

Someone should check how they are declared at the technical level

joostin81:
I believe HyperSwitch shows file copy progress windows but not dialogs/HUDs

lwouis:
Anyone knows a trick to spawn such window and have it stay for a little while? My SSD is really fast, and most file copies don't spawn a window, or very briefly.

adarshvadwitiya:
Not sure if they’ll help but you can create large dummy files using `mkfile <size>G your-file-name`. So if you move that to an external drive, the window should stick around for a while

XInTheDark:
> My SSD is really fast, and most file copies don't spawn a window, or very briefly.

What SSD do you use? I wanna get one 😂

lwouis:
The issue I think is that on macOS file-system, moving or even copying a large file is just creating/moving a file descriptor. So it's instant. I was able to get a Finder copy window by copying a folder with lots of subfolders (this repo actually).

So the issue is that the Finder copy window doesn't pass the height > 100px check. On my system, in my language (English), its height is 69px.

I'm not sure how to deal with this situation. I explored the heuristic in https://github.com/lwouis/alt-tab-macos/issues/456 in the past. Here is an example of an app for which we need to filter out windows smaller than 100px: https://github.com/lwouis/alt-tab-macos/issues/562.

What I could do is add a whitelist for Finder: if it's Finder, we show windows regardless of their size. However it doesn't seem ideal as other apps could have legitimate small windows. Yet other apps have small non-windows.

What to do?

XInTheDark:
[...]

PR #1871, opened 2022-08-14
Reorganize how window validity exceptions are checked and add ColorSlurp exception

zacharee:
This PR has 2 changes:

- I noticed that checking for exceptions for when to show windows is getting a little complex and hard to read. This PR extracts checks into separate WindowCheck sub-classes and iterates over those instances, instead of creating and ORing a new function in for every exception. I also added a provision for excepting smaller-than-100x100 windows in the future.
- ColorSlurp's picked-color dialog is meant to behave as a full window, but has an abnormal role. This adds an exception check to allow AltTab to show the ColorSlurp window.

### DIFF
--- a/src/api-wrappers/AXUIElement.swift
+++ b/src/api-wrappers/AXUIElement.swift
@@ -7,6 +7,7 @@ import ApplicationServices.HIServices.AXAttributeConstants
 import ApplicationServices.HIServices.AXActionConstants
 
 // if the window server is busy, it may not reply to AX calls. We retry right before the call times-out and returns a bogus value
+
 func retryAxCallUntilTimeout(_ group: DispatchGroup? = nil, _ timeoutInSeconds: Double = Double(AXUIElement.globalTimeoutInSeconds), _ fn: @escaping () throws -> Void, _ startTime: DispatchTime = DispatchTime.now()) {
     group?.enter()
     BackgroundWork.axCallsQueue.async {
@@ -34,7 +35,10 @@ extension AXUIElement {
     // See https://humanbenchmark.com/tests/reactiontime
     static let retryDelayInMilliseconds = 250
 
+    static let minWindowSize = 100.0
+
     // default timeout for AX calls is 6s. We increase it in order to avoid retrying every 6s, thus saving resources
+
     static func setGlobalTimeout() {
         // we add 5s to make sure to not do an extra retry
         AXUIElementSetMessagingTimeout(AXUIElementCreateSystemWide(), globalTimeoutInSeconds + 5)
@@ -43,9 +47,9 @@ extension AXUIElement {
     func axCallWhichCanThrow<T>(_ result: AXError, _ successValue: inout T) throws -> T? {
         switch result {
             case .success: return successValue
-            // .cannotComplete can happen if the app is unresponsive; we throw in that case to retry until the call succeeds
+                // .cannotComplete can happen if the app is unresponsive; we throw in that case to retry until the call succeeds
             case .cannotComplete: throw AxError.runtimeError
-            // for other errors it's pointless to retry
+                // for other errors it's pointless to retry
             default: return nil
         }
     }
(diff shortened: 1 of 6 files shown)


src/api-wrappers/AXUIElement.swift
```diff
@@ -81,99 +85,16 @@ extension AXUIElement {
...
-    private static func androidEmulator(_ runningApp: NSRunningApplication, _ title: String?) -> Bool {
-        // android emulator small vertical menu is a "window" with empty title; we exclude it
-        return title != "" && Applications.isAndroidEmulator(runningApp)
+        return wid != 0 &&
+                size != nil && size!.width > minWindowSize && size!.height > minWindowSize &&
+                level == CGWindow.normalLevel &&
+                ([kAXStandardWindowSubrole, kAXDialogSubrole].contains(subrole) ||
+                        windowChecks.contains(where: { element in
+                            element.check(opts)
+                        }))
```
zacharee, line 97: This should have a similar short-circuiting effect compared to the old ORing method. If the checking order matters, it can be changed by rearranging the items in the `windowChecks` array.

src/logic/windowcheck/WindowCheck.swift
```diff
@@ -0,0 +1,175 @@
...
+    BattleNetBootstrapperWindowCheck(),
+    DrBetotteWindowCheck(),
+    DVDFabWindowCheck(),
+    SanGuoShaAirWDWindowCheck(),
+    SteamWindowCheck(),
+    FirefoxFullscreenVideoWindowCheck(),
+    VLCFullscreenVideoWindowCheck(),
+    AndroidEmulatorWindowCheck(),
+    ColorSlurpWindowCheck()
+]
```
zacharee, line 27: I'm not too familiar with Swift so there may be a better way collect all the subclasses.

src/logic/windowcheck/WindowCheck.swift
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     This allows for overriding the default requirements of windows being at least 100x100
+     to show in the switcher.
+     - Parameter size: the size of the window being checked.
+     - Returns: true if the window's size is correct.
+     */
+    func isValidSize(_ size: CGSize?) -> Bool {
+        return size != nil && size!.width > AXUIElement.minWindowSize && size!.height > AXUIElement.minWindowSize
+    }
```
zacharee, line 58: This could be useful for https://github.com/lwouis/alt-tab-macos/issues/1466

zacharee:
I also have the Ventura workaround commit here, but it should be possible to cherry-pick the other two commits.

src/api-wrappers/AXUIElement.swift
lwouis: I think you lost some of the nuances of the checks in the process of uniform-ing the checks. Apps like Books and Keynote are at a [different level of the boolean chain than the apps later on](https://github.com/lwouis/alt-tab-macos/pull/1871/files#diff-ba0d5f463a7a746a839b98f1f7ab668395bfea51a4a29dad57dae4b26237b6c9L86). Same thing for Jetbrains.

zacharee: Oops, you're right. I added in an `isValidLevel()` function to `WindowCheck` in the latest push.

lwouis:
@zacharee thank you for sharing this work.

I would like to see what you see in this refactor, but I'm afraid I don't. The business logic is moved from simple functions into something more complex involving a class hierarchy with subclass overriding methods, instantiating classes, etc.

I think if you find the current `AXUIElement.swift#isActualWindow` method noisy, you could create this `WindowCheck` file you created here, and simply move all the various functions in there to contain the "zoo". I'm not bothered by the methods and the current volume of lines in `AXUIElement` to be honest, but I would understand someone wanting to cut `isActualWindow` out of this file. However, I don't see the value in moving from straight-forward functions that are unique and customized to each business rule, into some generic system using a class hierarchy. I find for instance that each overriden method having the whole "`opts`" object passed *reduces* clarity, versus the current implem where each business function gets passed only what it needs to discriminate the window.

zacharee:
> @zacharee thank you for sharing this work.
> 
> I would like to see what you see in this refactor, but I'm afraid I don't. The business logic is moved from simple functions into something more complex involving a class hierarchy with subclass overriding methods, instantiating classes, etc.
> 
> I think if you find the current `AXUIElement.swift#isActualWindow` method noisy, you could create this `WindowCheck` file you created here, and simply move all the various functions in there to contain the "zoo". I'm not bothered by the methods and the current volume of lines in `AXUIElement` to be honest, but I would understand someone wanting to cut `isActualWindow` out of this file. However, I don't see the value in moving from straight-forward functions that are unique and customized to each business rule, into some generic system using a class hierarchy. I find for instance that each overriden method having the whole "`opts`" object passed _reduces_ clarity, versus the current implem where each business function gets passed only what it needs to discriminate the window.

The issue I was having with the original setup was parsing all of the nested groups of disjunctions and conjunctions and where a new exception should be added (for example, the JetBrains check was inverted vs. the rest).

There is more code, but imo now it's clearer how and when each step is checked, and adding a new exception just requires a new class and instantiating that class in the list, without needing to step through the parentheses to make sure it's placed correctly.

My inspiration was the [Strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern), which does have the drawback of needing more generic arguments, but the advantage of encapsulating similar functions into a predictable structure, i.e., prioritizing knowing where and when each one is used vs. knowing what each one does specifically.

Definitely up for more discussion on this.

lwouis:
Closing this PR as the ColorSlurp got merged from the other PR 👍

Issue #515, opened 2020-08-08
Sort windows by: creation date, manual order, etc


ghost:
Because AltTab is so smooth and performant, and due to its ability to close, minimise, hide, quit (etc.) windows all within the interface with either keyboard or mouse, it would be good to be able to organise one's entire workspace via AltTab. Re-arrange, and navigate, using keyboard or mouse or both.

lwouis:
Hey @mindfulsource, thank you for the suggestion!

You proposing that the user is able to drag the windows thumbnails to reorder them within AltTab. Here are some thoughts I have:

**Technical implementation**

Having drag-and-drop of the thumbnails be smooth, that is you move it and see it move in real-time, and as you drag it, other thumbnails move and reorganize to show you what the state would be if you dropped; that would require a huge amount of work. Honestly I don't think I could even do it. AppKit layout system is the worse I've ever worked on, and getting a smooth animation, while following the current layout rules (which are incredibly complex, due to the crazy amounts of customization the user can do, and screen sizes, screen ratio, vertical screens, etc).

**Value-proposition**

I don't get it to be honest. Why would one reorder the previews? You pop AltTab to access a window. If you open it and see the window is not close to the current selection, just go there, no?

The only use-case I can imagine is thinking "Ok now I will use these 2 apps together in back and forth, for example to do some copy-pasting from one to the other" and at this point in time the user pops up AltTab, and the 2 apps are at the bottom. Ok now it may be useful to drag them at the top so further switches are instant. However you can already do this scenario by actually switching to one app, then the other. At this point they are up in the order.

In summary, I don't see the value-proposition, and the work required would be enormous.

mindfulsource:
> Why would one reorder the previews?

Well, the idea and potential I've realised is that AltTab could not only be a pop-up program to find and go to another window, but actually to gain a whole sense of control and organisation over exactly what's open and how I'm working with all my windows.

Right now among my 27 windows I have 3 Firefox ones scattered across three different rows in my AltTab. Would be nice to organise things it might save some time and 'get organised'.

Or maybe I have two projects at once (both running for months) and I want one to be down the bottom, the other up the top.

(I just checked Mission Control for the first time since c. 2011 - it's crap, your interface is already 100x better than it at glancing at one's workspace, and imagine what more you could do with that interface!)

[...]

PR #1936, opened 2022-09-08
feat: Allow changing window sort from most recent to alphabetical

kadaan:
Add a now config option that allows the windows to be sorted recentFirst (as today) or alphabetical.

2022-09-08 at 11 22 30@2x

### DIFF
--- a/src/logic/Preferences.swift
+++ b/src/logic/Preferences.swift
@@ -31,6 +31,8 @@ class Preferences {
         "showHiddenWindows2": ShowHowPreference.show.rawValue,
         "showFullscreenWindows": ShowHowPreference.show.rawValue,
         "showFullscreenWindows2": ShowHowPreference.show.rawValue,
+        "sortWindows": SortHowPreference.lastUsed.rawValue,
+        "sortWindows2": SortHowPreference.lastUsed.rawValue,
         "showTabsAsWindows": "false",
         "hideColoredCircles": "false",
         "windowDisplayDelay": "0",
@@ -118,6 +120,7 @@ class Preferences {
     static var showMinimizedWindows: [ShowHowPreference] { ["showMinimizedWindows", "showMinimizedWindows2"].map { defaults.macroPref($0, ShowHowPreference.allCases) } }
     static var showHiddenWindows: [ShowHowPreference] { ["showHiddenWindows", "showHiddenWindows2"].map { defaults.macroPref($0, ShowHowPreference.allCases) } }
     static var showFullscreenWindows: [ShowHowPreference] { ["showFullscreenWindows", "showFullscreenWindows2"].map { defaults.macroPref($0, ShowHowPreference.allCases) } }
+    static var sortWindows: [SortHowPreference] { ["sortWindows", "sortWindows2"].map { defaults.macroPref($0, SortHowPreference.allCases) } }
     static var shortcutStyle: [ShortcutStylePreference] { ["shortcutStyle", "shortcutStyle2"].map { defaults.macroPref($0, ShortcutStylePreference.allCases) } }
     static var menubarIcon: MenubarIconPreference { defaults.macroPref("menubarIcon", MenubarIconPreference.allCases) }
 
@@ -426,6 +429,22 @@ enum ShortcutStylePreference: String, CaseIterable, MacroPreference {
         switch self {
             case .focusOnRelease: return NSLocalizedString("Focus selected window", comment: "")
             case .doNothingOnRelease: return NSLocalizedString("Do nothing", comment: "")
+        }
+    }
+}
+
+enum SortHowPreference: String, CaseIterable, MacroPreference {
+    case lastUsed = "0"
+    case lastOpened = "1"
+    case mostFrequently = "2"
+    case application = "3"
+
+    var localizedString: LocalizedString {
+        switch self {
+            case .lastUsed: return NSLocalizedString("Last Used", comment: "")
+            case .lastOpened: return NSLocalizedString("Last Opened", comment: "")
+            case .mostFrequently: return NSLocalizedString("Most Frequently Used", comment: "")
+            case .application: return NSLocalizedString("Name", comment: "")
         }
     }
 }
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -20,7 +20,22 @@ class Windows {
                let bool = sortByBooleanAttribute($0.isMinimized, $1.isMinimized) {
                 return bool
             }
-            return $0.lastFocusOrder < $1.lastFocusOrder
+            let sortType = Preferences.sortWindows[App.app.shortcutIndex]
+            var order: ComparisonResult = .orderedSame
+            if sortType == .mostFrequently {
+                order = sortByIntAttribute($1.focusCount, $0.focusCount)
+            } else if order == .orderedSame && sortType == .lastOpened {
+                order = sortByIntAttribute($0.lastOpenedOrder, $1.lastOpenedOrder)
+            } else if sortType == .application {
+                order = ($0.application.runningApplication.localizedName ?? "").localizedStandardCompare($1.application.runningApplication.localizedName ?? "")
+                if order == .orderedSame {
+                    order = ($0.title ?? "").localizedStandardCompare($1.title ?? "")
+                }
+            }
+            if order == .orderedSame || sortType == .lastUsed {
+                order = sortByIntAttribute($0.lastFocusOrder, $1.lastFocusOrder)
+            }
+            return order == .orderedAscending
         }
     }
 
@@ -73,6 +88,7 @@ class Windows {
     static func updateLastFocus(_ otherWindowAxUiElement: AXUIElement, _ otherWindowWid: CGWindowID) -> [Window]? {
         if let focusedWindow = (list.first { $0.isEqualRobust(otherWindowAxUiElement, otherWindowWid) }) {
             let focusedWindowOldFocusOrder = focusedWindow.lastFocusOrder
+            focusedWindow.focusCount += 1
             var windowsToRefresh = [focusedWindow]
             list.forEach {
                 if $0.lastFocusOrder == focusedWindowOldFocusOrder {
@@ -243,6 +259,16 @@ class Windows {
                 !(Preferences.screensToShow[App.app.shortcutIndex] == .showingAltTab && !window.isOnScreen(screen)) &&
                 (Preferences.showTabsAsWindows || !window.isTabbed))
     }
+}
+
+func sortByIntAttribute(_ i1: Int, _ i2: Int) -> ComparisonResult {
+    if i1 == i2 {
+        return .orderedSame
+    }
+    if i1 < i2 {
+        return .orderedAscending
+    }
+    return .orderedDescending
 }
 
 func sortByBooleanAttribute(_ b1: Bool, _ b2: Bool) -> Bool? {
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -1,8 +1,11 @@
 import Cocoa
 
 class Window {
+    static var maxOpenedOrder = Int.zero
     var cgWindowId: CGWindowID?
     var lastFocusOrder = Int.zero
+    var lastOpenedOrder = Int.zero
+    var focusCount = Int.zero
     var title: String!
     var thumbnail: NSImage?
     var thumbnailFullSize: NSSize?
@@ -45,6 +48,8 @@ class Window {
         self.position = position
         self.size = size
         self.title = bestEffortTitle(axTitle)
+        Window.maxOpenedOrder += 1
+        self.lastOpenedOrder = Window.maxOpenedOrder
         if !Preferences.hideThumbnails {
             refreshThumbnail()
         }
@@ -58,6 +63,8 @@ class Window {
         isWindowlessApp = true
         self.application = application
         self.title = application.runningApplication.localizedName
+        Window.maxOpenedOrder += 1
+        self.lastOpenedOrder = Window.maxOpenedOrder
         debugPrint("Adding app-window", title ?? "nil", application.runningApplication.bundleIdentifier ?? "nil")
     }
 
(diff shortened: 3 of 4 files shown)


kadaan:
@lwouis What do you think?

src/logic/Preferences.swift
```diff
@@ -193,6 +196,8 @@ class Preferences {
                             migrateMenubarIconFromCheckboxToDropdown()
                             // "Show minimized/hidden/fullscreen windows" checkboxes were replaced with dropdowns
                             migrateShowWindowsCheckboxToDropdown()
+                            // "Sort windows" checkboxes were replaced with dropdowns
+                            migrateSortWindowsCheckboxToDropdown()
```
lwouis, line 200: I'm a bit confused by this. Migration have been added in the past when we have saved a preference with the user already. Let's say we have a checkbox yes/no. Later, we change that checkbox to a dropdown menu with yes/no/maybe. We now need to **migrate** the users who already set is to yes or no, to the new serialization of yes/no from the dropdown.

In other words, migrations make sense *after* a feature has been introduced, when it evolves later. Here you're releasing the new preference, so no need to migrate anything.

src/logic/Preferences.swift
```diff
@@ -417,6 +436,18 @@ enum ShortcutStylePreference: String, CaseIterable, MacroPreference {
     }
 }
 
+enum SortHowPreference: String, CaseIterable, MacroPreference {
+    case recentFirst = "0"
+    case alphabetical = "1"
+
+    var localizedString: LocalizedString {
+        switch self {
+            case .recentFirst: return NSLocalizedString("Recent first", comment: "")
```
lwouis, line 445: There are a few more steps to do to make the localization work, but I will do them later on top of this PR, so no problem there.

src/logic/Windows.swift
```diff
@@ -8,7 +22,7 @@ class Windows {
 
     // reordered list based on preferences, keeping the original index
     static func reorderList() {
-        list.sort {
+        list = list.stableSorted {
```
lwouis, line 25: Could you share why this was needed?

src/logic/Windows.swift
```diff
@@ -20,10 +34,21 @@ class Windows {
...
+                        return false
+                    case .orderedSame:
+                        break
+                }
+            }
             return $0.lastFocusOrder < $1.lastFocusOrder
         }
     }
-
+
```
lwouis, line 51: Maybe remove that change? It doesn't look intentional

lwouis:
@kadaan thank you for sharing this PR! I left a few comments. Overall it seems good, and almost mergeable 👍 

There is an existing ticket for this feature request actually: https://github.com/lwouis/alt-tab-macos/issues/515. Would it be ok to add the third sorting preference: by creation date? The dropdown could read maybe:

```
Sorted by: * Recently focused first
           * Recently opened last
           * Alphabetical
```

src/logic/Preferences.swift
kadaan: Yeah, I figured that, but given this was my first Swift and I wanted to see something work, I was just trying to get it all wired up.  I'd love to not do it this way, but am unsure ATM how to write one with no migration.  I don't recall seeing one like that.

src/logic/Windows.swift
kadaan: My IDE does that...

src/logic/Preferences.swift
```diff
@@ -242,12 +247,12 @@ class Preferences {
     @available(OSX, deprecated: 10.11)
     private static func migrateLoginItem() {
         do {
-            let loginItems = LSSharedFileListCreate(nil, kLSSharedFileListSessionLoginItems.takeRetainedValue(), nil).takeRetainedValue()
-            let loginItemsSnapshot = LSSharedFileListCopySnapshot(loginItems, nil).takeRetainedValue() as! [LSSharedFileListItem]
+            let loginItems = LSSharedFileListCreate(nil, kLSSharedFileListSessionLoginItems.takeRetainedValue(), nil)!.takeRetainedValue()
```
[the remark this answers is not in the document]
kadaan: How am I to test it then :P

src/logic/Preferences.swift
kadaan: Figured it out...

src/logic/Windows.swift
kadaan: My thinking at the time was that we would prefer to have the windows consistently stay in the same order and that without a stable sort windows with the same name could arbitrarily change positions.  Given that I fallback to the recency ordering when they have the same title, there is no need for this...  Thoughts?

kadaan:
@lwouis Rather than 
```
Sorted by: * Recently focused first
           * Recently opened last
           * Alphabetical
```
it seems better to just have another dropdown that allows ascending/descending.

Thoughts?

kadaan:
@lwouis How about now?

kadaan:
I realize I am missing `Recently opened last`, that said I don't know where the data about when the window was opened is.  NSRunningApplication has launchDate,  but it says that is only available to applications launchControl.  If that is not what we need, then we will want to add a openedTimestamp to Window that gets set when we create it from `fileprivate func windowCreated(_ element: AXUIElement, _ pid: pid_t) throws {`

src/logic/Preferences.swift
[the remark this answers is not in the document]
lwouis: Remove the changes from thus commit, then rebase your git branch on top of `dev` and you will have this fixed from that branch.

src/logic/Windows.swift
lwouis: Yes indeed, it's already deterministic without this

lwouis:
> it seems better to just have another dropdown that allows ascending/descending.

I think it would be a bit overengineered. I can't imagine anyone wanting the reverse order on any of the 3 sorting criterias. No-one asked for these alternate orders in 2 years. I think the 3 labels i suggested will satisfy everyone involved in the tickets/discussions.

kadaan:
I guess the confusion is that you wrote both `focused first` and `opened last`.  To me, `focused first`  reads like the window that was focused furthest ago should be the first one displayed.  Whereas `opened last` reads like the most recently opened window should be the first one displayed.  Given that, having a `direction` dropdown seems to make sense.

If, on the other hand, you mean `most recently focused` and `most recently opened`, then I can see why the `direction` dropdown would be unneeded.

src/logic/Windows.swift
kadaan: To be clear. from my reading list.sort is not stable, so in the event that the result of a comparison is equal, then the order is non-deterministic.  That said, your focus Id seems to make this deterministic.

src/logic/Preferences.swift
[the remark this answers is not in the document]
kadaan: That is not the case for me on XCode 13.4.1.  I am unable to build/deploy without making these changes.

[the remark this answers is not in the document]
kadaan: Ohhhh...  `dev`...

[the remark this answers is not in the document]
kadaan: Ugh

lwouis:
Regarding the `Recently opened last` order: some people want windows to show like they would in the Windows taskbar. That is: windows stack from left to right, as you open then. This means a recently open window is all the way to the right.

It's the opposite as the recently-focused-first logic from alt-tab.

Is it more clear now?

lwouis:
> we will want to add a openedTimestamp to Window that gets set when we create it

Yes, you can either do that, or do similar to what i did for the focus order: you maintain an order of integers. Upside of the integers method is that you don't deal with dates/timestamps (bad perf, issues with timezones, etc). The downside is that you have to maintain the indexes like shifting some windows up and down and windows get created or deleted.

kadaan:
> > we will want to add a openedTimestamp to Window that gets set when we create it
> 
> Yes, you can either do that, or do similar to what i did for the focus order: you maintain an order of integers. Upside of the integers method is that you don't deal with dates/timestamps (bad perf, issues with timezones, etc). The downside is that you have to maintain the indexes like shifting some windows up and down and windows get created or deleted.

Yeah, I was planning on using epoch timestamp, which keeps it simple.

> > we will want to add a openedTimestamp to Window that gets set when we create it
> 
> Yes, you can either do that, or do similar to what i did for the focus order: you maintain an order of integers. Upside of the integers method is that you don't deal with dates/timestamps (bad perf, issues with timezones, etc). The downside is that you have to maintain the indexes like shifting some windows up and down and windows get created or deleted.

kadaan:
> Regarding the `Recently opened last` order: some people want windows to show like they would in the Windows taskbar. That is: windows stack from left to right, as you open then. This means a recently open window is all the way to the right.
> 
> It's the opposite as the recently-focused-first logic from alt-tab.
> 
> Is it more clear now?

Hmmm, yeah, I have no idea why anyone would want the most recently opened application to be the furthest to navigate to...  I just fail to see the benefit.

kadaan:
@lwouis The more I think about it the more useful it seems to have the most frequently focused windows first.  Those the the ones that the user is frequently switching to and therefor it reduces the switch time the most by putting them at the head of the list.

lwouis:
@kadaan yes, recency of use is really useful. That's why it's the current way AltTab works. But some people don't want to use this approach. They want a more static order: alphabetical, or by creation date like the Windows taskbar. I don't think it's a good approach, but different people have different preferences, so we are discussing adding that preference.

src/logic/Application.swift
```diff
@@ -125,7 +125,7 @@ class Application: NSObject {
     private func addWindows(_ axWindows: [(AXUIElement, CGWindowID, String?, Bool, Bool, CGPoint?, CGSize?)]) -> [Window] {
         let windows: [Window] = axWindows.compactMap { (axUiElement, wid, axTitle, isFullscreen, isMinimized, position, size) in
             if (Windows.list.firstIndex { $0.isEqualRobust(axUiElement, wid) }) == nil {
-                let window = Window(axUiElement, self, wid, axTitle, isFullscreen, isMinimized, position, size)
+                let window = Window(axUiElement, self, wid, axTitle, isFullscreen, isMinimized, position, size, Int(NSDate().timeIntervalSince1970))
```
lwouis, line 128: I think it would be easier to use a simple `Int` index here, like is done for `lastFocusOrder`. You can simply increment from the highest index when a new window is spawned. When windows are deleted, you don't even need to do anything, the order will work still. Regarding a potential overflow, it would deal with `9223372036854775807` (`Int.max`) windows being opened, before overflowing, so I think it's fine.

This would make the code easier, faster, and simpler.

kadaan:
@lwouis not sure if you saw, but I mentioned my desire for `most frequently focused` rather than `most recently focused`.  What do you think about that?

lwouis:
> @lwouis not sure if you saw, but I mentioned my desire for most frequently focused rather than most recently focused. What do you think about that?

How would you implement that? Having a counter for each window, incrementing the counter on each focus, and then sorting by these counters? I think it may not produce such a great UX.

I know that some apps like [`z`](https://github.com/rupa/z) "[frecency](https://en.wikipedia.org/wiki/Frecency)" (a mix of recency and frequency). Maybe that would bring a better UX for users who care about frequency, rather than raw frequency.

Just to clarify: we are talking about adding another sorting option here, right? There would be: recency (default), frecency/frequency, open date, alphabetical.

decodism:
May I suggest splitting the sorting into apps and windows, sorting apps first and then windows for each app.

The default would be:
- Apps: none
- Windows: recently focused

lwouis:
@decodism it's interesting. There are 3 types of thumbnails that are not discussed:

* Apps thumbnails
* Minimized windows
* Hidden windows

The last 2 items can either be sorted like the other thumbnails, or be moved at the end of the list by changing these dropdowns:

image

For the first item (apps thumbnails), they are always at the end.

There may be a way to integrate these 3 items in the new system introduced by this PR. There would then be a one-stop shop to change sorting. On the other side, it may make it unnecessarily complex. Not sure.

tonykaram1993:
Any updates on this? For my current workflow I set up a shortcut to switch between windows in the same space and screen. Like so:
CleanShot 2022-11-26 at 00 42 43@2x
However I prefer the ordering to be static and not change based on usage/recency specifically for this shortcut. I do want recency for CMD+TAB (shortcut 1)

TheFoobarGuru:
Hello,
Any reason why delay the merge of the PR ? I would really appreciate this new feature.
Thank you

lwouis:
@TheFoobarGuru @tonykaram1993 this PR needs more work. As it stands, it doesn't take into account some elements. [More details in my last comment](https://github.com/lwouis/alt-tab-macos/pull/1936#issuecomment-1243351921). We need @kadaan to continue work on this, or someone else to maybe take over this work, and bring it to the finish like in another PR.

XOR-op:
@lwouis Could you explain more about your last comment? Since this PR is inactive, I want to open another PR to implement this feature and solve your comment. However, I'm not clear if you mean we should integrate these into sort option. Integrating all these settings into one sort option will lead to a 4x3x3=36 permutation, which is unaccpetable. And I don't see why they are relevant. Or do you mean this PR does not respect these settings?

lwouis:
Hi @XOR-op,

This PR adds a new dropdown, bellow the existing ones. It looks like this:

image

If I select in the new dropdown `Name`, but then for the Minimized Window, I pick `Show at the End`, then I'm not sure it's clear what should happen. Maybe non-minimized windows are listed first, alphabetically, and after them come the minimized windows, also listed alphabetically, separately. That could be acceptable. Maybe I was overthinking it. I don't think most users want that fine of a control like "i want the normal windows first sorted by name, then I want the hidden ones sorted by last open first, and finally I want the minimized ones sorted by last created first". I guess the approach in this PR could be good enough.

The implementation doesn't do that though. The current after `at the end` is random. Note that the current implementation suffers from this randomness as well. But having an explicit sort will make that more obvious to users who would expect their order to be respected, in the `at the end` sections as well.

Some other issues with the current PR:

* Sorting by frequency is too plain in my opinion. I mentionned above that [frecency may be useful](https://github.com/lwouis/alt-tab-macos/pull/1936#issuecomment-1242945354). I don't see how raw frequency would be a good idea as a UX.
* The sorting code is broken because "Last Used", "Last Opened" are concepts that mean different things for a Window vs an App. An app will show in AltTab if you close the last window of that app. You can then create a new window, then close it. That makes the app dissapear and reappear from AltTab. It doesn't changed when the user opened the app though.

I hope this was helpful 🙇‍♂️

XOR-op:
Thanks for your explanation! I'll give it a try to fix these problems.

lwouis:
Closing this in favor of @XOR-op's PR which got merged today as well 👍

Issue #1458, opened 2022-03-31
Having more than two shortcuts


daniel-zeng:
Hello,

Currently, it seems that there's only two customizable shortcuts, and I've searched the Github issues but I don't seem to find anything to add additional shortcuts. 

Specifically, I wanted to have a way to have different types of release behavior, but using the two shortcuts for ("All Apps" and "Active App") for "Focus selected" takes up these two slots and I also wanted similar for the other option as well.

Thank you!

PrayagS:
+1 on this.

GJRobert:
I have also suggested in #1064

lwouis:
Closing as duplicate of #1064

Issue #967, opened 2021-05-24
Add a full screen live preview when cycling through the apps


mlubej:
**Is your feature suggestion related to a problem? Please describe.**
Sometimes seeing just the icon or the small window tile is not enough, I'd like to see the focused app also focused for real, no matter which space/display it occupies.

**Describe alternatives you've considered**
Hyper-switch has a similar functionality, but it's very slow and it ignores any preferences I specify.

lwouis:
It would be an interesting feature. It would need to be shaped a bit more though. Would we show it on mouse-hover like HyperSwitch, or maybe just when cycling through the thumbnail with the keyboard shortcut?

Regarding implementation, I'm not sure how they do it. I was guessing that they create a fake window and print the screenshot of the actual window on it. But then I took a slow-motion video, and the animation is seamless. This gets me thinking that they may be using the private SPI to bring a window up without focusing it (focus order is not messed up by the preview; as is the top left app name not changing).

mlubej:
> It would be an interesting feature. It would need to be shaped a bit more though. Would we show it on mouse-hover like HyperSwitch, or maybe just when cycling through the thumbnail with the keyboard shortcut?

I guess both, or related to the focus of the app, so if the app is selected (either with cycling through or with the mouse), the app is brought to the front.

I'm not exactly sure how to achieve this. I'm fairly new to mac and just thought I'd make the suggestion though. I really like the tool!

safield:
I think it would be nice if the window associated with the currently selected thumbnail would simmply be brought to the foreground as you cycled through the thumbnails.

XInTheDark:
This feature is very similar to the one that Windows offers, and it has proven quite useful:

https://user-images.githubusercontent.com/61100393/196666609-783e1e5f-6cd3-445f-8cd5-5f54164419a8.mov

bishshoy:
I have got a great idea to have this feature implemented, without requiring too many modifications to the existing way of doing things. Instead of making the selected app come to foreground, AltTab could draw the screenshot of the app on the location of the app itself. If the app is behind another app, then it could draw it over the top app. The size of the screenshot would match the size of the window. And every time the user presses Cmd+Tab, the screenshot and the location of the screenshot gets updated. Also, at any given time only one app's screenshot is displayed on the switcher pane. This would make the user feel that AltTab is switching amongst the windows in realtime. Makes sense?

Issue #529, opened 2020-08-13
Always-on UI (i.e. bar or dock)


lwouis:
I've been thinking about alternatives such as HyperDock and uBar. They provide always-on UIs:

![image](https://user-images.githubusercontent.com/106195/90106168-00f89700-dd82-11ea-85e3-aad910dc5d28.png)

![image](https://user-images.githubusercontent.com/106195/90106198-09e96880-dd82-11ea-9879-b454a92f72b1.png)

This type of bar / Dock is something I miss from Windows and Linux. By virtue of being always-on, they allow users to get a spacial sense of where windows are, as time goes by. AltTab currently shows chronological order of focus. An always-on display can display something very interesting: chronological order of creation. Windows stack into the bar as they are created, but then stay in place when interacted with. It's a different UX which allows user to get spacial awareness by seeing windows being added in the bar, and remembering the place. Later when they want to switch they can simply click that area of the bar.

I think it's a cool UX direction to explore. I think it could be a separate app, or a feature of AltTab. Please share feedback if you have some! :)

rxhfcy:
I'm pretty sure at least these kinds of potential different "customers" exist:
- people who want an exact bug-for-bug copy of the Windows 10 UX
- people who want an uBar clone
- people who want a HyperDock clone

I haven't used HyperDock or uBar so I don't have an informed opinion on them. My own personal take is that since there are **lots** of current and former Windows users, maybe make something that's familiar to them?

I have a feeling it's impossible to make everybody completely happy, so you could just start by implementing what *you* personally want and iterate from there.

PS. I have no idea what the preferences window would look like for AltTab with this kind of functionality included (confusing?), so maybe a separate app, at least in the beginning?

j0nnala:
Integrating with native dock will be super cool. There is no app currently in the market that can do that.

I know it easier said than done. It would be like having additional triggering point on mouse over on bar icon with the option of "show same following windows" to "Mouse over dock app"

lix0x7:
This is really cool and what I am looking for. There are no alternative apps like this except for uBar. I used to be a uBar user, but there are lots of bug which I don't know why they exist. The company seems like that they have abandoned the uBar project. The uBar even couldn't show most of running apps normally. From my perspective, the uBar seems like that it cannot deal with multi-tab apps well, which leads to endless bugs. 
[...]

PR #2010, opened 2022-10-12
feat: add 3 more shortcut tabs

decodism:
Add more shortcut tabs and an option to interact with the Dock.

### DIFF
--- a/src/logic/Preferences.swift
+++ b/src/logic/Preferences.swift
@@ -13,8 +13,14 @@ class Preferences {
         "fontHeight": "15",
         "holdShortcut": "⌥",
         "holdShortcut2": "⌥",
+        "holdShortcut3": "⌥",
+        "holdShortcut4": "⌥",
+        "holdShortcut5": "⌥",
         "nextWindowShortcut": "⇥",
         "nextWindowShortcut2": keyAboveTabDependingOnInputSource(),
+        "nextWindowShortcut3": "",
+        "nextWindowShortcut4": "",
+        "nextWindowShortcut5": "",
         "focusWindowShortcut": "Space",
         "previousWindowShortcut": "⇧",
         "cancelShortcut": "⎋",
@@ -27,10 +33,19 @@ class Preferences {
         "cursorFollowFocusEnabled": "false",
         "showMinimizedWindows": ShowHowPreference.show.rawValue,
         "showMinimizedWindows2": ShowHowPreference.show.rawValue,
+        "showMinimizedWindows3": ShowHowPreference.show.rawValue,
+        "showMinimizedWindows4": ShowHowPreference.show.rawValue,
+        "showMinimizedWindows5": ShowHowPreference.show.rawValue,
         "showHiddenWindows": ShowHowPreference.show.rawValue,
         "showHiddenWindows2": ShowHowPreference.show.rawValue,
+        "showHiddenWindows3": ShowHowPreference.show.rawValue,
+        "showHiddenWindows4": ShowHowPreference.show.rawValue,
+        "showHiddenWindows5": ShowHowPreference.show.rawValue,
         "showFullscreenWindows": ShowHowPreference.show.rawValue,
         "showFullscreenWindows2": ShowHowPreference.show.rawValue,
+        "showFullscreenWindows3": ShowHowPreference.show.rawValue,
+        "showFullscreenWindows4": ShowHowPreference.show.rawValue,
+        "showFullscreenWindows5": ShowHowPreference.show.rawValue,
         "showTabsAsWindows": "false",
         "hideColoredCircles": "false",
         "windowDisplayDelay": "0",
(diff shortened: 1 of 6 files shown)


decodism:
With the Dock options enabled (selection by hovering and showing by shortcut):

https://user-images.githubusercontent.com/77468771/195438196-435b41e0-4865-4e5f-a694-dd5a4fd4ebb2.mov

lwouis:
The 5 shortcuts addresses #1458 so it's an easy merge. The other feature, I'm afraid seems very niche. I don't think many people would use it, and it adds quite a lot of complexity in the preferences UI. I'm already struggling to reach the goals set in #351

decodism:
I answer for here and #2012: I can't evaluate the potential use of these features but I think having a lot of options is not an issue, especially for an app aimed at rather advanced users, as long as the defaults are adequate and these options are clear and well categorized, which might be a topic for #351.

lwouis:
> I think having a lot of options is not an issue [...] as long as the defaults are adequate and these options are clear and well categorized

I think even clear and categorized preferences are still an issue in-and-of-itself. For example, a dropdown with 50 options is overwhelming for most people. Even if you reduce it to 5 options that each lead to 10 sub-options as a way to manage complexity, I think that system is still intimidating vs agreeing on the top 5 options most users use, and having only these 5 in the dropdown.

I made the "mistake" with AltTab of adding too much customization early on. Based on 2 years of reading user feedback here and on other channels, I think the community at large would be better off with a few presets in terms of how the app looks, behave, and the use-cases people have with it. This is the purpose of #351. I hope this context helps you see my position

src/logic/Preferences.swift
```diff
@@ -373,6 +400,21 @@ class Preferences {
...
+    
+    static func nameToIndex(_ name: String) -> Int {
+        var index = name.endIndex
+        while index > name.startIndex {
+            let newIndex = name.index(before: index)
+            guard name[newIndex].isWholeNumber else { break }
+            index = newIndex
+        }
+        let suffix = name.suffix(from: index)
+        return suffix == "" ? 0 : Int(suffix)! - 1
```
lwouis, line 416: Maybe something simpler would work as well here? I'm thinking to check the last char directly like:

```swift
if name[name.endIndex].isWholeNumber {
  return  Int(name[name.endIndex])
} else {
  return 0
}
```

src/ui/generic-components/CustomRecorderControl.swift
```diff
@@ -86,15 +86,14 @@ class CustomRecorderControl: RecorderControl, RecorderControlDelegate {
...
+                let index = Preferences.nameToIndex(id)
+                let otherIndex = Preferences.nameToIndex($0.id)
                 if id.starts(with: "holdShortcut") {
                     return Preferences.nextWindowShortcut[index] == Preferences.nextWindowShortcut[otherIndex] &&
-                        shortcut.modifierFlags == ControlsTab.shortcutControls["holdShortcut" + otherSuffix]!.0.objectValue!.modifierFlags
+                        shortcut.modifierFlags == ControlsTab.shortcutControls[Preferences.indexToName("holdShortcut", otherIndex)]!.0.objectValue!.modifierFlags
                 }
                 if id.starts(with: "nextWindowShortcut") {
-                    if let nextWindowShortcut = ControlsTab.shortcutControls["nextWindowShortcut" + otherSuffix]?.0.objectValue {
+                    if let nextWindowShortcut = ControlsTab.shortcutControls[Preferences.indexToName("nextWindowShortcut", otherIndex)]?.0.objectValue {
```
lwouis, line 96: Thank you for doing this house cleaning after you generalized the index system!

src/ui/preferences-window/tabs/ControlsTab.swift
```diff
@@ -41,14 +47,23 @@ class ControlsTab {
...
+        let (holdShortcut2, nextWindowShortcut2, tab2View) = toShowSection(1)
+        let (holdShortcut3, nextWindowShortcut3, tab3View) = toShowSection(2)
+        let (holdShortcut4, nextWindowShortcut4, tab4View) = toShowSection(3)
+        let (holdShortcut5, nextWindowShortcut5, tab5View) = toShowSection(4)
+        let tabView = TabView([
+            (NSLocalizedString("Shortcut 1", comment: ""), tab1View),
+            (NSLocalizedString("Shortcut 2", comment: ""), tab2View),
+            (NSLocalizedString("Shortcut 3", comment: ""), tab3View),
+            (NSLocalizedString("Shortcut 4", comment: ""), tab4View),
+            (NSLocalizedString("Shortcut 5", comment: ""), tab5View),
```
lwouis, line 60: I thought we may use a variable like `"Shortcut %@"` here to avoid having each language translate 5 times the same thing, but then I checked, and some languages wouldn't work with this pattern (e.g. `"Shortcut 2" = "الإختصار ٢";`). Oh well 😅

src/logic/Preferences.swift
decodism: It's done.

XInTheDark:
Can't we make this process of adding more shortcut tabs simpler, by adding a `+` button to the right of "Shortcut 2" (i.e. here:
image
So that the user can add as many shortcuts they wish, instead of setting a fixed limit?

Also, I think this is a great feature (which ironically is directly copied from Windows):
> With the Dock options enabled (selection by hovering and showing by shortcut):

A lot of Windows users who switched to Mac would appreciate this handy feature.

XInTheDark:
> I made the "mistake" with AltTab of adding too much customization early on. Based on 2 years of reading user feedback here and on other channels, I think the community at large would be better off with a few presets in terms of how the app looks, behave, and the use-cases people have with it.

Hmm, I have to politely disagree with this point; IMO, the options added to AltTab preferences allow users to customise the app better, and users who prefer not to customise the app can simply choose to ignore the preferences.

XInTheDark:
(the feature on windows:)

https://user-images.githubusercontent.com/61100393/196595262-4388eccd-4278-4721-a222-78166c1bae41.mov

It's definitely a good feature, and maybe we should polish it (e.g. add pop-up animations to the apparition, and fade-out once the window is selected) to make it look even better.

lwouis:
@XInTheDark what you show in the video is closer to what has been discussed in #529 #337 #967. What @decodism had proposed in this PR is that after hovering the Dock with the mouse, it would influence the next AltTab trigger, which would only show windows of the hovered app. It's quite different from hovering the Dock, and getting previews of the windows above the Dock, as in your video.

lwouis:
> Can't we make this process of adding more shortcut tabs simpler, by adding a + button to the right of "Shortcut 2"

It's actually easier both for the developpers to implement, and for the users, to have 5 shortcuts. It's less complex, flexible, changing. It's simple, set, static. In your example UI, we could see arise questions such as: what happens when we reach 5: the plus buttons disapears? How would I remove tabs to go back to 2 tabs?

@decodism implementation provides the features with a simpler UI. It's 5 shortcuts, with the last 3 not set by default. I think it's a simple solution which has the very important property of being real since he implemented it. We discussed dynamic UIs in the past, but no-one took on to do it. The best property of an idea, in a volonteers project, is to be worked on and made concrete.

decodism:
@lwouis

> it would influence the next AltTab trigger

Just to be clear: it's a dedicated shortcut

lwouis:
I have cherry-picked the commit on the next-release branch. Thank you again for sharing this work!

PR #2013, opened 2022-10-12
fix: wrap thumbnail buttons when needed

decodism:
Do not let the thumbnail buttons extend beyond the thumbnail.

### DIFF
--- a/src/ui/main-window/ThumbnailView.swift
+++ b/src/ui/main-window/ThumbnailView.swift
@@ -74,16 +74,21 @@ class ThumbnailView: NSStackView {
             isShowingWindowControls = shouldShow
             let target = (window_?.isWindowlessApp ?? true) ? windowlessIcon : thumbnail
             target.addSubview(quitIcon, positioned: .above, relativeTo: nil)
-            var offset = CGFloat(0)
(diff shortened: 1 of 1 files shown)


decodism:
OK, here are some images.

Before:
thumbnail_buttons_before

After:
thumbnail_buttons_after

lwouis:
Thank you!

Does it handle the other patterns?

```
xxx
x
```

or


```
x
x
x
x
```

And what happens when the thumbnail is too short vertically? Does it clip? How does it clip?

I'm asking not because I think it's important to support those (it's ok to clip in extreme cases like that), but because I want to avoid that some scenarios break the layout code and either crash or create a bigger layout issue. Essentially it would reassure me that this was "QA'd" for edge-cases to avoid breaking existing users.

decodism:
> Does it handle the other patterns?

Yes

> And what happens when the thumbnail is too short vertically? Does it clip? How does it clip?

The same behavior as now: they overflow.

lwouis:
I tested and indeed I think it's a straight improvement. Thank you! I'm closing this PR. I cherry-picked the commit on the release branch so it will be part of the next release 👍

Issue #1540, opened 2022-05-03
Better tabs detection


lwouis:
Today, AltTab uses a full bag of tricks to detect windows which are grouped as tabs. These tricks have limitations that result in features we can't provide, and bugs where AltTab fails to show or not show certain windows.

### Current implementation (very simplified and high-level)

* AltTab subscribe to the Accessibility event called `kAXFocusedUIElementChangedNotification` on every app
* The event triggers when the user interacts with the app UI in almost any way
* On every trigger, we make an API call to ask for the app windows
* If the windows returned by the API are different from some of the windows we already were tracking, it means some windows just became tabs

### Limitations / Bugs

* If there are multiple tab-groups for a given app, AltTab doesn't know which tab-group has which tabs. All it knows is which windows are the active tabs of the tab-groups.
* Tab detection seems to be flacky, especially when dealing with animations like Space transitions, windows going full-screen, etc. Overall, the `kAXFocusedUIElementChangedNotification` doesn't cover all use-cases for tabs, and doesn't happen always at the right time (i.e. it may trigger too early or too late, and we get the wrong understanding of the OS state)

lwouis:
FYI, we can't use the Accessibility API to simply ask the tab-group for its children. The API simply doesn't return children windows. We can only ask children windows of the whole app, and we will get a single list, with no sorting depending on which tab-group the windows are in.

Furthermore, the OS handles tabs in a lazy way. A window becoming a tab will not be resized, repositioned, and redrawned, until it is clicked/seen by the user. So after a user makes a window a tab, if they don't click on it, and we make an Accessibility call to get its size, position, or screenshot, we will get the values of when it used to be a window - so completely wrong.

lwouis:
I looked into the following APIs:

```swift
SLSCopyAssociatedWindows
```

It returns the same ID as the one given as input

```swift
let query = SLSWindowQueryWindows(cgsMainConnectionId, [try currentWindows![0].cgWindowId()!] as CFArray, 1)
let iterator = SLSWindowQueryResultCopyWindows(query)
while SLSWindowIteratorAdvance(iterator) == .success {
  let parent_wid = SLSWindowIteratorGetParentID(iterator)
  let wid = SLSWindowIteratorGetWindowID(iterator)
  let tags = SLSWindowIteratorGetTags(iterator)
}
```

`SLSWindowIteratorGetParentID` returns `0` (window is top level / has no parent, I guess)
[...]

PR #2020, opened 2022-10-15
fix: tab detection

### LINKED ISSUE #2017: AltTab always shows tabs as separate window, when the window is hidden/minimized, on Monterey
**Describe the bug**

When the (1) "Show Standard tabs as windows" option is disabled in the Appearance options, and (2) the user has elected to show "Hidden windows" from the Control options, AltTab ignores the first option. In other words, it still shows standard tabs as windows. To be clear, this only occurs when the window is hidden. Otherwise, AltTab performs as expected, and only shows windows.

**Screenshots / video**

When the window isn't hidden, AltTab performs exactly as it should—displaying only the window. This is illustrated in the screenshot below using a Finder window with two tabs.

Not Hidden

However, once you hide the window, AltTab will now show both of Finder's tabs as windows—effectively ignoring the disabled option to "Show Standard tabs as windows".

When Hidden copy

1. Controls > Hidden windows > Show
2. Appearance > Show Standard tabs as windows > Uncheck/Disable
3. Minimize an app that uses standard tabs, like Finder, and then trigger AltTab. You'll notice that its tabs are being treated like windows.

decodism:
Tries to fix #2017.

### DIFF
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -229,10 +229,21 @@ class Windows {
     /// tabs detection is a flaky work-around the lack of public API to observe OS tabs
     /// see: https://github.com/lwouis/alt-tab-macos/issues/1540
     static func detectTabbedWindows() {
-        let cgsWindowIds = Spaces.windowsInSpaces(Spaces.idsAndIndexes.map { $0.0 })
+        lazy var cgsWindowIds = Spaces.windowsInSpaces(Spaces.idsAndIndexes.map { $0.0 })
+        lazy var visibleCgsWindowIds = Spaces.windowsInSpaces(Spaces.idsAndIndexes.map { $0.0 }, false)
         list.forEach {
             if let cgWindowId = $0.cgWindowId {
-                $0.isTabbed = !cgsWindowIds.contains(cgWindowId)
+                if $0.isMinimized || $0.isHidden {
+                    if #available(macOS 13, *) {
+                        // not exact after window merging
+                        $0.isTabbed = !cgsWindowIds.contains(cgWindowId)
+                    } else {
+                        // not known
+                        $0.isTabbed = false
+                    }
+                } else {
+                    $0.isTabbed = !visibleCgsWindowIds.contains(cgWindowId)
+                }
             }
         }
     }
--- a/src/api-wrappers/PrivateApis.swift
+++ b/src/api-wrappers/PrivateApis.swift
@@ -112,10 +112,10 @@ func CGSCopyManagedDisplaySpaces(_ cid: CGSConnectionID) -> CFArray
 
 struct CGSCopyWindowsOptions: OptionSet {
     let rawValue: Int
-    static let minimizedAndTabbed = CGSCopyWindowsOptions(rawValue: 1 << 0)
+    static let invisible1 = CGSCopyWindowsOptions(rawValue: 1 << 0)
     // retrieves windows when their app is assigned to All Spaces, and windows at ScreenSaver level 1000
     static let screenSaverLevel1000 = CGSCopyWindowsOptions(rawValue: 1 << 1)
-    static let minimized2 = CGSCopyWindowsOptions(rawValue: 1 << 2)
+    static let invisible2 = CGSCopyWindowsOptions(rawValue: 1 << 2)
     static let unknown1 = CGSCopyWindowsOptions(rawValue: 1 << 3)
     static let unknown2 = CGSCopyWindowsOptions(rawValue: 1 << 4)
     static let desktopIconWindowLevel2147483603 = CGSCopyWindowsOptions(rawValue: 1 << 5)
--- a/src/logic/Spaces.swift
+++ b/src/logic/Spaces.swift
@@ -69,11 +69,14 @@ class Spaces {
         return idsAndIndexes.filter { $0.0 != currentSpaceId }.map { $0.0 }
     }
 
-    static func windowsInSpaces(_ spaceIds: [CGSSpaceID]) -> [CGWindowID] {
+    static func windowsInSpaces(_ spaceIds: [CGSSpaceID], _ includeInvisible: Bool = true) -> [CGWindowID] {
         var set_tags = ([] as CGSCopyWindowsTags).rawValue
         var clear_tags = ([] as CGSCopyWindowsTags).rawValue
-        let options = ([.minimizedAndTabbed, .screenSaverLevel1000] as CGSCopyWindowsOptions).rawValue
-        return CGSCopyWindowsWithOptionsAndTags(cgsMainConnectionId, 0, spaceIds as CFArray, options, &set_tags, &clear_tags) as! [CGWindowID]
+        var options = [.screenSaverLevel1000] as CGSCopyWindowsOptions
+        if includeInvisible {
+            options = [options, .invisible1, .invisible2]
+        }
+        return CGSCopyWindowsWithOptionsAndTags(cgsMainConnectionId, 0, spaceIds as CFArray, options.rawValue, &set_tags, &clear_tags) as! [CGWindowID]
     }
 
     static func isSingleSpace() -> Bool {


lwouis:
@decodism this seems like a draft. Is this WIP that you're sharing early?

decodism:
It wasn't but I've reworked it to include an improvement of #2021. It is now possible to minimize and fullscreen tabs if the app only has one tab group in the space.

src/logic/events/AccessibilityEvents.swift
```diff
@@ -130,9 +136,9 @@ fileprivate func focusedWindowChanged(_ element: AXUIElement, _ pid: pid_t) thro
 
 fileprivate func windowDestroyed(_ element: AXUIElement) throws {
     let wid = try element.cgWindowId()
-    DispatchQueue.main.async {
-        if let index = (Windows.list.firstIndex { $0.isEqualRobust(element, wid) }) {
-            let window = Windows.list[index]
+    if let index = (Windows.list.firstIndex { $0.isEqualRobust(element, wid) }) {
+        let window = Windows.list[index]
+        DispatchQueue.main.async {
```
lwouis, line 141: I think it's dangerous to call objects like `Windows.list` from the AX background thread. That's why all the functions in this file have a pattern of doing a few lines of AX calls (all the `try`s), then doing work inside a `DispatchQueue.main` block.

src/ui/App.swift
```diff
@@ -248,7 +248,7 @@ class App: AppCenterApplication, NSApplicationDelegate {
             isFirstSummon = false
             if Windows.list.count == 0 || CGWindow.isMissionControlActive() { hideUi(); return }
             // TODO: can the CGS call inside detectTabbedWindows introduce latency when WindowServer is busy?
-            Windows.detectTabbedWindows()
+            //Windows.detectTabbedWindows()
```
lwouis, line 251: You replaced this unconditional call, before showing the UI, with a similar call, but only when a window gets deleted. I think it will miss many cases where tabs are reconfigured. I was originally tracking tabs in the events (window resized, created, deleted, etc). Then I faced many subtle issues where tabs wouldn't properly be noticed.

I ended up putting an unconditional call upfront before showing the UI, and it proved way more reliable. I'm worried here that you are going back to the old approach, and it will not pick up the right state. The user can combine tabs with the mouse, with keyboard shortcuts, by closing windows, by creating windows, etc.

src/logic/Window.swift
```diff
@@ -116,6 +116,12 @@ class Window {
 
     func minDemin() {
         if isWindowlessApp { return }
+        guard !isTabbed else {
+            if let tabHoldingWindows = application.tabHoldingWindows[spaceId], tabHoldingWindows.count == 1 {
+                tabHoldingWindows.first!.minDemin()
+            }
+            return
+        }
```
lwouis, line 124: I think it would be best to hide the buttons, and not try to do things that the OS itself doesn't allow the user to do. This is in line with our previous discussions here: https://github.com/lwouis/alt-tab-macos/pull/2021#issuecomment-1284711792

src/logic/events/AccessibilityEvents.swift
decodism: Fixed.

src/ui/App.swift
decodism: The issue with this new method compared to the old one is that it only works for the current space. So calling it when the UI is displayed would not be sufficient. It is called for the application when a window of the application is focused or deleted. The only case that is not covered that I can think of is when AltTab closes a window that is part of a group of tabs in another space.

src/logic/Window.swift
decodism: It seems to me that AltTab already does things not allowed by the OS, depending on your definition. My issue with this though is that it only covers one case.

lwouis: Yes, I meant that if we can't have great support for some approach, it's probably better to give up on it, and it shouldn't be too missed by users since they are used to not having such capabilities in the first place, from the native OS UI.

src/ui/App.swift
lwouis: I'll give you an example:

* Create 2 Finder windows
* Click in the menubar at the top: `Window > Merge All Windows`
* This will not trigger any accessibility events (except `kAXFocusedUIElementChangedNotification` which we used to listen to, but no longer do)
* It will thus never trigger `detectTabsInCurrentSpace()`
* AltTab will not be aware of this change in tabs/windows

decodism:
The first approach being too flawed, I changed to a simpler one.

src/logic/Windows.swift
```diff
@@ -162,8 +162,8 @@ class Windows {
     static func detectTabbedWindows() {
         let cgsWindowIds = Spaces.windowsInSpaces(Spaces.idsAndIndexes.map { $0.0 }, [])
         list.forEach {
-            if let cgWindowId = $0.cgWindowId {
-                $0.isTabbed = !$0.isMinimized && !$0.isHidden && !cgsWindowIds.contains(cgWindowId)
+            if let cgWindowId = $0.cgWindowId, !$0.isMinimized && !$0.isHidden {
```
lwouis, line 165: I think one scenario that this would struggle with is when a window with tabs has been minimized/hidden before AltTab runs. AltTab would then shows those tabs as windows since their `isTabbed` would be at default value `false`

decodism: The bigger flaw is that it does not take into account the creation/deletion of a group of tabs and its minimizing/hiding between two invocations of AltTab.

lwouis: I've experienced with detecting tabs better.

I found that there is no need to check for minimized/hidden. If we use `CGSCopyWindowsWithOptionsAndTags` with the `.minimizedAndTabbed` option (and actually use the option in the method).

It seems to handle every case I tested.

Except one of course. When you make 2 windows of 2 tabs each, then use the menubar action `Window > Merge All Windows`. Then, even though there is now only 1 window with 4 tabs, `CGSCopyWindowsWithOptionsAndTags` still returns an array containing the 2 tabs that were the visible tabs of the previous 2 windows. Even though now there is only 1 active tab in the 4-tabs window, the API still returns 2 tabs.

I experimented with using `axUiElement.children()`, and checking if there is a child with `role() == kAXTabGroupRole`. First of all, using AX calls is really bad when the user invokes AltTab. We want to avoid it, as these calls can resolve **seconds** later sometimes. That's why we normally process them on a background thread. But even if I ignore this technical issue, it's still not enough because it doesn't let us distinguish a window with no tabs (has no tabs-bar), from a non-visible tab (has no tabs-bar).

I also re-read, and advise you do the same, this ticket were I explained many approaches I tried in the past: https://github.com/lwouis/alt-tab-macos/issues/1540

My conclusion for now is that there is no perfect way still. All approaches have issues.

I think I'll release a variation of this commit, with the `minimized`/`hidden` checks removed, and instead adding `.minimizedAndTabbed` to `CGSCopyWindowsWithOptionsAndTags`. That seems to me like the most robust/comprehensive check we can do for the time being.

decodism: I had not caught this exception. It can be partially mitigated by using the option only for minimized or hidden windows.

src/logic/Windows.swift
```diff
@@ -159,17 +159,23 @@ class Windows {
     }
 
     static func detectTabbedWindows() {
-        let cgsWindowIds = Spaces.windowsInSpaces(Spaces.idsAndIndexes.map { $0.0 }, [])
+        let cgsWindowIdsIncludingMinimizedAndHidden = Spaces.windowsInSpaces(Spaces.idsAndIndexes.map { $0.0 }, true)
+        let cgsWindowIds = Spaces.windowsInSpaces(Spaces.idsAndIndexes.map { $0.0 })
         list.forEach {
             if let cgWindowId = $0.cgWindowId {
-                $0.isTabbed = !$0.isMinimized && !$0.isHidden && !cgsWindowIds.contains(cgWindowId)
+                if #available(macOS 13, *), $0.isMinimized || $0.isHidden {
```
decodism, line 166: @lwouis Maybe implement this check for #2136.

src/logic/Windows.swift
```diff
@@ -162,9 +162,20 @@ class Windows {
...
+                if $0.isMinimized || $0.isHidden {
+                    if #available(macOS 13, *) {
+                        // not exact after window merging
+                        $0.isTabbed = !cgsWindowIds.contains(cgWindowId)
+                    } else {
+                        // not known
+                        $0.isTabbed = false
+                    }
+                } else {
+                    $0.isTabbed = !visibleCgsWindowIds.contains(cgWindowId)
```
lwouis, line 177: Maybe here we could do this to avoid making the call if not needed?

```swift
if let visibleCgsWindowIds = visibleCgsWindowIds {
  $0.isTabbed = !visibleCgsWindowIds.contains(cgWindowId)
else {
  visibleCgsWindowIds = Spaces.windowsInSpaces(Spaces.idsAndIndexes.map { $0.0 }, false)
}
```

lwouis:
Hi @decodism,

I've taken a look at the latest code. The code is all heuristics and workarounds. There is no way for me to tell if it works just by reading it. I have to actually QA it.

An issue I have is that since moving to an M2 macbook and upgrading to Ventura, I can no longer run previous macOS releases in a VM. It's not supported. So I can QA on Ventura, but I can't make sure that the change doesn't bring things on macOS 12.

Do you have access to a macOS 12 machine? Could you test the changes there? I'm a bit worried to release a fix like this without test, because changes in that area can brick the app by mis-detecting tabs.

Thank you

lwouis:
I remember exploring this. I had the setup working but something was not working well. Maybe the performance was horrible? I forgot. I remember concluding that it's not viable for me to use to retro-test AltTab, then I removed all the UTM VMs from my machine.

If it's not too much of a bother, could you please test it on your side?

Thank you 🙇

lwouis:
I did some tests on macOS 13.4, and it seems to not have broken things. I'll ship this in the next release. Closing this MR.

Thank you very much for you work, as usual, @decodism!

PR #2021, opened 2022-10-15
fix: hide some thumbnail buttons for tabs

decodism:
Do not show the minimize and full screen thumbnail buttons on the tabs as they do not seem to work.

### DIFF
--- a/src/ui/main-window/ThumbnailView.swift
+++ b/src/ui/main-window/ThumbnailView.swift
@@ -78,7 +78,8 @@ class ThumbnailView: NSStackView {
             [quitIcon, closeIcon, minimizeIcon, maximizeIcon].forEach { icon in
                 icon.isHidden = !shouldShow ||
                     ((window_?.isWindowlessApp ?? true) && icon.type != .quit) ||
-                    (icon.type == .quit && window_?.application.runningApplication.bundleIdentifier == "com.apple.finder" && !Preferences.finderShowsQuitMenuItem)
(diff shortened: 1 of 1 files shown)


XInTheDark:
Maybe instead of removing them, we should try to get them to work?

For example, when the minimise action is applied to a tab, minimise the window containing the tab; and when the full screen button is pressed, make the window containing the tab full screen?

lwouis:
@XInTheDark I spent some time investigating, and it seems tricky to support this.

A tab will ignore these commands (min/de-min and fullscreen/de-fullscreen). What we can do is that we can send a combo of : focus the tab > send command to the tab. When adding a delay between them (I was succesful with `250ms`, but maybe a smaller delay is possible), it works. However, I ran into an issue where the tab detection broke after minimizing > de-minimizing > minimizing a tab that way.

I think it may be done with more effort. Now I'm not sure it's desirable UX in the first place. When a user minimizes a tab, I don't think they expect that it would focus it. It may mess with their ordering. Like minimizing an old window, and it becomes the most-recently-used suddenly.

In that regard, the approach from @decodism produces a clearer UX I think.

I'm not sure what's best overal. I lean toward disabling actions on tabs since the OS doesn't let users do those currently with the OS UX. So it would be consistent and less hacky. But I'm open to discussions.

XInTheDark:
Good idea! Simply removing the actions for the tab would be more intuitive for the user. In the system UI, the user is also not given the option to minimise/fullscreen a stand-alone tab, so AltTab should follow that.

decodism:
Closing in favor of #2020.

lwouis:
Thank you for this. I cherry-picked and tested it. I think it's great. I'm closing this PR, but the commit will be shipped in the next release 👍

PR #2047, opened 2022-10-26
fix: display app name in System Settings

### LINKED ISSUE #2044: Allow in Background relevant information
Currently AltTab displays the Dev name in Macos 13.

It should be changed to something relative for track-ability.

Screenshot 2022-10-26 at 1 17 55 PM

```
13:44:36 ❯ sw_vers
ProductName:            macOS
ProductVersion:         13.0
BuildVersion:           22A380
```

decodism:
Fixes #2044.

### DIFF
--- a/src/ui/preferences-window/tabs/GeneralTab.swift
+++ b/src/ui/preferences-window/tabs/GeneralTab.swift
@@ -54,6 +54,7 @@ class GeneralTab {
                 "Program": Bundle.main.executablePath ?? "/Applications/\(App.name).app/Contents/MacOS/\(App.name)",
                 "RunAtLoad": true,
                 "LimitLoadToSessionType": "Aqua",
+                "AssociatedBundleIdentifiers": App.id,
             ]
             plist.write(to: launchAgentsPath, atomically: true)
         } else {


lwouis:
Thanks for sharing this PR @decodism!

I checked the [official doc from Apple](https://developer.apple.com/documentation/servicemanagement/updating_helper_executables_from_earlier_versions_of_macos/) and it seems indeed that we need to add this new meta-data field.

I would like to confirm 2 things:

* Have you tested that it indeed shows the app name now? Does it also fix the icon? Could you share a screenshot?
* There is a yellow warning section in the docs saying the following:
  > Important
  > 
  > The Team Identifier of the Program or ProgramArguments executable in the legacy property list must match that of the app bundle for the AssociatedBundleIdentifiers key. Launch Services needs to know about the apps that the AssociatedBundleIdentifiers key targets. If necessary, register the app using [LSRegisterURL(_:_:)](https://developer.apple.com/documentation/coreservices/1446350-lsregisterurl).
  I'm not sure this PR honors this. But I'm not to understand what they mean by `The Team Identifier of the Program` either. What do you think about this?

decodism:
- I confirm that the name and icon are correctly displayed.

- I guess it's just to check that the displayed app is related to the target program. For `LSRegisterURL`, I don't know in which cases it is necessary.

lwouis:
I have cherry-picked the commit in the branch for the next release. Closing this. Thanks again for this quick fix!

PR #2059, opened 2022-10-29
fix: update thumbnail size

decodism:
When a window is created, its thumbnail is not always ready (non-existent or in appearance animation). The thumbnails were correctly updated afterwards but not their display which forced the previous size.

### DIFF
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -203,17 +203,10 @@ class Windows {
                     let window = list[currentIndex]
                     if window.shouldShowTheUser && !window.isWindowlessApp {
                         window.refreshThumbnail()
-                        DispatchQueue.main.async {
-                            let view = ThumbnailsView.recycledViews[currentIndex]
-                            if view.thumbnail.image != window.thumbnail {
-                                let oldSize = view.thumbnail.frame.size
-                                view.thumbnail.image = window.thumbnail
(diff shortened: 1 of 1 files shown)


lwouis:
Hi @decodism! Thank you for this PR!

Looking at the code, I can't understand the logic unfortunately. It would seem to me that in the case where there are thumbnails to refresh async, you don't recalculate their size anymore, which seems like it would make things worse, not better.

How did you test this PR? Do you have a series of steps to reproduce the issue it addresses?

decodism:
Thumbnail size was not recalculated but reused through `oldSize`. `refreshOpenUi` -> `updateItemsAndLayout` -> `layoutThumbnailViews` -> `updateRecycledCellWithNewContent` refreshes their display correctly.

Spamming Command-N in an app called "Find Any File" reproduces it for me.

lwouis:
Ah, I see! I looked again into the logic, and I think this change makes a lot of sense. Indeed the previous logic was assuming that the thumbnails wouldn't change size after it's updated asynchronously, but it's not a safe assumption.

And there is no event that triggers when these animations play. They are not done by resizing the window, but by scaling the final pixels through a GPU call I imagine. So no accessibility event to tell us that the window has changed size.

lwouis:
I took this commit in the dev branch. Closing this PR. It will be shipped next release 👍

PR #2186, opened 2022-12-05
feat: update russian localization

Armanio:


Armanio:
Oh my bad
I found info about poeditor
i'll close it

PR #2268, opened 2023-01-23
fix: reduce flag event tap latency

decodism:
Using a dedicated thread. Too high latency of an event tap can affect other apps.

### DIFF
--- a/src/logic/BackgroundWork.swift
+++ b/src/logic/BackgroundWork.swift
@@ -10,6 +10,7 @@ class BackgroundWork {
     static var mouseEventsThread: BackgroundThreadWithRunLoop!
     static var systemPermissionsThread: BackgroundThreadWithRunLoop!
     static var repeatingKeyThread: BackgroundThreadWithRunLoop!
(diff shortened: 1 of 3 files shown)


lwouis:
Hi @decodism,

Thank you for sharing this PR. Could you please share a bit more context on this?

I was surprised to see that the keyboard events are handled on the main thread RunLoop. It used to be on a background thread, exactly like with your diff here (see d3253ba4e1a2d599613ff06882f5781a7d475f06).

I'd like to clarify:

* What is your PR solving exactly? You mention that other apps may be impacted?
* Have you tested that moving this event processing to a background thread is working ok? Things off the main thread of very tricky, especially calling into macOS framework APIs. Few of them work correctly off the main thread. I would imagine that wrapping all, or parts of, `cgEventFlagsChangedHandler`, inside a dispatch to `DispatchQueue.main.async` may be safer, so avoid concurrency issues in the AltTab's code (e.g. touching `ControlsTab.shortcuts` while the main thread may be updating it), and to avoid crashes in AppKit APIs (e.g. is `CGEvent.tapEnable` thread-safe?)

PR #2271, opened 2023-01-25
feat: allow status item removal by dragging

decodism:
While holding command.


lwouis:
Thank you @decodism for this PR!

This is a nice refinement to the app. I've edited a bit the changes to avoid the issue when the preference dropdown is not yet created. Indeed, it's simpler to wait for that dropdown to exist, and toggle the removability and the observability from that point. Less code, all in the same file. I kept one commit with your author name for simplicity sake.

Closing this PR. I'll release this feature in the next release 👍

PR #2274, opened 2023-01-26
fix: exclude org.mozilla.plugincontainer

### LINKED ISSUE #2225: Alt-tab showing errors in system console launchd.log
The same error it's shown in the console log every milisecond.

I'm not sure what is it. 

I'm having a bit of high CPU usage, not sure if this is the cause.

Screenshot 2022-12-25 at 16 42 48

* AltTab version: 6.51.0
* macOS version: 13.1
* System: Macbook Pro M1 Pro (Apple Silicon)

siikamiika:
As a special case, make Applications.isActualApplication return false for Firefox plugincontainer so that it is not observed with the accessibility API. These processes do not support AX observer notifications although they report activationPolicy=accessory.

ref:
https://developer.apple.com/documentation/appkit/nsapplication/activationpolicy/accessory

github: fix #2225

### DIFF
--- a/src/logic/Applications.swift
+++ b/src/logic/Applications.swift
@@ -110,7 +110,7 @@ class Applications {
     private static func isActualApplication(_ app: NSRunningApplication) -> Bool {
         // an app can start with .activationPolicy == .prohibited, then transition to != .prohibited later
         // an app can be both activationPolicy == .accessory and XPC (e.g. com.apple.dock.etci)
-        return (isNotXpc(app) || isAndroidEmulator(app)) && !app.processIdentifier.isZombie()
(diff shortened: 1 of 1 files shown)


lwouis:
Hi @siikamiika,

I believe you shared the bug tracker were Mozilla fixed this issue on their side. This means we don't need a workaround in AltTab anymore, right?

siikamiika:
Yes, I tested a build that worked without issues on AltTab 6.52.1. It was on Firefox Nightly 111 so if it gets accepted without a delay it would be released on 2023-03-14 according to the [release calendar](https://wiki.mozilla.org/Release_Management/Calendar), but as far as I know it's not passed anything else but my review yet. I'm fine running my build of AltTab until the Firefox fix goes stable, but the release cycle is longer

lwouis:
Closing this PR as I believe it's absolete. Please let me know if it's still relevant. Thank you 🙇

Issue #590, opened 2020-09-09
Quick switch with fuzzy matching


max-sixty:
**Is your feature suggestion related to a problem? Please describe.**
I was directed here from [this discussion](https://news.ycombinator.com/item?id=24404996). I've been a heavy user and fan of Contexts. Unfortunately it's no longer in development (though works fine atm).

To quote that discussion:
> One of my favorite features is quick switching by holding down Caps/Esc and typing some of the Application / Window name. It'll be fuzzy matched and when releasing Caps/Esc, the selected window will be focused. So if you know the window you want to switch to, it's extremely fast.

And `mindfulhack` responded:
> The dev is currently extraordinarily dedicated to responding to tickets and is in heavy development.
I've already helped out and contributed a bit. Recommend to start Github issues for any gripes right away.

To the extent that's something you think would fit into the vision of this app, I find it invaluable.

lwouis:
Hi @max-sixty! Thanks for sharing this suggestion!

This is a great idea. I had a lot I wanted to discuss, so I made a [video response](https://www.loom.com/share/5163eabd79754297878f375b47f12ed1). Please check it out

max-sixty:
This is amazing! I'm watching now. I already really appreciate the response!

max-sixty:
This is really one of the most impressive interactions I've had in OSS, I really appreciate it.

A couple of small replies after watching the video:
- Thanks for showing those customizations, I copied them
- Your quick read on Contexts very much resonates. Both the parts you liked — the memory of searches — and the parts you found over-complicated — too many options, too many states
- Definitely no need for arrows keys within the search — generally it's <5 letters. Backspace is more than enough

Thanks again!

nfekete:
I love this idea a lot. Two things popped to my mind while watching the video.
1. If you were to implement it without a special trigger shortcut, just by having the user start typing in order to narrow the list of windows by fuzzy match, that would conflict with the current shortcuts used to close windows, minimize/unminimize window, quit app, hide/show app. These shortcuts are really useful, I use them quite often so I think they are really worth keeping. Maybe another (optional) modifier key could be used for either of these functions. You could create settings for having another modifier key for the fuzzy match filter and another additional modifier key for the window/app shortcuts. You could make any of these optional, so as no additional modifier key would be required for that particular function. Just need to make sure they are not conflicting with each other. You can't have both fuzzy match filter and window/app shortcuts with the same/no modifier key.
[...]

PR #2278, opened 2023-01-28
feat: Add search bar

ayroblu:
Initially I wanted to integrate alt-tab with Alfred as per #768, this covered a bunch of different things I tried, including extracting the images, and trying to figure out cross process communication in some way, such as with CFMessagePort etc.
But I realised that the primary benefit was for searching and a lot of the ordering and filtering tools options are already available in alt tab under the commands, and you can add lots of commands.

I realised that ultimately the best way was to integrate directly inside alt-tab, where everything is already in memory and low latency. I also saw that there's some buy in to do this with #590 so making this as a draft PR. This is "enough" for me, but there are some issues such as the changing window size, commented out instructions, manually changing the params to fit my needs, but maybe can get some direction or get some help with this.

### DIFF
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -95,7 +95,7 @@ class Windows {
         let focusedView = ThumbnailsView.recycledViews[focusedWindowIndex]
         focusedView.highlight(true)
         App.app.thumbnailsPanel.thumbnailsView.scrollView.contentView.scrollToVisible(focusedView.frame)
-        voiceOverFocusedWindow()
+        // voiceOverFocusedWindow()
     }
 
     static func voiceOverFocusedWindow() {
@@ -230,7 +230,8 @@ class Windows {
             } ?? false) &&
             !(Preferences.appsToShow[App.app.shortcutIndex] == .active && window.application.runningApplication.processIdentifier != NSWorkspace.shared.frontmostApplication?.processIdentifier) &&
             !(!(Preferences.showHiddenWindows[App.app.shortcutIndex] != .hide) && window.isHidden) &&
-            (searchText.isEmpty || (window.title?.contains(searchText) ?? true)) &&
+            (searchText.isEmpty || (window.title?.lowercased().contains(searchText) ?? true) ||
+             (window.application.runningApplication.localizedName?.lowercased().contains(searchText)) ?? true) &&
             ((!Preferences.hideWindowlessApps && window.isWindowlessApp) ||
                 !window.isWindowlessApp &&
                 !(!(Preferences.showFullscreenWindows[App.app.shortcutIndex] != .hide) && window.isFullscreen) &&
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -220,6 +220,7 @@ class Windows {
     }
 
     static func refreshIfWindowShouldBeShownToTheUser(_ window: Window, _ screen: NSScreen) {
+        let searchText = App.app.thumbnailsPanel.thumbnailsView.searchField.stringValue
         window.shouldShowTheUser =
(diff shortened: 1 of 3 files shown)


hexadecagram:
The need to select "Do nothing" under as the "then release" action could be obviated by making "Search" a configurable keypress to be used while open, with the default of course as `/`, which is a common default.

ayroblu:
> The need to select "Do nothing" under as the "then release" action could be obviated by making "Search" a configurable keypress to be used while open, with the default of course as `/`, which is a common default.

Yeah I wasn't sure, but I guess that making a special keypress that is for "search" would make sense. Cause there's no need for the search bar when pressing and holding. However some of the other shortcuts also don't make a lot of sense. It would be nice to have shortcuts per trigger type rather than having them be global

wiwo192:
Thanks, @ayroblu and @decodism and for working on adding the search feature. The feature would be super useful.

Want to check to see if we have any good news on this. Haven't seen any update since Feb.

ayroblu:
We need to settle on how the UI/UX should look. In the meantime I use this daily as is and it's "fine"

wiwo192:
> We need to settle on how the UI/UX should look. In the meantime I use this daily as is and it's "fine"

Is there a possibility to share the .app? Would love to use and try it. :).

Thanks again, on making it happen. :)

ayroblu:
[alttab-search.zip](https://github.com/lwouis/alt-tab-macos/files/11414094/alttab-search.zip)
Yeah sure, here you go, note that it's not up to date with the latest changes, and so provided as is

vurtomatic:
Excited for this.

mortang2410:
Thanks. I have been using the version with search for a day now. 
There is a slight usability problem in that when you are typing for fuzzy search, it can also trigger the other hotkeys for hiding / minimizing etc. My current workaround is disabling those other hotkeys.

ayroblu:
@mortang2410 please see the original post I have a screenshot of how I set it up. This is a critical part of setup.
As a reminder, we haven't solved the UI/UX issues yet so the point is that it's not supposed to be working well yet, so I'd like to focus discussion on how to transition the current layout to a more UX friendly layout.

mortang2410:
Has there been any news on this? This would be a great addition overall.

yakirlog:
Hi! any news about search bar addition? waiting for it long time 🤗
Thank you all for you hard work 🙏🏻

ayroblu:
No news, but I use this everyday. It's up to the maintainer, theoretically we could land the half baked version as is behind a switch, but it requires a lot of setup (turning off hotkeys etc) that might not be intuitive to the wider group. I already published usable versions above if you want to try it

damnever:
I think we could leverage [levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) to implement fuzzy search.

akira-toriyama:
+1

umang-govil:
@mortang2410 @ayroblu How have you been disabling the hotkeys in Mac? I liked the whole experience of using the alt-search version that @ayroblu posted above but struggled a bit with the fuzzy search as while searching for some text (like qw, qe, etc.), the search bar collapses.

ayroblu:
Please see original post

umang-govil:
@ayroblu Thanks for replying! I have the same setup as you have in your screenshot but I am still struggling with fuzzy search in the app that you shared [here](https://github.com/lwouis/alt-tab-macos/pull/2278#issuecomment-1537334112)

lwouis:
https://github.com/lwouis/alt-tab-macos/pull/4962 seems to go further, and is more active. Closing this PR to focus efforts on https://github.com/lwouis/alt-tab-macos/pull/4962 👍

Issue #210, opened 2020-04-03
Make it possible to select window by keyboard click


baest:
It would be very nice to be able to select the window to focus by using the other keys on the keyboard. So alt-tab and then the windows appears with a character in front of it (like the Desktop number). When you press that character (I'm thinking a-z, 0-9) then the window opens. If you have more windows than 35 I don't know if there should be multiple characters or just no character for the last ones.

lwouis:
Hi @baest! Thank you for sharing this idea!

There is a similar ticket with #178. Before anyone spends time implementing such a feature, I think it is important that there is a more developed specification for it. You mentioned for example the shortcut would be a-z, 0-9. Is this really a good mapping on the keyboard, when you left hand is already holding a modifier key, probably with your thumb? Look at RTS games like Starcraft 2. They use shortcuts close to the left hand like number, for control groups.

Also what happens for conflicting shortcuts, let's say if you set your main alt-tab shortcut to `alt-1` already?

I think all these questions and more should be answered now, so that we get a clearer picture of:

* Is the feature really a good idea? Use-cases, ergonomics, etc
* How much work is involved? Who's gonna take on the work, can we split it into installments, etc
* What are the interactions with existing features? How could this merge with #178 for example?
* Edge-cases

SoPat712:
This doesn't seem like a very logical issue, and mouse hover seems like it was the best solution. I think this should just be marked closed, to be honest.

lwouis:
I think @baest suggestion is interesting. Are you sure you understand the scenario? They basically would like hardcoded shortcuts like `alt+1` always focuses the first window, `alt+2` the second, etc. It's a very popular feature of other similar apps so I don't think we should discard it.

I think it could be done in a lightweight way by adding a checkbox `Number keys` here:

![image](https://user-images.githubusercontent.com/106195/92389200-9bff4980-f153-11ea-983d-62abc73d0dcb.png)

A deeper implementation would maybe allow more than 1 to 9, customize windows, etc. @baest What do you think?

SoPat712:
I think it would personally be a feature that I would never use, but I could see it being useful. I have too many windows open at a time for that to work well, however.

baest:
[...]

PR #2290, opened 2023-02-03
feat: add option to minimize other windows when switching

JonyEpsilon:
This PR adds a feature to minimise other windows, both as an intentional action from the switching window, and as a configurable default action when completing a switch.

I'm not sure you'll want to keep adding features to AltTab until it becomes all things to all people, so I won't be offended if you'd prefer not to add this functionality :-) There do look to be a few requests for something similar: https://github.com/lwouis/alt-tab-macos/issues/1530, https://github.com/lwouis/alt-tab-macos/issues/1902, https://github.com/lwouis/alt-tab-macos/issues/2086 . These FRs asked for the ability to *hide* other apps on switching, but that feels like quite a heavy action on MacOS - hiding all windows across all spaces - whereas minimising other windows can be implemented in a way that just applies to the set of windows that the current AltTab action is applying to, whether that be a single space or all spaces.

I've tried to change the existing logic as little as possible, as my impression is that it's probably been polished in the face of a thousand edge cases! I'm not sure the implementation is as clean as it could be, and we could happily streamline it if you're happy to risk changing the logic more (in particular, the thing that could be cleaned up, I think, is how the "action" to either focus the selected window, or minimise the other windows is triggered inside `ATShortcut` when the `triggerPhase` is `up`).

### DIFF
--- a/src/logic/Preferences.swift
+++ b/src/logic/Preferences.swift
@@ -28,6 +28,7 @@ class Preferences {
         "minDeminWindowShortcut": "M",
         "quitAppShortcut": "Q",
         "hideShowAppShortcut": "H",
+        "minimizeOtherWindowsShortcut": "1",
         "arrowKeysEnabled": "true",
         "mouseHoverEnabled": "true",
         "cursorFollowFocusEnabled": "false",
@@ -117,6 +118,7 @@ class Preferences {
     static var minDeminWindowShortcut: String { defaults.string("minDeminWindowShortcut") }
     static var quitAppShortcut: String { defaults.string("quitAppShortcut") }
     static var hideShowAppShortcut: String { defaults.string("hideShowAppShortcut") }
+    static var minimizeOtherWindowsShortcut: String { defaults.string("minimizeOtherWindowsShortcut") }
     static var arrowKeysEnabled: Bool { defaults.bool("arrowKeysEnabled") }
     static var mouseHoverEnabled: Bool { defaults.bool("mouseHoverEnabled") }
     static var cursorFollowFocusEnabled: Bool { defaults.bool("cursorFollowFocusEnabled") }
@@ -458,11 +460,13 @@ enum MenubarIconPreference: String, CaseIterable, MacroPreference {
 enum ShortcutStylePreference: String, CaseIterable, MacroPreference {
     case focusOnRelease = "0"
     case doNothingOnRelease = "1"
+    case minimizeOthersOnRelease = "2"
 
     var localizedString: LocalizedString {
         switch self {
             case .focusOnRelease: return NSLocalizedString("Focus selected window", comment: "")
             case .doNothingOnRelease: return NSLocalizedString("Do nothing", comment: "")
+            case .minimizeOthersOnRelease: return NSLocalizedString("Minimize other windows", comment: "")
         }
     }
 }
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -114,6 +114,14 @@ class Windows {
         return list.count > focusedWindowIndex ? list[focusedWindowIndex] : nil
     }
 
+    static func minimizeOtherWindows() {
+        Windows.list.enumerated().forEach {
+            if $0.offset != focusedWindowIndex && !$0.element.isMinimized && !$0.element.isWindowlessApp && $0.element.shouldShowTheUser {
+                $0.element.minDemin()
+            }
+        }
+    }
+
     static func cycleFocusedWindowIndex(_ step: Int) {
         let nextIndex = windowIndexAfterCycling(step)
         if ((step > 0 && nextIndex < focusedWindowIndex) || (step < 0 && nextIndex > focusedWindowIndex)) &&
--- a/src/ui/preferences-window/tabs/ControlsTab.swift
+++ b/src/ui/preferences-window/tabs/ControlsTab.swift
@@ -5,11 +5,11 @@ class ControlsTab {
     static var shortcuts = [String: ATShortcut]()
     static var shortcutControls = [String: (CustomRecorderControl, String)]()
     static var shortcutsActions = [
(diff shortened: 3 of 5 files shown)


src/ui/preferences-window/tabs/ControlsTab.swift
```diff
@@ -5,11 +5,11 @@ class ControlsTab {
     static var shortcuts = [String: ATShortcut]()
     static var shortcutControls = [String: (CustomRecorderControl, String)]()
     static var shortcutsActions = [
-        "holdShortcut": { App.app.focusTarget() },
-        "holdShortcut2": { App.app.focusTarget() },
-        "holdShortcut3": { App.app.focusTarget() },
-        "holdShortcut4": { App.app.focusTarget() },
-        "holdShortcut5": { App.app.focusTarget() },
+        "holdShortcut": { App.app.handleHold(0) },
```
JonyEpsilon, line 8: I'm not sure about the name of this function, but I can't think of a better one :-)

src/logic/Windows.swift
```diff
@@ -114,6 +114,14 @@ class Windows {
         return list.count > focusedWindowIndex ? list[focusedWindowIndex] : nil
     }
 
+    static func minimizeOtherWindows() {
+        Windows.list.enumerated().forEach {
+            if $0.offset != focusedWindowIndex && !$0.element.isMinimized && !$0.element.isWindowlessApp && $0.element.shouldShowTheUser {
+                $0.element.minDemin()
```
lwouis, line 120: You check `shouldShowTheUser` here. Is it what the user would expect though? Let's imagine I use the ``` command+` ``` shortcut. It shows me only windows of the active app. When I set it to "minimize other windows" on release, I think I expect that every other window on my screen will minimize, except the one I just focused, no? Here it will only minimize the other windows of the active app.

Same logic will keep some windows on-screen, depending on the shortcut filtereing logic behind `shouldShowTheUser`.

Another point: I wonder if people will expect this single-window mode to be per-Space and/or per-Screen. I'm not sure users expect windows in every Space and on every Screen to minimize when they go for this type of workflow. Maybe they would like to have a "main window" per Screen?

What do you think?

src/logic/Preferences.swift
```diff
@@ -28,6 +28,7 @@ class Preferences {
         "minDeminWindowShortcut": "M",
         "quitAppShortcut": "Q",
         "hideShowAppShortcut": "H",
+        "minimizeOtherWindowsShortcut": "1",
```
lwouis, line 31: I think adding the option to minimize other windows is not adding much noise in the existing dropdown "on release:". However, as a secondary shortcut like quit/hide/etc, I think it's too niche to have such a prominent UI in the preferences. This shortcut `1` would also conflict with #210. I think the feature is a bit too niche to have a dedicated shortcut as such.

lwouis:
Hi @JonyEpsilon 👋 

> This PR adds a feature to minimise other windows, both as an intentional action from the switching window, and as a configurable default action when completing a switch.
> 
> I'm not sure you'll want to keep adding features to AltTab until it becomes all things to all people, so I won't be offended if you'd prefer not to add this functionality :-) There do look to be a few requests for something similar: https://github.com/lwouis/alt-tab-macos/issues/1530, https://github.com/lwouis/alt-tab-macos/issues/1902, https://github.com/lwouis/alt-tab-macos/issues/2086 .

I think it's ok to add yet another preference, if it's in the existing dropdown "On release:". I think it's a dropdown very few people touch in the first place, so it's ok to cram another power-user option there. Regarding adding a new secondary shortcut, I think it's too niche to warrant the added complexity though.

> These FRs asked for the ability to hide other apps on switching, but that feels like quite a heavy action on MacOS - hiding all windows across all spaces - whereas minimising other windows can be implemented in a way that just applies to the set of windows that the current AltTab action is applying to, whether that be a single space or all spaces.

I made a comment in the PR to discuss the topic of which windows should be affected by the trigger. 

I will add, regarding other people wanting to hide windows instead of minimizing, that I think if we add this feature, we could as well add yet another dropdown option to hide instead of minimizing.

> I've tried to change the existing logic as little as possible, as my impression is that it's probably been polished in the face of a thousand edge cases! I'm not sure the implementation is as clean as it could be, and we could happily streamline it if you're happy to risk changing the logic more (in particular, the thing that could be cleaned up, I think, is how the "action" to either focus the selected window, or minimise the other windows is triggered inside ATShortcut when the triggerPhase is up).

You're very kind. I think the codebase is a bit messy, as a result of the million complexities and tricks we do to get macOS to comply to basic functionality (e.g. there is no public API to manipulate Spaces, so we use private APIs/SPIs and side-effects to detect things).

The code in this PR is of very good quality, in my eyes. I think if you're ok to make a few changes from my suggestions, I can merge this PR and release soon after 👍

JonyEpsilon:
Hey, just a very quick comment to say that the above all sounds reasonable, and I'll update the PR when I have a chance - been a bit busy lately.

I might need some advice on how to select the proper set of windows to minimise, but I should think about it myself first!

shredthaGNAR:
easier to just use BTT aka BetterTouchTool to add. action to activate via trackpad when hovered over AltTab App Window ex to. hide app window or closing etc and with DockAltTab. and BTT you do the same for dock previews when you hover over apps in doc,. I set it. up. on my system works. great.....still which you would add the few trivial applescript commands SteventheWorker creator of DockAltTab requested because it would allow for apps. to be kept up to date aka not have to use a specially made AltTab instance that only. difference is those. few. lines. of innocuous appleescipt functionality. Having to constantly recompiling every time you update make its confusing and harder users as opposed to having DockAltTabjust be a companioon-like app that track. your up to date AltTab

lwouis:
I think this feature goes too much into windows-management. This is the line I want to draw with AltTab. That it doesn't venture into windows-management. Better solution exists for this, like [yabai](https://github.com/asmvik/yabai) 👍

Issue #2209, opened 2022-12-15
Per shortcut "press X to select previous window"


olehmisar:
**Is your feature suggestion related to a problem? Please describe.**
I have two shortcuts in place: option+tab and option+backtick. I want to use option+shift+tab to select previous window when I activated the first shortcut and to use option+shift+backtick to select previous window when I activated the second shortcut.

**Describe the solution you'd like**
Right now, the setting "press X to select previous window" is linked to all shortcuts. I would appreciate it if you make this setting per-shortcut.

(this is how it looks now)
<img width="574" alt="image" src="https://user-images.githubusercontent.com/29802592/207795680-6e3ab061-f64b-40fc-909c-5674f9181a7c.png">
<img width="574" alt="image" src="https://user-images.githubusercontent.com/29802592/207795712-ea3b84f3-cd47-41bb-bd67-e505431737cc.png">

**Describe alternatives you've considered**
\-

lwouis:
This would complicate the app for most users, for the benefice of very few users. In other words, i'm afraid it is too niche, sorry.

olehmisar:
no problem. I can totally live with that because I use only 1 shortcut most of the time.

chase-miller:
fwiw I'm in favor of this. I'm coming from the Witch app, which follows the "shift + x" pattern to select a previous window.

chase-miller:
I've implemented this feature via https://github.com/lwouis/alt-tab-macos/pull/2311

PR #2311, opened 2023-02-10
feat: per-section previousWindowShortcut bindings

chase-miller:
Implements feature request https://github.com/lwouis/alt-tab-macos/issues/2209. 

This allows for alt-tab to handle cycling backwards in the same way many window managers operate: by including `shift` in the key combo. Without this, only one shortcut can operate under this model.

It defaults to the same behavior that's existed (⇧ cycles backwards), and migration will bind all shortcut keys to whatever is existing, ensuring functional equivalency until a user explicitly changes it.

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,6 @@
 **AltTab** brings the power of Windows alt-tab to macOS
 
 [Find out more on the official website](https://alt-tab-macos.netlify.app/)
+
+## Contributing
+[See contributing on the official website](https://alt-tab-macos.netlify.app/contributing)
--- a/src/logic/KeyRepeatTimer.swift
+++ b/src/logic/KeyRepeatTimer.swift
@@ -5,12 +5,12 @@ class KeyRepeatTimer {
     static var timer: Timer?
     static var isARepeat = false
 
-    static func toggleRepeatingKeyPreviousWindow() {
-        if let shortcut = ControlsTab.shortcuts["previousWindowShortcut"],
+    static func toggleRepeatingKeyPreviousWindow(_ shortcutIndex: Int) {
+        if let shortcut = ControlsTab.shortcuts[Preferences.indexToName("previousWindowShortcut", shortcutIndex)],
            // events already repeat when using a shortcut with a keycode; no need for artificial repeat
            shortcut.shortcut.keyCode == .none {
             toggleRepeatingKey(shortcut) {
-                App.app.previousWindowShortcutWithRepeatingKey()
+                ControlsTab.shortcutsActions[Preferences.indexToName("previousWindowShortcut", shortcutIndex)]!()
             }
         }
     }
--- a/src/logic/Preferences.swift
+++ b/src/logic/Preferences.swift
@@ -23,6 +23,10 @@ class Preferences {
         "nextWindowShortcut5": "",
         "focusWindowShortcut": "Space",
         "previousWindowShortcut": "⇧",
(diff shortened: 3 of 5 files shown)


lwouis:
Hi @chase-miller,

Thank you for sharing your work in this PR.

I'm afraid that, as indicated in the [ticket's discussion](https://github.com/lwouis/alt-tab-macos/issues/2209#issuecomment-1352679561), I don't think it's a positive for most users to move this action in a per-shortcut mode.

I'm afraid I won't be able to merge this PR 🙇

chase-miller:
@lwouis is there some way we can test our hypotheses? Maybe add a checkbox next to the global previous binding field that enables per-shortcut bindings? We could then passively track usage of the new fields (maybe a feedback button), or we could simply see if there's a reduction in previous window-related issues on GitHub. Or something else :-)

I hypothesize that it will actually be **less** confusing to pair previous bindings alongside forward bindings. It will also facilitate the idiomatic `shift + shortcut` way of navigating backwards for users that leverage more than one shortcut.

Here’s an example of Witch’s preference pane as an example of a similar app that does it this way:
Pasted Graphic


What do you think about somehow testing this out to see how users react?


P.S. In case others find it useful I’ll aim to maintain a fork at https://github.com/chase-miller/alt-tab-macos that adds this feature.

lwouis:
We don't do telemetry and never will. This leaves us with only 1 source of user feedback: github tickets. Either from people visiting github directly, or people using the in-app "Send feedback..." form, which automatically creates a new github ticket.

I've been working on this app for 3 years. I've been responding, classifying, discussing tickets alone during that time. I've never found someone who would be interested to co-pilot this project (neither has anyone shown serious interest for #1179). Through that time, I've observed the type of users this software has. It's a biased sample of the total userbase, which I estimate at around 100K people. But it's the only data I have, beyond my own desires/vision for the product.

Overall, I think a handful of people, at most, have asked for this. There are more people, for instance, who are asking to have the backward-cycling shortcut be triggerable when AltTab is not open (i.e. "I want to summon AltTab by pressing option+shift+tab"). They expect it from Windows. Both are quite niche, as defined by the low interest shown for it on github.

My best judgment is that it's too niche to move forward, and that it degrades UX. Of course both are subjective. But given that we have little (and sample-biased) data, and that I've been maintaining this and listening/discussing with users through 2000 tickets, it seems to me that, until someone takes over, my opinion might be considered the most informed, and my decision to not move forward with this PR reasonable. I understand though, that from where you stand, this PR looks like a clear win. I'm sorry about that 🙇

chase-miller:
> There are more people, for instance, who are asking to have the backward-cycling shortcut be triggerable when AltTab is not open

Ultimately this is what I wanted to do, but it appeared to involve considerably more effort.

> Both are quite niche, as defined by the low interest shown for it on github.

I wish there was some way to better quantify this. The previous tab behavior really threw me off when I first used the app, and I wonder if others feel the same.

> my opinion might be considered the most informed, and my decision to not move forward with this PR reasonable

can't argue with you there ;-)

> I understand though, that from where you stand, this PR looks like a clear win. I'm sorry about that 🙇

It's all good, and this is one of the ways in which open source is great. Also fwiw you do an excellent job handling user feedback and PRs, particularly with the way in which you deliver bad news and show empathy. Be proud :-)

Ultimately I love the app and thank you for making it something we're even talking about. Thanks!

PR #2316, opened 2023-02-11
fix: mouse hover

decodism:
Revert to the old behavior of `Preferences.mouseHoverEnabled` and add `Preferences.hideMouseHoverEffect`.


syrte:
Related: https://github.com/lwouis/alt-tab-macos/issues/2317

lwouis:
I think it's correct to rework the existing `Also select windows using: Mouse hover` preference. Some people have complained they miss the old behavior, so we could keep that preference, but default it to `false`.

Now the new `Hide effect on mouse hover` I think is too niche. It's just a visual indicator. Most people can accept it's there. It doesn't hurt their workflow. It's also there on Windows, which inspired to add it in AltTab.

PR #2332, opened 2023-02-14
feat: allow shortcut switching

decodism:
Without having to close the panel.

### DIFF
--- a/src/logic/ATShortcut.swift
+++ b/src/logic/ATShortcut.swift
@@ -60,7 +60,7 @@ class ATShortcut {
 
     func shouldTrigger() -> Bool {
         if scope == .global {
-            if triggerPhase == .down && (!App.app.appIsBeingUsed || index == App.app.shortcutIndex) {
+            if triggerPhase == .down && (!App.app.appIsBeingUsed || index == nil || index == App.app.shortcutIndex) {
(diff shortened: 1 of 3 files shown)


lwouis:
Hi @decodism,

I'm not sure to understand this change. For example, if we consider default shortcuts, the user would do `option+tab` to bring up AltTab UI. Then they would press ``` ` ```, and it would update the UI to only show the thumbnails of windows of the active app (i.e. Shortcut 2). Changing the scope of AltTab, on the fly, as the UI is already opened.

Am I understanding the goal of this change correctly?

Thank you

lwouis:
Hi @decodism,

Sorry for the delay to get back to you on this PR.

I've tested the PR and it seems to work correctly.

When I look at the code, it seems to be reaching the goal with very few changes, and in a smart way. However, I'm very afraid to do changes in these particular areas of the codebase.

Are you confident that the code is correct? I'm sorry to ask such a subjective question, but I don't have much time to give this project anymore, so I'm more worried about breaking the existing app that adding small refinements. Your confidence level would help me decide the risk level here.

Thank you 🙇

lwouis:
Thank you @decodism. I'm shipping this in the next release 👍

Issue #730, opened 2020-12-14
Support 3-finger scroll to navigate the thumbnails


ndaskalovic:
**Is your feature suggestion related to a problem? Please describe.**
would be much faster to add trackpad gestures, and/or support trackpad scrolling to cycle through windows. similar to windows 10 gestures.

**Describe the solution you'd like**
three finger swipe on trackpad to scroll through windows backwards and forwards.
**Describe alternatives you've considered**
ive tried using BTT and native app switcher which does work when assigning three finger swipe to activate app switcher. but of course alt-tab has much better features that I require such as showing all open windows instead of just apps as well as app previews. would be nice to have the best of both worlds.

lwouis:
I'd like to clarify the feature you touch on.

I played with the built-in app switcher. It seems to respond to scrolling. That is, with the mouse wheel, or with the trackpad:

* Scrolling the mouse wheel up and down scroll the app icons left and right
* Swiping with 2 fingers left and right scroll the app icons left and right
* Swiping with 2 fingers up and down scroll the app icons left and right

I think it can be summed up with: scrolling with any device moves selection left or right. Scrolling horizontally or vertically are remapped to the same direction (i.e. up is right, bottom is left).

Currently in AltTab, you can already scroll. Mouse of trackpad. The difference is that scrolling doesn't move the selection around. Instead it move the `viewport` (i.e. visible portion of the windows list) around. It doesn't show if you have only a few windows listed, but try to open like 30 windows, and you will see how scrolling moves the viewport.

I think this should be a preference, as the current behavior may be preferred by some people. If you have frequently 100+ open windows, you may not want you scroll down to scan all rows from left to right until you reach the bottom. You may want what is today current behavior instead.

The new preference could be:

```
Scrolling: * Scrolls the view up or down
           * Select the next or previous window
```

Note: what the OS qualifies as a scroll event can be customized through System Preferences > Trackpad, or with tools such as BTT, but AltTab shouldn't bind to specifics, and instead obey scroll events, however they get generated.

What do you think?

ndaskalovic:
Hi,

[...]

Issue #756, opened 2020-12-31
Support 3-finger swipe to invoke AltTab


klarasch:
First of all, thank you so much for developing this tool, just switched from Windows and couldn't understand how folks on MacOS can live without switching windows, not apps and how Mission Control doesn't have window selection through keyboard and ability to close windows from there. So frustrating to go to each windows individually to close it – and so painfully slow. Your tool takes it even further (closing apps by shortcut, I'm in love!) I might've switched back, hadn't found this. Thank. You. ❤️

One thing that keeps it from complete perfection (and not sure it's possible, but hopefully given BetterTouchTool is tiny close to this) is Windows' **ability to trigger the alt-tab menu by swiping with three fingers left or right.** I would like to have default MacOS four-finger swipe to switch desktops, but three-finger (horizontal, would like to keep Mission Control and Exposé on vertical) swipe to trigger alt-tab menu and then select the window by hovering over it with the three fingers, i.e. the same logic Windows use. Currently by default, both three- and four-finger swipe does the same on MacOS – switches desktops; such a wasted gesture distinction. 

Thank you for considering this suggestion!

(sorry if this was already suggested, checked the other issues though!)

PS: if this makes AppStore, I'd pay and I don't like paying for apps; this is far more life-improving than most of paid stuff there

lwouis:
Hey @klarasch! Glad you enjoy the app!

Your feature request reminds me of the discussion in https://github.com/lwouis/alt-tab-macos/issues/730. I think I would probably want to implement those together. I need to get my hands on a Windows laptop to experience that gesture UX though. For now I've no idea about its details, and god knows details matter with these kinds of interactions.

> if this makes AppStore

It will never, unfortunately. The app uses a number of private APIs to achieve its functionality. There are no public APIs to manipulate Spaces, and that made my work very hard at the beginning. Now I've found enough private APIs, and they seem stable over macOS releases, so the app works. But because of that it can never be distributed on the MAS.

Thanks for the feature request. I think it's a great proposal. I need to check out how Windows 10 does it, then see what it would mean for AltTab in terms of maybe new preferences, and implementation details that make sense on the mac platform.

klarasch:
Hey, thanks for the response and yeah, apologies! I overlooked that issue. Yes, it's described well there. I understand it may be pretty complicated to implement, though, and it's just a suggestion, the app takes the great experience further than Windows in many aspects already. :)

[...]

PR #2355, opened 2023-02-21
Swipe gestures support PoC

lwouis:
[Context](https://github.com/lwouis/alt-tab-macos/issues/730#issuecomment-1438979530)

### DIFF
--- a/src/logic/events/TrackpadEvents.swift
+++ b/src/logic/events/TrackpadEvents.swift
@@ -1,0 +1,113 @@
+import M5MultitouchSupport
+
+//TODO: patch M5MultitouchSupport pod to make it not to crash after sleep. See https://github.com/mhuusko5/M5MultitouchSupport/issues/1
+//TODO: App Expose activates if swipe down with 3-fingers right after AltTab activation. There is no issue if wait for some time.
+//TODO: enable/disable trackpad usage in settings
+class TrackpadEvents {
+    //TODO: Should we add a sensetivity setting instead of these magic numbers?
+    private static let accVelXThreshold: Float = 10
+    private static let accVelYThreshold: Float = 15
+    
+    private static var listener: M5MultitouchListener?
+
+    static func addSwipeListener() {
+        var accVelX: Float = 0
+        var accVelY: Float = 0
+        // We need to track an UI activation via gesture to not to hide UI on trackpad events for other activations.
+        var activated: Bool = false
+
+        listener = M5MultitouchManager.shared().addListener {event in
+            if event == nil {
+                return
+            }
+
+            let touches = event!.touches as! [M5MultitouchTouch]
+
+            // We don't care about non-3-fingers swipes.
+            if touches.count != 3 {
+                // Except when we already started a gesture, so we need to end it.
+                if activated {
+                    if App.app.appIsBeingUsed {
+                        DispatchQueue.main.async {
+                            App.app.focusTarget()
+                            App.app.hideUi()
+                        }
+                    }
+                    accVelX = 0
+                    accVelY = 0
+                    activated = false
+                }
+                return
+            }
+
+            let velocity = swipeVelocity(touches: touches)
+            // We don't care about gestures other than horizontal or vertical swipes.
+            if velocity == nil {
+                return
+            }
+
+            accVelX += velocity!.x
+            accVelY += velocity!.y
+            // Not enough swiping.
+            if abs(accVelX) < accVelXThreshold && abs(accVelY) < accVelYThreshold {
+                return
+            }
+
+            let isHorizontal = abs(velocity!.x) > abs(velocity!.y)
+            if App.app.appIsBeingUsed {
+                let direction: Direction = isHorizontal
+                    ? accVelX < 0 ? .left : .right
+                    : accVelY < 0 ? .down : .up
+                DispatchQueue.main.async { App.app.cycleSelection(direction) }
+            } else {
+                if isHorizontal {
+                    DispatchQueue.main.async { App.app.showUi() }
+                    activated = true
+                }
+            }
+            accVelX = 0
+            accVelY = 0
+        }
+    }
+
+    private static func swipeVelocity(touches: [M5MultitouchTouch]) -> SwipeVelocity? {
+        var allRight = true
+        var allLeft = true
+        var allUp = true
+        var allDown = true
+        var sumVelX = Float(0)
+        var sumVelY = Float(0)
+        for touch in touches {
+            allRight = allRight && touch.velX >= 0
+            allLeft = allLeft && touch.velX <= 0
+            allUp = allUp && touch.velY >= 0
+            allDown = allDown && touch.velY <= 0
+            sumVelX += touch.velX
+            sumVelY += touch.velY
+        }
+        // All fingers should move in the same direction.
+        if !allRight && !allLeft && !allUp && !allDown {
+            return nil
+        }
+
+        let velX = sumVelX / Float(touches.count)
+        let velY = sumVelY / Float(touches.count)
+        return SwipeVelocity(x: velX, y: velY)
+    }
+
+    static func removeSwipeListener() {
+        if listener != nil {
+            M5MultitouchManager.shared().remove(listener)
+        }
+    }
+    
+    private class SwipeVelocity {
+        let x: Float
+        let y: Float
+        
+        init(x: Float, y: Float) {
+            self.x = x
+            self.y = y
+        }
+    }
+}
--- a/Podfile
+++ b/Podfile
@@ -7,4 +7,5 @@ target 'alt-tab-macos' do
   pod 'ShortcutRecorder', :git => 'https://github.com/lwouis/ShortcutRecorder.git', :branch => 'alt-tab-current'
   pod 'SwiftyMarkdown', '1.1.0'
   pod 'AppCenter/Crashes', '4.3.0'
+  pod 'M5MultitouchSupport', '1.1'
 end
--- a/src/ui/App.swift
+++ b/src/ui/App.swift
@@ -61,6 +61,8 @@ class App: AppCenterApplication, NSApplicationDelegate {
             self.feedbackWindow = FeedbackWindow()
             KeyboardEvents.addEventHandlers()
             MouseEvents.observe()
+            //TODO: we need to remove swipe listener somewhere
+            TrackpadEvents.addSwipeListener()
             // TODO: undeterministic; events in the queue may still be processing; good enough for now
             DispatchQueue.main.async { () -> () in Windows.sortByLevel() }
             self.preloadWindows()


src/logic/events/TrackpadEvents.swift
```diff
@@ -0,0 +1,113 @@
+import M5MultitouchSupport
+
+//TODO: patch M5MultitouchSupport pod to make it not to crash after sleep. See https://github.com/mhuusko5/M5MultitouchSupport/issues/1
+//TODO: App Expose activates if swipe down with 3-fingers right after AltTab activation. There is no issue if wait for some time.
+//TODO: enable/disable trackpad usage in settings
+class TrackpadEvents {
+    //TODO: Should we add a sensetivity setting instead of these magic numbers?
```
lwouis, line 7: We can either hardcode such numbers if they are stable in macOS over time, or we may be able to read the system settings, like we read the key repeat rate today, or other System Settings settings.

src/logic/events/TrackpadEvents.swift
```diff
@@ -0,0 +1,113 @@
+import M5MultitouchSupport
+
+//TODO: patch M5MultitouchSupport pod to make it not to crash after sleep. See https://github.com/mhuusko5/M5MultitouchSupport/issues/1
```
lwouis, line 3: I'm a bit worried to rely on a third-party library that hasn't been updated in 8 years, and that has a crash on wake that has only been found this year.

Is there no way to roll our own code for multi-fingers gesture? Is it too rough when using the AppKit APIs?

src/logic/events/TrackpadEvents.swift
```diff
@@ -0,0 +1,113 @@
...
+            let isHorizontal = abs(velocity!.x) > abs(velocity!.y)
+            if App.app.appIsBeingUsed {
+                let direction: Direction = isHorizontal
+                    ? accVelX < 0 ? .left : .right
+                    : accVelY < 0 ? .down : .up
+                DispatchQueue.main.async { App.app.cycleSelection(direction) }
+            } else {
+                if isHorizontal {
+                    DispatchQueue.main.async { App.app.showUi() }
+                    activated = true
```
lwouis, line 65: Do we need to track this bool actually? Maybe `App.app.appIsBeingUsed` is enough?

Shouldn't the user be able to open AltTab with alt+tab, but then 3-fingers swipe left or right?

src/logic/events/TrackpadEvents.swift
lwouis: **Update:** i just searched online for this, and there is no information out there. I understand better why you went with this library. Classic Apple move to not properly support such a classic interaction decades after it's been added to macOS.

I've updated the issue on the `M5MultitouchSupport` repo. Maybe we could add something like this to fix the issue.

lwouis:
Here are comments on your message in the original ticket.

> User has to disable 3-finger swipe between full-screen apps or make it 4-finger in System Settings > Trackpad > More Gestures > Swipe between full-screen apps.

Today when the user sets their keyboard shortcut to `command+tab`, we use a private API to disable the native shortcut, so that we can replace it. There may be a way to disable the native 3-finger swipe, if the user sets AltTab to use 3-finger swipes in the preferences, in the same fashion. Maybe.

> No settings. It's not clear to me where they should be added.

[I suggested how to implement the preference here](https://github.com/lwouis/alt-tab-macos/issues/756#issuecomment-754549805).

> App crashes after sleep. See https://github.com/mhuusko5/M5MultitouchSupport/issues/1

I added a comment there. Maybe we could dig into that workaround?

> Only 3-finger swipes are supported.

Maybe it's ok with 3 fingers for now. Maybe no-one needs 4 fingers. And 2 fingers is definitely not a good idea for a global trigger.

> App Expose activates if swipe down with 3-fingers right after AltTab activation. There is no issue if wait for some time.

Again, we could maybe disable that with private API calls. Alternatively, we could maybe absorb some of the gesture events like we absorb keyboard events (so that the apps under AltTab don't get a `keyUp` event when we release `option` to close AltTab).

Furthermore, I want to add that in this PR, you have implemented left/right swipes to move around the thumbnails. I think ideally we would want to implement **360 movement** to move around the grid of thumbnails if the user has many.

src/logic/events/TrackpadEvents.swift
ris58h: I don't want to add a 3-rd party library either but it's just a PoC.
We could patch M5MultitouchSupport as I did it for Touch-Tab https://github.com/ris58h/Touch-Tab/blob/master/patches/M5MultitouchSupport%2B1.1.0.diff (it works fine) or use any other library.
Seems like https://github.com/Kyome22/OpenMultitouchSupport is a good candidate.

P.S.:
Code in both libraries is nearly identical ([M5MultitouchSupport](https://github.com/mhuusko5/M5MultitouchSupport/blob/master/M5MultitouchSupport/M5MultitouchManager.m#L214) vs [OpenMultitouchSupport](https://github.com/Kyome22/OpenMultitouchSupport/blob/master/OpenMultitouchSupport/OpenMTManager.m#L121)) but OpenMultitouchSupport checks if [MTDeviceIsRunning](https://github.com/Kyome22/OpenMultitouchSupport/blob/master/OpenMultitouchSupport/OpenMTManager.m#L122) before releasing the device.

src/logic/events/TrackpadEvents.swift
ris58h: We need to track that the UI was triggered by swiping because end of the swipe (number of fingers not equal to 3 in our case) hides the UI. If a user opens the UI from menu or by shortcut, every mouse move except 3-finger swipe will close the UI.

ris58h:
> Today when the user sets their keyboard shortcut to command+tab, we use a private API to disable the native shortcut, so that we can replace it. There may be a way to disable the native 3-finger swipe, if the user sets AltTab to use 3-finger swipes in the preferences, in the same fashion. Maybe.

I couldn't find the way to do it using private API used in M5 library because it's just a callback. Maybe we could use `CGEventTap` as it's done here https://github.com/Kolsha/Force-Command-Click/blob/master/main.m

> I suggested how to implement the preference here

Just to make it clear: do you want to add `Gesture` tab to `Shortcut` tabs?

> I added a comment there. Maybe we could dig into that workaround?

I've answered there.

> Maybe it's ok with 3 fingers for now. Maybe no-one needs 4 fingers. And 2 fingers is definitely not a good idea for a global trigger.

Yeah, 2-fingers is not an option here.

> Again, we could maybe disable that with private API calls. Alternatively, we could maybe absorb some of the gesture events like we absorb keyboard events (so that the apps under AltTab don't get a keyUp event when we release option to close AltTab).

It's strange to me that App Expose doesn't activate if I wait for a couple of seconds. I've answered in the code comments about disabling native behavior.

> Furthermore, I want to add that in this PR, you have implemented left/right swipes to move around the thumbnails. I think ideally we would want to implement 360 movement to move around the grid of thumbnails if the user has many.

What is 360 movement? I've already implemented thumbnails navigation in all directions (left/right/up/down).

src/logic/events/TrackpadEvents.swift
lwouis: Maybe the following logic could work?

We track that we reached 3 fingers either:

* Before AltTab was opened with the trackpad gesture
* After it was opened with the keyboard shortcut

From that point on, any < 3 fingers gesture means we should focus the window and close UI.

lwouis:
> Just to make it clear: do you want to add `Gesture` tab to `Shortcut` tabs?

Yes. That way people can decide which windows they will see in that mode. Originally I was thinking people could bind different swipes (e.g. 3 fingers vs 4 fingers) to different windows to show. A bit like how `option+tab` shows all windows, and ``` option+` ``` shows only the active app windows, with keyboard shortcuts.

> What is 360 movement? I've already implemented thumbnails navigation in all directions (left/right/up/down).

The way I would describe it is: imagine that moving your 3-fingers gesture on the trackpad moves a hidden mouse cursor. When you move in 2D space, in 360 degrees of freedom, you're now moving around precisely. Now we make that hidden mouse cursor select thumbnails.

It's very different from only being able to go in 1 direction at a time. You can move in diagonal.

To imagine the difference, imagine that you are inside a folder, and in grid view. Now imagine wanting to go to up-and-left to a particular file. Either you swipe directly towards the file (i.e. 360 movement), and the selection roughly follows your motion, or you would have to swipe horizontally, then vertically, a distance that potentially quite long, and see your selection move in a L pattern.

If you use a Windows machine, and check out the built-in alt+tab UX with a trackpad, you can experience how great the 360 experience is.

On that note, currently we can scroll up and down with the mouse wheel or 2-fingers trackpad, when the grid has a lot to display. That code should probably be updated to accomodate the new 3-fingers interactions.

ris58h:
> It's very different from only being able to go in 1 direction at a time. You can move in diagonal.

Oh, I got it.

> On that note, currently we can scroll up and down with the mouse wheel or 2-fingers trackpad, when the grid has a lot to display. That code should probably be updated to accomodate the new 3-fingers interactions.

I've tried to allow both scrolling and moving with trackpad with 2 and 3 finger swipes accordingly but it was clunky. Unfortunately I don't have a Windows laptop right now to check. I believe we should mimic Windows behavior if it's possible.

src/logic/events/TrackpadEvents.swift
ris58h: We could track what caused the UI to be shown by adding a parameter to `showUI` method and storing it.

lwouis:
> I've tried to allow both scrolling and moving with trackpad with 2 and 3 finger swipes accordingly but it was clunky. Unfortunately I don't have a Windows laptop right now to check. I believe we should mimic Windows behavior if it's possible.

Windows behavior is actually pretty bad in that regard. I generally find their UX very good, but for some reason their scrolling is horrible. It's row-by-row instead of pixel-by-pixel, so it can be confusing to the user. They also don't do any form of "look ahead" like they would do in a platformer video game where we want to see a bit of what's ahead, not be at the bleeding edge.

Finally, if you own Parallels, or maybe even with the free VirtualBox or BootCamp, you can install a free Windows 10 VM these days, from Microsoft website. It's an easy way to play around with Windows on a mac, for free.

src/logic/events/TrackpadEvents.swift
ris58h: It turned out that OpenMultitouchSupport has the same issue https://github.com/Kyome22/OpenMultitouchSupport/issues/2 It seems like we have to patch anyway.

Should I commit changes in `Pods` dir after a new pod is installed?

ris58h:
@lwouis it almost works now. Things to do:
- Settings layout in `Gesture` tab. It's just a cosmetic issue.
- Patch M5MultitouchSupport lib.
- 360 movement.
- `On release do nothing` for the swipe.

I believe we should go with MVP. 360 movement and `On release do nothing` aren't that important. What do you think?

src/ui/preferences-window/tabs/ControlsTab.swift
```diff
@@ -90,14 +93,28 @@ class ControlsTab {
...
+        var holdShortcut = LabelAndControl.makeLabelWithRecorder(NSLocalizedString("Hold", comment: ""), Preferences.indexToName("holdShortcut", index), Preferences.holdShortcut[index], false, labelPosition: .leftWithoutSeparator)
+        holdShortcut.append(LabelAndControl.makeLabel(NSLocalizedString("and press:", comment: "")))
+        let holdAndPress = StackView(holdShortcut)
+        let nextWindowShortcut = LabelAndControl.makeLabelWithRecorder(NSLocalizedString("Select next window", comment: ""), Preferences.indexToName("nextWindowShortcut", index), Preferences.nextWindowShortcut[index], labelPosition: .right)
+        let tab = controlTab(index, [holdAndPress, StackView(nextWindowShortcut)])
+        return (holdShortcut, nextWindowShortcut, tab)
+    }
+
+    //TODO: controls in Gesture tab aren't centered like in Shortcut tab
+    //TODO: 'On release do nothing' doesn't work. See ShortcutStylePreference.focusOnRelease usage and ATShortcut.shouldTrigger usage
```
ris58h, line 106: For keyboard shortcut 'On release do nothing' is implemented via skipping shortcut action. I haven't found workaround to the swipe yet. We could just not support such option for swiping.

lwouis:
> I believe we should go with MVP. 360 movement and On release do nothing aren't that important. What do you think?

This app has no deadline pressure. We can take time to polish things before shipping 👍

Shahriarnz14:
Hi just wondering, if there is an update on this feature yet? :) 
Find it as the only missing piece of an already amazing work!

ris58h:
@Shahriarnz14 I haven't touched it for a while - I use [Touch-Tab](https://github.com/ris58h/Touch-Tab) instead. The current state of the PR is described in this [comment](https://github.com/lwouis/alt-tab-macos/pull/2355#issuecomment-1445177973). I'm new to macOS development so there are some issues:
- AltTab uses some hacks to make layout looks good but `Settings` layout is broken for the new `Gesture` tab because it has different layout than `Shortcut` tabs.
- AltTab uses background threads to process system events but I couldn't make a separate thread that accepts touch events.

lwouis:
@ris58h I just checked out the branch, and played with it. You've done some really good work. The feature works well. I'm really impressed you were able to implement the event tracking without the third-party library. I remember trying in the past and hitting a wall, so I'm really impressed. Thank you for all the work here!

It's very promising, but there needs to be a few more things before we can ship it. In addition to the checklist you shared above, I would add one: clarifying conflicts/issues with the native gestures to the user. For instance, should we have AltTab use gestures by default? Would it work on a default macOS install? Would it conflict with other gestures? We probably should add some form of tooltip/warning when the user could get a broken experience. We could (probably) read the state of some of the System Settings, and depending on the situation, warn the user to toggle some things off to not conflict with AltTab. Some of the TODOS you've written in the PR seem to show non-obvious conflicts, for instance:

> //TODO: underlying content scrolls if both Mission Control and App Expose use 4-finger swipes or are off in Trackpad settings. It doesn't scroll if any of them use 3-finger swipe though.

billjcnickolas:
I'm also interested in this PR. Any chance there's an update on this?
Migrating from Windows to Mac would be hell of a lot easier if this is added.

ris58h:
@billjcnickolas it was just a PoC. I'm stuck on [some issues](https://github.com/lwouis/alt-tab-macos/pull/2355#issuecomment-1509563153) but I'm not going to spent my time on them anymore.

billjcnickolas:
@ris58h Thank you for your response. It is indeed a really good PoC and I feel like it's really close to completion. I'm not familiar with swift development at all otherwise would've wanted to help.
I feel like it's only lacking one small push to get completed.
I wonder if @lwouis can take a look at this some time and see if it can make it to the end product!
I realized there are many others looking for this feature (from various youtube channels recommending alt-tab-macos to gh issues and even in this pr) that I think there's enough demand for it. 🤞

src/logic/events/TrackpadEvents.swift
```diff
@@ -0,0 +1,150 @@
+import Cocoa
+
+fileprivate var eventTap: CFMachPort!
+fileprivate var shouldBeEnabled: Bool!
+
+//TODO: Should we add a sensetivity setting instead of these magic numbers?
+fileprivate let accVelXThreshold: Float = 0.05
+fileprivate let accVelYThreshold: Float = 0.075
+fileprivate var accVelX: Float = 0
+fileprivate var accVelY: Float = 0
```
Jack-Works, line 10:
```suggestion
fileprivate let accVelXThreshold: Float = 0.03
fileprivate let accVelYThreshold: Float = 0.06
fileprivate var accVelX: Float = 0.35
fileprivate var accVelY: Float = 0.125
```

Thanks @lwouis for your work! I also tried your touch-pad and found the same problem with this PR. The sensitivity is not matched with Windows (in my experience). I tried different numbers and found that 0.03, 0.06, 0.35, and 0.125 are more sensitive and like how Windows works.

ris58h:
@lwouis we should close this PR because it's already implemented

PR #2399, opened 2023-03-07
feat: preview selected window

decodism:
Implements #967.

### DIFF
--- a/src/ui/main-window/PreviewPanel.swift
+++ b/src/ui/main-window/PreviewPanel.swift
@@ -1,0 +1,27 @@
+import Cocoa
+
+class PreviewPanel: NSPanel {
+    private let previewView = NSImageView()
+
+    convenience init() {
+        self.init(contentRect: .zero, styleMask: [.nonactivatingPanel, .titled, .fullSizeContentView], backing: .buffered, defer: false)
+        isFloatingPanel = true
+        animationBehavior = .none
+        hidesOnDeactivate = false
+        titleVisibility = .hidden
+        titlebarAppearsTransparent = true
+        backgroundColor = .clear
+        contentView = previewView
+        // triggering AltTab before or during Space transition animation brings the window on the Space post-transition
+        collectionBehavior = .canJoinAllSpaces
+        // 2nd highest level possible; this allows the app to go on top of context menus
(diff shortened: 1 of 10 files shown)


src/ui/main-window/PreviewPanel.swift
```diff
@@ -0,0 +1,28 @@
...
+        contentView! = previewView
+        // triggering AltTab before or during Space transition animation brings the window on the Space post-transition
+        collectionBehavior = .canJoinAllSpaces
+        // 2nd highest level possible; this allows the app to go on top of context menus
+        // highest level is .screenSaver but makes drag and drop on top the main window impossible
+        level = .popUpMenu
+        // helps filter out this window from the thumbnails
+        setAccessibilitySubrole(.unknown)
+        // for VoiceOver
+        setAccessibilityLabel(App.name)
```
lwouis, line 22: Maybe this is not useful here?

lwouis:
This is very impressive. Thank you @decodism! It's very snappy. Almost too snappy: I wonder if it's clear enough when the previews are popping up, that they just did. It's easy to miss maybe. I think maybe we could copy what HyperSwitch is doing here: they are adding a blue shadow to the window, to signal that it's a preview, not an actual window. They are also making the window a bit transparent, which is interesting.

I'm not a fan of the delay they have, before showing the preview, however I like the idea to have some visual clue that the preview is a preview. The shadow is interesting; we could play around with something like this. The transparency is also interesting, but I'm not sure it's that useful.

Besides the UX, I'm afraid I faced a bug during my tests. It's easy to reproduce:

* Launch AltTab (with previews enabled)
* Alt+tab once -> no preview is shown
* Further tabs or change in selection -> previews show correctly

It's an issue on the first summon which doesn't preview the most-recently-active window.

Do you think you could have a look?

I reviewed the code, and it's very thoughtful and concise 🙇

lwouis:
Hi @decodism,

Thank you for your amazing work, as usual!

The outline your added looks really nice. I've taken the liberty to change it from a fixed red, to instead use `NSColor.controlAccentColor` which can be customized by the user in System Settings:

image

On older macOS versions, I defaulted to a blue outline that matches the current default blue on macOS 13.2.1.

I'll ship this feature in the next release 👍

PR #2413, opened 2023-03-17
fix: update preview window in more cases

decodism:
For example, when closing the focused window.


lwouis:
Indeed, I was able to reproduce the issue. Thank you for the fix. I'll ship it in the next release 👍

Issue #337, opened 2020-05-13
Two recommendations on the use of software


iheshun:
I'm sorry, I'm a Chinese person, so I'm using translation software to write this paragraph, if there is a problem with the expression, please understand.

Currently using the software you provided and it's great. But in use, I have 2 small comments that I hope you will consider.

1、Does activating (command+tap) the "panel" allow setting shortcut keys? This allows me to set shortcuts on the mouse and activate the panel with one click.
2、After the panel is activated, the current App is sorted according to the order of the window, whether you can support sorting by application category. For example, 1(word)3(word),2(Powerpoint),3(Excel),3(Evernote),4(Facetime)

Finally thanks again for making such great software and good luck!

Translated with www.DeepL.com/Translator (free version)

lwouis:
Hi @iheshun! Thanks for sharing this feedback!

I will try to answer your suggestions:

> 1. Does activating (command+tap) the "panel" allow setting shortcut keys

You can change the shortcuts easily in the preferences panel:

![image](https://user-images.githubusercontent.com/106195/81789298-baa35a00-953e-11ea-8ea5-75e42cce7cc9.png)

There you can configure every shortcut:

![image](https://user-images.githubusercontent.com/106195/81789339-cd1d9380-953e-11ea-9b9d-7903a9a90f91.png)

If you want to trigger the shortcuts using the mouse, most mouse manufacturers let you assign keyboard keys to a mouse click. On my Logicool mouse for example, I set it like this:

![image](https://user-images.githubusercontent.com/106195/81789480-02c27c80-953f-11ea-8d53-0436d0d4a332.png)

For example, you would set your mouse click4 to send `alt-tab`, then you would set AltTab to trigger on `alt-tab`. You can change the keyboard shortcut of course.

> 2. After the panel is activated, the current App is sorted according to the order of the window, whether you can support sorting by application category. For example, 1(word)3(word),2(Powerpoint),3(Excel),3(Evernote),4(Facetime)

It is an interesting idea, but I think it is a bit complex and may not be as useful as it seems. Let's say I have this today:

Word1, Powerpoint, Excel, Evernote, Facetime, Word2, Word3, Word4

If we group by app, then you would see this:

Word1, Word2, Word3, Word4, Powerpoint, Excel, Evernote, Facetime

So even though you haven't opened Word2, Word3, Word4 maybe in a long time, they are taking up all the prime real-estate on the UI, just because you recently opened Word1.

[...]

PR #2583, opened 2023-06-09
feat: allow choosing windows order

XOR-op:
This PR tries to fix https://github.com/lwouis/alt-tab-macos/issues/515. Since the current PR https://github.com/lwouis/alt-tab-macos/pull/1936 is inactive and there are still some problems left, I opened this PR for an alternative.

**Summary**
This PR introduces an option called "Window order". The four options available now is:
- Last Focused: the default behavior same as the original alt-tab does.
- Last Created: we record the creation time of a window, then sort the windows by descending time order, i.e. eariler created window will be put more at the end.
- Alphabetical: we sort the windows first by its application name, then by its window title. A lot of users want this mainly because they expect a deterministic order. Since the window title is more often to be changed (e.g. a browser window title will show the title of the webpage), application name first should be a better choice.
- Space: we sort the windows first by its space index, then use the same rule in Alphabetical. Users need this also because of a deterministic order. And they also organize windows by space (e.g. work stuff on space 1, personal stuff on space 3).

**Changes compared to #1936**
- I don't add "Most Frequently" here: how to define "frequency" is unclear, the "frecency" as well. I have tried some algorithms but they look a little weird and out of expectation. I think we can add this in later PR with more deep discussion. We can add an option to change order first, then decide what order should be added.
- I renamed "Last Used" to "Last Focused", and "Last Opened" to "Last Created". This should be unambiguous, since we won't say "focus an app" or "create an app".
- For order other than "Last Focused", I will make the initial focused window the same as the last focused window. When we use order other than "Last Focused", our window may (most probably!) not the first window in the order. Initially focus the second window don't make sense. And since users may not navigate right, move the focused window right by 1 is not a good choice. So I decide to make it the same as the last focused window, leaving the user to choose where to navigate.

**Other things**
In #1936 you state that the original PR failed to keep order for minimized/hidden window. I cannot reproduce the wrong behavior. I use the similar logic to #1936 and it looks fine to me. Can you provide more details to point out what is wrong?

### DIFF
--- a/src/logic/Preferences.swift
+++ b/src/logic/Preferences.swift
@@ -47,11 +47,11 @@ class Preferences {
         "showFullscreenWindows3": ShowHowPreference.show.rawValue,
         "showFullscreenWindows4": ShowHowPreference.show.rawValue,
         "showFullscreenWindows5": ShowHowPreference.show.rawValue,
-        "windowOrder": WindowOrderPreference.lastUsed.rawValue,
-        "windowOrder2": WindowOrderPreference.lastUsed.rawValue,
-        "windowOrder3": WindowOrderPreference.lastUsed.rawValue,
-        "windowOrder4": WindowOrderPreference.lastUsed.rawValue,
-        "windowOrder5": WindowOrderPreference.lastUsed.rawValue,
+        "windowOrder": WindowOrderPreference.lastFocused.rawValue,
+        "windowOrder2": WindowOrderPreference.lastFocused.rawValue,
+        "windowOrder3": WindowOrderPreference.lastFocused.rawValue,
+        "windowOrder4": WindowOrderPreference.lastFocused.rawValue,
+        "windowOrder5": WindowOrderPreference.lastFocused.rawValue,
         "showTabsAsWindows": "false",
         "hideColoredCircles": "false",
         "windowDisplayDelay": "0",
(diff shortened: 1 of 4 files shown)


lwouis:
Hi @XOR-op,

Thank you for sharing this PR. It's quite thoughtful and from my tests it seems to work as advertised.

I have a few comments:

* For `Last Created`, I wonder if using `kCGWindowNumber` would work to guess the creation order before AltTab was launched. Maybe they are sequential, and we could use it instead of defaulting to `Last Focused`?
* For `Alphabetical`, I'm not sure if everyone would agree to sort by app name > window title, vs directly window title. I don't understand the idea of not using `Last Focused`, so I can't really put myself in the shoes of the user who would want this. I imagine what you wrote can make sense. I'm just wondering if it will indeed, when released, for the few people who will use that order.
* For `Space`, it really feels to me like such a user would really want #337. Sorting by Space looks really weird to me when not visually separating the Spaces with some UI like a separator line, background colors for each Space, or something. It looks scrambled, especially if the user is not showing Space numbers on windows.

> For order other than "Last Focused", I will make the initial focused window the same as the last focused window  
 
This is really weird. I understand your approach, yet I can't help but feel it's weird. Again, I don't understand why anyone would do that, so I guess it's what those users would want?

An exception though: I think for `Last Created`, the user may actually want to use Created instead of Focused. And not have thumbnail selection follow Focused still.

> In https://github.com/lwouis/alt-tab-macos/pull/1936 you state that the original PR failed to keep order for minimized/hidden window. I cannot reproduce the wrong behavior. I use the similar logic to https://github.com/lwouis/alt-tab-macos/pull/1936 and it looks fine to me. Can you provide more details to point out what is wrong?

I tested, and it seems to work in your MR: each section follows the order from the dropdown, but the sections are separated by [standard windows | minimized | hidden | apps].

lwouis:
Thank you for clarifying things. It is much appreciated.

I'm a bit worried with the `Alphabetical` and `Last Created` order since you and I don't/won't use these, and the people asking for those are not speaking up here, and explaining what they would like to have. That being said, I guess we can just tweak the behavior later, as people send feedback.

Regarding the last point about `Last Created`, I'll try to explain a bit better. I think that someone who wants to sort by `Last Created` wants an experience a bit similar to what the Dock or the Windows bottom-bar does. They want to order windows based on when during the day that window was created. It seems then, that they would like the initial selection to copy what it does today, but using `Last Created` instead of `Last Focused`. Today, `Last Focused` picks the second-most-recently-focused window, uppon AltTab's UI showing up, then every `tab` will go to a less-recently-focused window. I think if the user sorts by `Last Created`, they would probably want the initial selection to pick the second-most-recently-create window, uppon AltTab's UI showing up, then every `tab` will go to a less-recently-created window. What do you think?

Thank you

lwouis:
I tried to use `kCGWindowNumber`. The numbers are not always given to windows sequentially, so we can't rely on it. I kept your implementation, and extended it to windowless apps as well

lwouis:
The content of this PR have been cherry-picked onto the master branch, and the latest release of AltTab ships with it.

Thank you for your contribution 🙇

PR #2589, opened 2023-06-16
fix: support CrossOver and ColorSlurp windows

zacharee:
This PR adds two new functions to `AXUIElement`:

- `colorSlurp()`
- `crossoverWindow()`

For whatever reason, ColorSlurp's window that pops up when you pick a color isn't presented as a normal window, so I made it so that if the window has the ColorSlurp bundleID, it will always be included. ColorSlurp doesn't seem to have any other window types, so this should be safe.

CrossOver windows are very weird, probably because of how hacky the whole WINE layering stuff is in general. I'm not fully convinced that the method I used for checking for CrossOver windows is safe, but it doesn't seem to hit any false positives in my testing. (Maybe it should be a default-disabled setting?)

With Game Porting Toolkit being based on CrossOver, and it already becoming popular, it's probably a good idea to be able to handle those windows.

### DIFF
--- a/src/api-wrappers/AXUIElement.swift
+++ b/src/api-wrappers/AXUIElement.swift
@@ -84,27 +84,31 @@ extension AXUIElement {
 
         // Some non-windows have cgWindowId == 0 (e.g. windows of apps starting at login with the checkbox "Hidden" checked)
         return wid != 0 && size != nil &&
-            (books(runningApp) || keynote(runningApp) || iina(runningApp) || openFlStudio(runningApp, title) || (
-                // CGWindowLevel == .normalWindow helps filter out iStats Pro and other top-level pop-overs, and floating windows
-                level == CGWindow.normalLevel &&
-                    ([kAXStandardWindowSubrole, kAXDialogSubrole].contains(subrole) ||
-                        openBoard(runningApp) ||
-                        adobeAudition(runningApp, subrole) ||
-                        adobeAfterEffects(runningApp, subrole) ||
-                        steam(runningApp, title, role) ||
-                        worldOfWarcraft(runningApp, role) ||
-                        battleNetBootstrapper(runningApp, role) ||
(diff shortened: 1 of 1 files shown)


src/api-wrappers/AXUIElement.swift
```diff
@@ -200,6 +205,15 @@ extension AXUIElement {
         return title != "" && Applications.isAndroidEmulator(runningApp)
     }
 
+    private static func colorSlurp(_ runningApp: NSRunningApplication) -> Bool {
+        // ColorSlurp presents its dialog as a kAXSystemDialogSubrole, so we need a special check
+        return runningApp.bundleIdentifier == "com.IdeaPunch.ColorSlurp"
+    }
+
+    private static func crossoverWindow(_ runningApp: NSRunningApplication, _ role: String?, _ subrole: String?, _ level: CGWindowLevel) -> Bool {
+        runningApp.bundleIdentifier == nil && role == kAXWindowRole && subrole == kAXUnknownSubrole && level == 0
```
lwouis, line 214: This check here is too broad. Many app could have `runningApp.bundleIdentifier == nil`. We don't want those to have all their windows shown without filtering.

Maybe try and pin-point this CrossOver app with something like what I did for `androidEmulator`?

As it stands, I can't merge this, unfortunately. It's too broad.

zacharee: It looks like anything launched from Game Porting Toolkit/CrossOver/WINE has a "localized" app name of `wine64-preloader`. I added that to the condition.

zacharee: Actually this won't work. Looking into it more...

lwouis:
The content of this PR have been cherry-picked onto the master branch, and the latest release of AltTab ships with it.

Thank you for your contribution 🙇 

If you manage to get the CrossOver whitelist code working, could you please share a new PR for just that?

PR #2609, opened 2023-06-23
fix: support CrossOver windows

zacharee:
This should have safer CrossOver checking.

On stable CrossOver and Game Porting Toolkit, all windows have a `localizedName` of `wine64-preloader`. 

However, it looks like in the nightly versions of CrossOver, they're naming each app with its `.exe` file name, which means checking for `wine64-preloader` alone doesn't work. To get around this, I also check to see if the executable path contains `/winetemp-`, which all EXEs run from CrossOver appear to contain.

It's still not very clean, but I think it's very unlikely that a non-WINE app will include `/winetemp-` in its executable path. It might be possible to make this more strict by also checking if the path ends with `.exe` or `.msi`, but that starts getting into table-checking territory and might hit a lot of false negatives where false positives are less likely as-is.

### DIFF
--- a/src/api-wrappers/AXUIElement.swift
+++ b/src/api-wrappers/AXUIElement.swift
@@ -84,27 +84,31 @@ extension AXUIElement {
 
         // Some non-windows have cgWindowId == 0 (e.g. windows of apps starting at login with the checkbox "Hidden" checked)
         return wid != 0 && size != nil &&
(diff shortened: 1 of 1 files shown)


lwouis:
Thank you for sharing this work. I've cherry-picked this change on the dev branch. Closing this PR. This will ship in the next release 👍

PR #2610, opened 2023-06-24
fix: filter out empty ColorSlurp window

zacharee:
A small update to the ColorSlurp checking to filter out an invisible/empty window that shows up in the list for a second or two after picking a color.


lwouis:
Thank you for sharing this work. I've cherry-picked this change on the dev branch. Closing this PR. This will ship in the next release 👍

PR #2620, opened 2023-06-28
fix: replace deprecated `activate()` calls with a hidden API

zacharee:
macOS 14 is deprecating the `NSRunningApplication.activate(options: .activateIgnoringOtherApps)` option. While not removed, it claims it will no longer have any effect.

This PR replaces that call (and similar calls to `NSApplicationDelegate.activate(ignoringOtherApps: true)`) with a hidden API with the same functionality where relevant, and by simply removing the option elsewhere.

As far as I can tell, `CGSSetWindowActive()` was added in macOS 10.10. It's also still present in the current macOS 14 beta SDK. That obviously isn't a guarantee that it won't be removed when the stable SDK comes out, and it looks like the old behavior of `ignoringOtherApps` is still working in the betas, so it probably doesn't make sense to merge this PR right away.

It would be good to have other people test this out, though. I'm not sure how to reproduce the deprecation warning dialog when running a debug app built on the test computer, so I don't know if this resolves everything.


koekeishiya:
Did you test this? 

Most of these functions are proxyed into the SkyLight.framework (SkyLight is the modern replacement for CoreGraphics on newer versions of macOS), and the implementation of the corresponding function is literally a NO-OP. See attached screenshot:

Screenshot 2023-06-29 at 14 28 17

zacharee:
I did test it. Windows properly focus for me when selecting them in alt-tab.

zacharee:
Actually, you're right. Something isn't working.

zacharee:
Is the `__XSetWindowActive` function also a no-op?

koekeishiya:
`__X` prefixes are server-side functions available to the `WindowServer` only.`CGS` and `SLS` are client-side, which lets user space applications send mach messages to the server running inside the `WindowServer` process.

PR #2624, opened 2023-06-28
feat: support for always-on-top scrcpy windows

zacharee:
scrcpy uses the floating window level when the "always on top" option is enabled, which means it doesn't get picked up by alt-tab.

This PR does a check for scrcpy windows at that level and adds them to the valid window list.


lwouis:
I'll take these changes on the dev branch. They will be shipped in the next release.

I also suggest you contact scrcpy maintainers so they can update the mac app to not require this hacky support from AltTab. They should definitely have a bundle, and handle windows in a more expected way. They are on github, so they can probably improve macOS support quickly 👍

zacharee:
One of the scrcpy developers seems unsure if it's possible since it's not an Xcode project: https://github.com/Genymobile/scrcpy/issues/3684.

I don't personally know much about how to specify bundle IDs either, so I don't think I can be of much help to them.

PR #2673, opened 2023-07-14
feat: support Xcode 15 splash window

zacharee:
The Xcode 15 splash window uses an "Unknown" subrole.

### DIFF
--- a/src/api-wrappers/AXUIElement.swift
+++ b/src/api-wrappers/AXUIElement.swift
@@ -106,7 +106,8 @@ extension AXUIElement {
                         sanGuoShaAirWD(runningApp) ||
                         dvdFab(runningApp) ||
                         drBetotte(runningApp) ||
-                        androidEmulator(runningApp, title)
+                        androidEmulator(runningApp, title) ||
+                        xcodeSplash(runningApp, level, role, subrole)
                 ) && (
                     mustHaveIfJetbrainApp(runningApp, title, subrole, size!) &&
                         mustHaveIfSteam(runningApp, title, role) &&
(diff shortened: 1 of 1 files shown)


lwouis:
Hi @zacharee,

Thank you for sharing this. I notice that you've shared multiple PRs around whitelisting specific apps. I wanted to discuss this topic with you.

Apps should be implemented properly. Ideally apps don't have bugs. When they have bugs, users report them to the app makers, and the bugs get fixed. Windows being correctly implemented is a general accessibility issues. It can break basic OS window management (e.g. when you close Slack on macOS, it actually hides the window instead of closing it), and it will most likely break accessibility apps. AltTab is an accessibility app. It uses Accessibility APIs to learn about other windows from other apps, and share that data with the users.

We opened the door to some whitelisting, for practical reasons. There are popular apps which are ported from other platforms. Such apps have very low chance to get rewrote deeply by their makers, to accomodate the mac platform. For instance Steam. It's a huge app used by many people. It's a QT app I think? It's a terrible implementation on macOS. Dropdown menus are coded as windows for instance. There is no chance, from where I stand, that Steam will get reworked. It would cost a fortune, would mean lots of risks, for little benefit. It's not going to happen, and that's why AltTab whitelisted it with detailed rules, so users get a decent experience on this popular app.

So the whitelisting is a practical compromise. What it should not become is a go-to solution anytime an app has a bug. If XCode 15 works great, but has this one window which is bugged, and AltTab reflects the bug, I don't think we should implement a custom workaround. We should keep it bugged, and hopefully Apple ends up fixing it from getting feedback from people using accessibility apps which are broken with that one window.

I hope I was able to make the case about the trade-off behind whitelisting apps, and why I think there are good candidates, and less good candidates. I understand the line is blurry and subjective, so I'm sorry if our judgement differs on case-by-case.

I hope you understand 🙇

PR #2750, opened 2023-08-24
Vulnerabilities fix 

honey-donkey:
fix: [docs/Gemfile & docs/Gemfile.lock to reduce vulnerabilities] after scanned by snyk


lwouis:
Hi @honey-donkey,

Thank you for sharing this PR. These files are involved in generated the [static site](https://alt-tab-macos.netlify.app/) for the project. Generation happens during CI, and is fully server-side. It doesn't affect the app in any capacity.

Thus, upgrades are not really needed here. I would rather keep the dependencies the way they are, and prevent some kind of regression or incompatibility.

Thank you

PR #2865, opened 2023-10-09
feat: add vim key window navigation

### LINKED ISSUE #1229: allow movement with VIM arrows (H J K L)
It would be mind-blowingly amazing if, among the arrow keys, it would be possible to move around with the vim movement keys (that, as we all know, are fairly superior and should also replace steering wheels and handlebars, but I'll submit that feature suggestion elsewhere)

[Hyperswitch](https://bahoom.com/hyperswitch/get) got me used and now it's difficult to live without
and it's a shame because as far as I know, Hyperswitch development ended and, except for this basic and necessary feature, Alt-Tab is a vastly superior program

The main advantage is that those keys are right where the right hand should be (the J key has a tactile bump for touch typing) and there is no need to blindly move it to the other side of the keyboard or, god forbid, shamelessly look down at the keyboard looking for those keys far in the right beyond any key row.
Also, users not blessed with this habit usually have their right hands sitting over inferior devices such as trackballs or, in unrecoverable cases, over a mouse. So they would hardly press those legendary keys by mistake, rendering this feature not confusing for people with little confidence with computers.

it basically works like this: (unfortunately not my keyboard, as I don't yet have a dedicated key for μ)
image

feel free to state your perplexities and I'll gladly explain why you are all wrong

skolj:
This change adds the ability to use vim keys (h, j, k, l for left, down, up, and right respectively) to cycle through windows with the _Vim Keys_ option. Closes #1229.

AltTab uses the `h` key to hide/show app. When the _Vim Keys_ option is turned on, a conflicting shortcuts alert is presented with options to: 
1. Unassign the existing shortcuts and activate vim keys which clears the existing conflicting shortcuts.
2. Cancel activating vim keys.

This PR is heavily inspired by #1272 and implements the UX discussed in the PR.

### Screenshots

**Vim keys option**

Screenshot 2023-10-09 at 4 44 45 PM

**Alert upon checking the Vim Keys option with conflicting shortcuts**

Screenshot 2023-10-09 at 4 45 26 PM

**Cancel the alert**
Screenshot 2023-10-09 at 4 45 40 PM

**Upon clicking unassign shortcuts button (Hide/show app shortcut is cleared)**
Screenshot 2023-10-09 at 4 45 57 PM

**Assigning a vim key shortcut to other action with Vim Keys enabled**
Screenshot 2023-10-09 at 4 50 30 PM

**Upon clicking unassign shortcut button (Vim Keys is disabled)**
Screenshot 2023-10-09 at 4 50 47 PM

### DIFF
--- a/src/logic/Preferences.swift
+++ b/src/logic/Preferences.swift
@@ -30,6 +30,7 @@ class Preferences {
         "quitAppShortcut": "Q",
         "hideShowAppShortcut": "H",
         "arrowKeysEnabled": "true",
+        "vimKeysEnabled": "false",
         "mouseHoverEnabled": "false",
         "cursorFollowFocusEnabled": "false",
         "showMinimizedWindows": ShowHowPreference.show.rawValue,
@@ -125,6 +126,7 @@ class Preferences {
     static var quitAppShortcut: String { defaults.string("quitAppShortcut") }
     static var hideShowAppShortcut: String { defaults.string("hideShowAppShortcut") }
     static var arrowKeysEnabled: Bool { defaults.bool("arrowKeysEnabled") }
+    static var vimKeysEnabled: Bool { defaults.bool("vimKeysEnabled") }
     static var mouseHoverEnabled: Bool { defaults.bool("mouseHoverEnabled") }
     static var cursorFollowFocusEnabled: Bool { defaults.bool("cursorFollowFocusEnabled") }
     static var showTabsAsWindows: Bool { defaults.bool("showTabsAsWindows") }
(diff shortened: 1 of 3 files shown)


lwouis:
Hi @skolj,

Thank you very much for sharing this work!

I've cherry-picked this to the dev branch. I've made a few edits. Notably, there was a bug when if you would check the `Vim keys` checkbox, then cancel, the `defaults` (i.e. the user preferences in the macOS DB) would still have `vimKeysEnabled = true`. Of course, I've kept you as the original commit author, so you'll be correctly credited both in the website and in the app 👍 

Closing this PR. I'll release a new version soon which will include this feature 👍 

Thank you

skolj:
Awesome! Thank you.

Ethansev:
Came back here just to say thank you! So glad this feature exists.

PR #2968, opened 2023-11-08
feat: adding shadows

Allsochen:
Adding shadows makes the panel easier to distinguish against window backgrounds of different colors.

* The feature of adding shadow is inconsistent with the default window switching style of mac system.
* But it has the same style as the dock, both with shadows.


lwouis:
Hi @Allsochen,

I'm afraid this PR can't be merged. We are trying to keep close to the native `command+tab` switcher.

Thank you 🙇

Allsochen:
> Hi @Allsochen,
> 
> I'm afraid this PR can't be merged. We are trying to keep close to the native `command+tab` switcher.
> 
> Thank you 🙇

Maybe it would be better to add an option in appearance for users to choose?

lwouis:
We have a bit too many preferences currently. Please see #351 for more info 👍

Issue #2988, opened 2023-11-16
Differentiate between left and right modifier keys


LiamKearn:
Is there any way Alt-Tab could differentiate between left and right modifiers?

This would allow a lot of apps to pass through what I imagine are common shortcuts via utilising a non-prefered set of modifiers. e.g. I would love to be able to use Left-CMD-Tab to switch windows with alt tab and then Right-CMD-Tab to cycle tabs in my browser or other applications. Others may prefer to use Right-CMD for AltTab and be able to fall back to their mac default when they're using that.

A quick google&github search yields the [following snippets](https://github.com/search?q=.contains%28.rightCommand%29&type=code) which might be one route but I have no background in swift, if this gets 10 reactions from others that would find this useful and someone @'s me I'll contribute it anyway :) 

My initial (probably horrid) idea for UI is to add a simple dropdown in the same section as the modifier key which allows selecting from:
- Both - Both modifiers will trigger the shortcut
- Left - Only the left hand modifiers will trigger the shortcut
- Right - Only the right hand side modifiers will trigger the shortcut

Here is a very quick and dirty mockup I made in preview by copying and pasting the ui elements:
<img width="577" alt="Screenshot 2023-11-16 at 9 32 07 pm" src="https://github.com/lwouis/alt-tab-macos/assets/76269376/8372649a-5483-43ca-801a-9f35c4ae52e2">

Please upvote to encourage me to be productive here 👍🏻

lwouis:
Hi @LiamKearn,

This is an interesting idea. I'm not sure if it's doable at a fundamental level. I'm not sure all keyboards send different signals for left or right modifiers.

In any case, we use the great [ShortcutRecorder](https://github.com/Kentzo/ShortcutRecorder/) library to handle shortcut capture/customization in the Preferences. You'll have to work there to achieve this ticket. I suggest you get in touch with the main maintainer before moving forward. This will also help clarify if he will merge your PR, or if we will need to add your commits to my fork of the repo.

Thank you

gabrielkuslansky:
> I'm not sure if it's doable at a fundamental level. I'm not sure all keyboards send different signals for left or right modifiers.

I believe it is technically possible per [rcmd](https://lowtechguys.com/rcmd/), but I'm unsure how non-trivial it will be to implement and handle all keyboards.

GabeKuslansky:
@lwouis 
[...]

PR #2993, opened 2023-11-17
feat: add shortcut modifier side

decodism:
My attempt at #2988.

### DIFF
--- a/src/api-wrappers/HelperExtensions.swift
+++ b/src/api-wrappers/HelperExtensions.swift
@@ -189,3 +189,17 @@ extension Optional where Wrapped == String {
         return (self ?? "").localizedStandardCompare(string ?? "")
     }
 }
+
+extension NSEvent.ModifierFlags {
+    static let leftShift = Self(rawValue: UInt(NX_DEVICELSHIFTKEYMASK))
+    static let rightShift = Self(rawValue: UInt(NX_DEVICERSHIFTKEYMASK))
+    
+    static let leftControl = Self(rawValue: UInt(NX_DEVICELCTLKEYMASK))
+    static let rightControl = Self(rawValue: UInt(NX_DEVICERCTLKEYMASK))
+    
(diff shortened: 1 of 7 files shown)


LiamKearn:
@decodism Thank you, I've tested locally and this is working fabulously for my use case of using rCMD+Tab for the browser.

lwouis:
Hi @decodism,

I tested this locally, and it seems to work correctly. I also works for multiple modifiers: all of them need to be on the specified side, for the shortcut to register.

I have an issue with this though: the UX. Merging this PR would mean adding yet another dropdown menu to the preferences UI. It's already very complex and rich, and I'd like to actually decrease the number of elements shown (see #351).

Furthermore, the position of this dropdown is not very clear I think. It says `modifier` when there could be multiple modifiers, and it's a bit far away from the left input-box, so it could be confusing for people who set for example `alt+shift` with `alt` on the left input-box, and `shift` on the right input-box. They may be confused as to which modifier is referenced by the label.

Overall, i think this feature is a bit too niche to be made mainstream. I'm not sure if there is a way to have a better UX that wouldn't confuse the 99.9%  of users who don't need and don't know that both sides of the keyboard can even be distinguished.

What do you think?

Thank you 🙇

LiamKearn:
@decodism The latest changes seem to break things for me.

When using a modifier side (`L` or `R`) windows don't focus on shortcut release (they do if I manually click them). Using `""` they work once-more.

LiamKearn:
@lwouis RE: the UI issues, I'm guessing you're opposed to merging "defaults only" features on the grounds that it may add complexity while not getting used by enough users?

e.g. I mean features that are only exposed by documenting that you could maybe do "defaults write etc. etc." to turn on/off xyz.

lwouis:
Hi @LiamKearn,
Yes indeed. I'd like to avoid adding features that are only used by a few people. 
Thank you

Issue #1901, opened 2022-08-24
Difficult to see which window is selected on dark background


heaven:
I use white MacOS theme but my IDE is configured to use a dark background in the editor. When switching between windows it is difficult to see which one is selected when on a dark background.

Here's an example
<img width="1373" alt="Screenshot 2022-08-24 at 18 16 57" src="https://user-images.githubusercontent.com/126732/186470634-63941b12-6bb2-484f-900b-3ba27f4c23dc.png">

An ability to change the selection color would solve the problem.

lwouis:
This has been discussed in details in a few tickets. Please see my 2 comments here: https://github.com/lwouis/alt-tab-macos/issues/1867

heaven:
@lwouis I see. The fact that this many similar tickets were open clearly shows there's a UX problem. I tried Windows theme but it looks odd. I think having a light Mac OS theme that would look similar to the native one would be a great solution, even if it's not as well rendered as the native one. The native switcher has a good color scheme that works in both dark and light environments (except the font, which is hardly readable).

lwouis:
Please read https://github.com/lwouis/alt-tab-macos/issues/82#issuecomment-577480530

PR #3193, opened 2024-02-21
feat: add support for macOS accent color

volskaya:
Adds an option to appearances to highlight `ThumbnailView` background with macOS accent color instead of the default black, while they're using the macOS theme.

*In the picture it's using the hidden iMac purple accent that everyone can unlock with:*
`defaults write -g NSColorSimulateHardwareAccent -bool YES`
`defaults write -g NSColorSimulatedHardwareEnclosureNumber -int 7`

### DIFF
--- a/src/logic/Preferences.swift
+++ b/src/logic/Preferences.swift
@@ -95,6 +95,7 @@ class Preferences {
         "hideWindowlessApps": "false",
         "hideThumbnails": "false",
         "previewFocusedWindow": "false",
+        "showMacosAccentColor": "false",
     ]
 
     // system preferences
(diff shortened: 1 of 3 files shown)


lwouis:
Hi @volskaya,

This is an interesting idea. Thank you for sharing a PR.

People have been asking to customize highlight color for a long time. [Here's my opinion on it](https://github.com/lwouis/alt-tab-macos/issues/1901#issuecomment-1225971978).

I'm afraid I can't merge this PR for a few reasons:
* When changing the accent color, the default command+tab macOS app-switcher doesn't change its highlight color. AltTab is always trying to stay close to the default app-switcher, so it feels native, and almost looks like Apple made it.
* A user may want to use the accent color, but they also want another color. Letting the user pick a color from a color picker here seems like a more open approach. People could pick what they like. See my comments above on limitations of that approach.
* It adds yet another toggle box in the preferences. We are trying currently to reduce their number (see #351).

I'm sorry to not merge the PR. I hope you understand the contraints we're in.

Thank you 🙇‍♂️

PR #3417, opened 2024-06-09
fix: Use EXPANDED_CODE_SIGN_IDENTITY in codesign_sparkle_embedded_apps.sh

delaneyb:
Updated the `codesign_sparkle_embedded_apps.sh` script to use `$EXPANDED_CODE_SIGN_IDENTITY` instead of `$CODE_SIGN_IDENTITY`. This prevents "Apple Development: ambiguous" errors when there are multiple matching code-signing identities from XCode with automatic code-signing enabled, ensuring the correct identity is used for signing the Sparkle embedded app AutoUpdate.

Example comparison of the envvar values. These can be observed by enabling the "show environment variables in build log" checkbox under the Run Script build phase for codesign_sparkle_embedded_apps.sh, and then checking the build log:
```
    export CODE_SIGN_IDENTITY\=Apple\ Development
    export CODE_SIGN_IDENTITY_NO\=Apple\ Development
    export EXPANDED_CODE_SIGN_IDENTITY\=2FCF2E9C47BBF33DA64E1BXC648A32A6D7C6F314
    export EXPANDED_CODE_SIGN_IDENTITY_NAME\=Apple\ Development:\ example@me.com\ \(7X2KK4Z37L\)
```

image-20240208144431245

### DIFF
--- a/scripts/codesign_sparkle_embedded_apps.sh
+++ b/scripts/codesign_sparkle_embedded_apps.sh
@@ -8,4 +8,4 @@ set -exu
 # codesign --deep is only 1 level deep. It misses Sparkle embedded app AutoUpdate
 # this build phase script works around the issue
 
-codesign --verbose --force --sign "$CODE_SIGN_IDENTITY" $OTHER_CODE_SIGN_FLAGS "${CODESIGNING_FOLDER_PATH}/Contents/Frameworks/Sparkle.framework/Versions/A/Resources/Autoupdate.app"
+codesign --verbose --force --sign "$EXPANDED_CODE_SIGN_IDENTITY" $OTHER_CODE_SIGN_FLAGS "${CODESIGNING_FOLDER_PATH}/Contents/Frameworks/Sparkle.framework/Versions/A/Resources/Autoupdate.app"


lwouis:
Hi,

On this project, we manually set variables in .xcconfig files, such as [this one](https://github.com/lwouis/alt-tab-macos/blob/master/config/debug.xcconfig). This is a way to have configuration on git, and avoid magic setup on XCode, for every contributor.

I'm not sure how this PR plays into this. I'm wondering: couldn't you just edit `debug.xcconfig` or `release.xcconfig` with the correct value for your case? Wouldn't that do it for you?

Thank you 🙇

delaneyb:
Hi @lwouis, big thanks for your work on this project firstly, real productivity game-changer and great codebase to explore to get into swift :)

Regarding the proposed change, even with the code signing identity specified in the xcconfig files, the CODE_SIGN_IDENTITY environment variable still results in the vague --sign "Apple Development" being passed to the codesign command in the codesign_sparkle_embedded_apps.sh script. If a developer has both an "Apple Development" personal team certificate and a company team certificate with the same name, it will lead to the ambiguous certificate error.

lwouis:
Thank you for your message

I'm confused how the ID could be `Apple Development `. In my case, it's `Developer ID Application: Louis Pontoise (QXD7GW8FHY)`. It seems to contain a unique ID. Is it just my setup? Could you not write a specific/qualified string for the `CODE_SIGN_IDENTITY` variable, and have that be unambiguous when passed to `codesign`?

Thank you

delaneyb:
Hi @lwouis,

Apologize for the confusion. I realized that the CODE_SIGN_IDENTITY specified in Debug.xcconfig was getting overridden by my changes in the Xcode signing & capabilities tab (which affect the .pbxproj file). These changes in Xcode seem to take precedence over the configuration from the .xcconfig files.

When I followed the instructions and properly set up CODE_SIGN_IDENTITY in Debug.xcconfig using scripts/codesign/setup_local.sh, the exact code signing identity from the xcconfig gets passed through as-is to scripts/codesign_sparkle_embedded_apps.sh. This resolves the ambiguous certificate issue.

Thank you for pointing me in the right direction. I'll close this PR as the proposed change is not necessary when the .xcconfig files are set up correctly.

Issue #2892, opened 2023-10-17
Sometimes, the colored icons on top of each window are translucid


liuzongyang255:
**Describe the bug**

Sometimes the icons blend into the background

Sometimes icons have transparency

**Screenshots / video**

<img width="670" alt="image" src="https://github.com/lwouis/alt-tab-macos/assets/54257047/58e68d29-430d-4662-b339-60ef17a41c15">

<img width="463" alt="image" src="https://github.com/lwouis/alt-tab-macos/assets/54257047/9f20f29e-b350-457a-b4cb-e922a5bf1992">

**Steps to reproduce the bug**

 just use the shortcut keys to enter

**Your environment**

* AltTab version: 6.62.0
* macOS version:  14.1 Beta(23B5067a)
* Other relevant info: e.g. using external keyboard, multiple monitors, etc

liuzongyang255:
sure!
<img width="254" alt="image" src="https://github.com/lwouis/alt-tab-macos/assets/54257047/34dc77bb-b593-4b6d-a6b7-30956f93a05a">


![Oct-19-2023 17-33-40](https://github.com/lwouis/alt-tab-macos/assets/54257047/26a39262-fbbc-4ba4-86d2-fb4b83a74a86)

lwouis:
Hi,

I can't reproduce the issue. Looking at the code, I don't see how this could be a bug on AltTab's side.

Could it be because you're in a macOS beta? Have you witnessed this issue on the public release of macOS, in the past, or on another machine?

Thank you 🙇

liuzongyang255:
Hello, the same applies to me after reinstalling the latest version of AltTab. I used to use Mac Ventura 13.5 without any issues, possibly due to issues with the beta version. Thank you for your patience in answering the questions

Allsochen:
I reproduce the issue again at the dark theme.
<img width="715" alt="image" src="https://github.com/lwouis/alt-tab-macos/assets/5805484/7a5becf4-da1d-4549-b522-46c80abf9376">
<img width="485" alt="image" src="https://github.com/lwouis/alt-tab-macos/assets/5805484/dabc8fa6-16a5-4bc9-8357-498056fc8d4b">
<img width="1659" alt="image" src="https://github.com/lwouis/alt-tab-macos/assets/5805484/1455ed1d-0da5-406d-9f2a-dda09cc0ac41">
[...]

PR #3433, opened 2024-06-14
fix: fixed the issue where the traffic light button on the white thumbnail is difficult to see in dark mode.

Allsochen:
Fixed the issue where the traffic light button on the white thumbnail is difficult to see in dark mode.

Fixed the issues https://github.com/lwouis/alt-tab-macos/issues/2892

In dark mode, there are two situations where it is difficult to distinguish:
- The thumbnail is gray, and the traffic light button appears faded.
- The thumbnail is white, and the traffic light button is not visible at all.

image

image

After fixed:
- Fixed the transparent icon
image
image

- Fix the fullscreen/defullsreen icon
image
image


Allsochen:
> It seems it could be done more directly by forcing the appearance of `TrafficLightButton` to aqua.

Thank you for the reminder. I re-implemented it on TrafficLightButton, but encountered an issue: adding a background to TrafficLightButton didn't work. Ultimately, changing the inheritance of TrafficLightButton to NSView solved the problem.

Allsochen:
> I'm not sure I follow you, but I was thinking of something like this:
> 
> ```swift
> class TrafficLightButton: NSButton {
>     [...]
>     init(_ type: TrafficLightButtonType, _ tooltip: String, _ size: CGFloat) {
>         [...]
>         appearance = .init(named: .aqua)
>     }
>     [...]
> }
> ```

I don't know how to fix it using `aqua`, but I fixed it another way by extending NSView.

src/logic/Preferences.swift
```diff
@@ -104,8 +104,8 @@ class Preferences {
     // not exposed as preferences now but may be in the future, probably through macro preferences
     static var fontColor: NSColor { .white }
     static var windowPadding: CGFloat { 18 }
-    static var interCellPadding: CGFloat { 5 }
-    static var intraCellPadding: CGFloat { 5 }
+    static var interCellPadding: CGFloat { 15 }
+    static var intraCellPadding: CGFloat { 8 }
```
lwouis, line 108: I'm not sure to understand this change. Is it related to the goal of the MR?

Allsochen:
> Hi everyone,
> 
> Indeed, I think this MR is quite large, for its intended goal. I was able to reproduce the issue by switching to Dark Mode. Then I applied [the fix suggested by @decodism](https://github.com/lwouis/alt-tab-macos/pull/3433#issuecomment-2171770891), and it fixes the issue for me. One line of code seems to fix the issue:
> 
> image

This MR not only fixes the issue with the transparency of the traffic lights but also does the following:

- Fixes the shape of fullscreen/defullscreen buttons.
- Adds code comments to make it easier to understand.

src/logic/Preferences.swift
Allsochen: This fix is unrelated to the changes here. The modification was made because many AltTab users around me have reported that the spacing between thumbnails is too small, so I changed it to the thumbnail spacing of Windows 11. If you disagree, we can leave this part unchanged for now and consider revising it later as a separate advanced configuration option.

Allsochen:
> Hi everyone,
> 
> Indeed, I think this MR is quite large, for its intended goal. I was able to reproduce the issue by switching to Dark Mode. Then I applied [the fix suggested by @decodism](https://github.com/lwouis/alt-tab-macos/pull/3433#issuecomment-2171770891), and it fixes the issue for me. One line of code seems to fix the issue:
> 

This MR not only fixes the issue with the transparency of the traffic lights button but also does the following:

Fixes the shape of fullscreen/defullscreen buttons.
Adds code comments to make it easier to understand.

src/logic/Preferences.swift
Allsochen: 83ab94fdb3b446c1e98e5f63bc2f60ef

Changed the thumbnail spacing equals to Windows 11 spacing.

src/ui/generic-components/TrafficLightButton.swift
```diff
@@ -1,219 +1,379 @@
...
-    }
-
+    /// Mouse entered event handler.
     override func mouseEntered(with event: NSEvent) {
         isMouseOver = true
-        setNeedsDisplay()
+        setNeedsDisplay(bounds)
     }
 
+    /// Mouse exited event handler.
```
lwouis, line 50: I don't think these comments are the way to go. When we see a method whith `override`, we can use our IDE to display the original method's documentation, very easily:

image

src/ui/generic-components/TrafficLightButton.swift
```diff
@@ -1,219 +1,379 @@
...
+
+        let disk = NSBezierPath()
+        disk.appendOval(in: NSMakeRect(bounds.origin.x + 0.5, bounds.origin.y + 0.5, bounds.width - 1, bounds.height - 1))
+
+        // Fill the disk with the black background color
+        NSColor.black.setFill()
+        disk.fill()
+
+        // Fill the disk with the background color
+        diskBackgroundColor.draw(in: disk, relativeCenterPosition: .zero)
```
lwouis, line 97: It seems you added drawing a black disk here. What issue is it fixing?

src/ui/generic-components/TrafficLightButton.swift
```diff
@@ -1,219 +1,379 @@
...
+    /// - Parameters:
+    ///   - disk: The `NSBezierPath` representing the button's disk.
     private func drawDimming(_ disk: NSBezierPath) {
+        // Set the line width of the disk
         disk.lineWidth = 1
-        if (isHighlighted) {
-            NSColor.black.withAlphaComponent(0.5).setFill()
-            disk.fill()
-        } else if (isMouseOver) {
+        if isMouseOver {
```
lwouis, line 113: You removed the difference of being highlighted = 50% opacity, and being hovered by the mouse = 25% opacity. Now the mouse hover is 25%, and the highlight is gone; it's always 100%. I'm not sure why this change is made. How is it related to the stated goal of the MR?

src/ui/generic-components/TrafficLightButton.swift
```diff
@@ -1,219 +1,379 @@
...
-            window_?.toggleFullscreen()
-        } else if (type == .miniaturize) {
-            window_?.minDemin()
-        } else if (type == .close) {
-            window_?.close()
-        } else if (type == .quit) {
-            window_?.application.quit()
-        }
-    }
-
```
lwouis, line 34: Was this really not necessary? I'm thinking that if it's there, there may be a reason. Have you tested thoroughly? Also checking the git history to see when and why this was added?

src/ui/generic-components/TrafficLightButton.swift
```diff
@@ -1,219 +1,379 @@
...
+    /// Mouse entered event handler.
     override func mouseEntered(with event: NSEvent) {
         isMouseOver = true
-        setNeedsDisplay()
+        setNeedsDisplay(bounds)
```
lwouis, line 47: What is the purpose of this change?

src/ui/main-window/ThumbnailView.swift
```diff
@@ -95,6 +95,8 @@ class ThumbnailView: NSStackView {
                         yOffset += ThumbnailView.windowsControlSize + ThumbnailView.windowsControlSpacing
                     }
                 }
+                // Force the icon to repaint
+                icon.display()
```
lwouis, line 99: Why is this necessary? What was failing before we added this new call?

src/logic/Preferences.swift
lwouis: I think changes shouldn't be bundled. It can create confusion while reviewing MRs. If you propose to make multiple unrelated changes, I suggest opening multiple MRs, each with their goal. They can be reviewed with more clarity.

Furthermore, we use a convention on commits. You can learn [a bit more here](https://alt-tab-macos.netlify.app/contributing#raising-a-pull-request). One commit = 1 change. Here there should be a `fix:` for the traffic light opacity issue, since it's indeed a bug fix. Then there should be a `feat:` to update the spacings, since it's not a bug fix bug an improvement. Finally, for documentation, one could use `docs:` or `refactor:` for the commit which documents the existing code.

Regarding the change of spacing: I think Windows 11 Spacings should live in a Windows 11 theme. We already support Themes in Preferences > Appearance. Let's add a Windows 11 theme. Other people were asking about it I believe.

src/ui/main-window/ThumbnailView.swift
Allsochen: If this modification is not added, an issue arises: after the user clik the fullscrren button to enters full screen, the fullscreen button does not redraw, resulting in the user still seeing the button as if it were in fullscreen mode, rather than the defullscreen button. To make this issue clearer to you, I will open a new MR, so you can see the problem and the solution more clearly.

src/ui/generic-components/TrafficLightButton.swift
Allsochen: In terms of the effect, their performance is the same. This class inherits from the NSView class, and through this modification, we can specify a fixed area for redrawing, which will be more precise.

src/ui/generic-components/TrafficLightButton.swift
Allsochen: The button is currently an NSView, and when it needs to be displayed, it should show its color at 100% opacity. When the mouse hovers over it, it should display at 25% opacity. The NSButton has an isHighlighted state, but this state is no longer needed with the current setup.

src/ui/generic-components/TrafficLightButton.swift
Allsochen: This is the key to fixing the issue of transparent button in the dark theme. The black disk serves as a background to prevent the buttons from being transparent.

src/ui/generic-components/TrafficLightButton.swift
Allsochen: Thank you for the reminder, this can be changed to mouseDown.

lwouis:
I'll grab the change on the `dev` branch. It will be in the next release 👍

PR #3454, opened 2024-06-27
fix: fix the the shape of fullscreen/defullscreen traffic light button

Allsochen:
Fix the the shape of fullscreen/defullscreen traffic light button.

before fixed:
- fullscreen
b0f0e599e5809beac06e90d8134c61fb

- defullscreen
b0f0e599e5809beac06e90d8134c61fb

after fixed:
- fullscreen
b43f7b01ab3e71a235c73710e6799051

- defullscreen
925086e1a44fc40c498d9af47f119d10

### DIFF
--- a/src/ui/generic-components/TrafficLightButton.swift
+++ b/src/ui/generic-components/TrafficLightButton.swift
@@ -117,32 +117,93 @@ class TrafficLightButton: NSButton {
     }
 
     private func drawSymbol(_ lineColor: NSColor) {
+        let symbol = NSBezierPath()
         if (type == .fullscreen) {
-            let symbol = NSBezierPath()
-            symbol.move(to: NSMakePoint(bounds.width * 0.25, bounds.height * 0.75))
-            symbol.line(to: NSMakePoint(bounds.width * 0.25, bounds.height * 1 / 3))
-            symbol.line(to: NSMakePoint(bounds.width * 2 / 3, bounds.height * 0.75))
-            symbol.close()
-            lineColor.setFill()
-            symbol.fill()
-            symbol.move(to: NSMakePoint(bounds.width * 0.75, bounds.height * 0.25))
-            symbol.line(to: NSMakePoint(bounds.width * 0.75, bounds.height * 2 / 3))
-            symbol.line(to: NSMakePoint(bounds.width * 1 / 3, bounds.height * 0.25))
-            symbol.close()
-            lineColor.setFill()
-            symbol.fill()
-            // maximize cross
-            // NSGraphicsContext.current?.shouldAntialias = false
-            // var symbol = NSBezierPath()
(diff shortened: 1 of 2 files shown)


Allsochen:
The sketch prototype is as follows.
0c978a994a107f08c5ebee016e2d2173

src/ui/generic-components/TrafficLightButton.swift
```diff
@@ -117,40 +117,55 @@ class TrafficLightButton: NSButton {
...
 
     private func drawSymbol(_ lineColor: NSColor) {
+        let symbol = NSBezierPath()
         if (type == .fullscreen) {
-            let symbol = NSBezierPath()
-            symbol.move(to: NSMakePoint(bounds.width * 0.25, bounds.height * 0.75))
-            symbol.line(to: NSMakePoint(bounds.width * 0.25, bounds.height * 1 / 3))
-            symbol.line(to: NSMakePoint(bounds.width * 2 / 3, bounds.height * 0.75))
+            var firstPointA, firstPointB, firstPointC, secondPointA, secondPointB, secondPointC: NSPoint!
+            if window_?.isFullscreen ?? true {
```
lwouis, line 123: I think it may be safer to default to the fullscreen symbol if we lost the `window_` reference here.

```swift
window_?.isFullscreen ?? false
```

src/ui/main-window/ThumbnailView.swift
```diff
@@ -17,7 +17,7 @@ class ThumbnailView: NSStackView {
     var quitIcon = TrafficLightButton(.quit, NSLocalizedString("Quit app", comment: ""), windowsControlSize)
     var closeIcon = TrafficLightButton(.close, NSLocalizedString("Close window", comment: ""), windowsControlSize)
     var minimizeIcon = TrafficLightButton(.miniaturize, NSLocalizedString("Minimize/Deminimize window", comment: ""), windowsControlSize)
-    var maximizeIcon = TrafficLightButton(.fullscreen, NSLocalizedString("Fullscreen window", comment: ""), windowsControlSize)
+    var maximizeIcon = TrafficLightButton(.fullscreen, NSLocalizedString("Fullscreen/Defullscreen window", comment: ""), windowsControlSize)
```
lwouis, line 20: I think this change is good. However, I would like to point out that this will be quite impactful on the localizations.  The app is translated in 50 languages, and this little change means that at least 50 people will have to come in and provide a translation.

It's probably still right to make this change. I just wanted to share the full picture of the consequences with you

lwouis:
I'll take this commit on the `dev` branch. It will be released with the next release 👍

src/ui/main-window/ThumbnailView.swift
lwouis: Actually, this string already exists in `ControlsTab.swift`. So you fixed the usage without any impact on the localization side 👍

PR #3470, opened 2024-07-07
feat: add function key modifier

decodism:
It's a bit hacky as Carbon doesn't support it.

### DIFF
--- a/Pods/ShortcutRecorder/Library/SRCommon.h
+++ b/Pods/ShortcutRecorder/Library/SRCommon.h
@@ -10,18 +10,21 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+static const UInt32 fakeFunctionKey = 1 << 16;
+
+
 /*!
  Mask representing subset of Cocoa modifier flags suitable for shortcuts.
  */
 NS_SWIFT_NAME(CocoaModifierFlagsMask)
-static const NSEventModifierFlags SRCocoaModifierFlagsMask = NSEventModifierFlagCommand | NSEventModifierFlagOption | NSEventModifierFlagShift | NSEventModifierFlagControl;
+static const NSEventModifierFlags SRCocoaModifierFlagsMask = NSEventModifierFlagCommand | NSEventModifierFlagOption | NSEventModifierFlagShift | NSEventModifierFlagControl | NSEventModifierFlagFunction;
 
 
 /*!
  Mask representing subset of Carbon modifier flags suitable for shortcuts.
  */
 NS_SWIFT_NAME(CarbonModifierFlagsMask)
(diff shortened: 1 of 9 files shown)


lwouis:
Hi @decodism,

Thank you for sharing this work.

Adding the function key was discussed in the past. In end, I concluded that opening that door would create support situation that would not be worth. Have you seen [this thread](https://github.com/Kentzo/ShortcutRecorder/issues/132#issuecomment-717471192) perhaps?

It's pretty cool that you managed to implement that! I need to test it to see how it behaves.

As I mentioned above, I'm not sure it's worth to add this feature. It would probably be of niche use, and open the door to all sorts of exotic bugs and behaviors.

What do you think?

Thank you 🙇

lwouis:
I played around with the PR. Here's some feedback, in addition to my reserves above about opening that door:

* I think the symbol `F` may be a bit confusing. Perhaps some `fn` in small letters would be clearer. It's a consistent wording on the physical keyboards
* The app crashed when I tried to use a shortcut with F1. I held `fn` then the f-keys appeared, and I pressed F1. I see in the debugger that in ControlsTabs.swift, line 194,`Shortcut(keyEquivalent: newValue)` was `nil`. `newValue` was equal to `"⌥FF1"`. It raises the question: how should the `fn` key interact with f-keys like F1. Should pressing `fn` then `F1` count as F1 only, or as modifier `fn` + `F1` key?

Thank you

lwouis:
Hi @decodism,

Thank you for the updates!

Regarding my concerns:

As you've said, it's not supported by Carbon. This means it's exotic and other apps supporting it will be few. Users probably are not used to it or expecting it to work, or the nuances which go with it. For instance, the cases with fn+F1 I've shared previously, or the fact that [Apple own UI in System Settings is broken currently](https://github.com/Kentzo/ShortcutRecorder/issues/132#issuecomment-718041180) when assigning shortcuts with `fn`

The `fn` key is used together with the globe key. I think this also adds lots of confusion to supporting this feature. For instance, my setup is an external Logitech K380 keyboard, and a 2022 macbook pro. Here's how they physically look like:

Logitech K380 | 2022 macbook pro
-|-
IMG_2324 Large | IMG_2323 Large

Notice how the System Settings mention the `fn` key for the external keyboard, and the `globe` key for the internal keyboard. Even though both keys features the `fn` marking on it:

Logitech K380 | 2022 macbook pro
-|-
image | image

I can set `fn` using the internal keyboard, with your PR, but I can't set `fn` with the external keyboard. There must be an implementation difference. Maybe you catch the `globe` key code instead of `fn`.

Do you see what I mean by adding complexity and exotic use-case if we merge this?

What do you think?

Thank you

PR #3468, opened 2024-07-06
feat: refactor the appearance preferences for simplification.

Allsochen:
Refactor the appearance preferences for simplification. There has been a lot of discussion about the issue of complex appearance preferences. You can check out this [post](https://github.com/lwouis/alt-tab-macos/issues/351#issuecomment-1317633417).

* All complex UI appearances will be replaced by the appearance model and appearance size. 
* Options will be categorized under various tabs or advanced button.
* Support white and black themes, and can automatically switch according to the system theme.
* Support displaying the title under the icon in app icons mode.
* Use new style settings UI like syteme settings at macOS.

Here is the snashoot of preferences:
QQ_1723391590487
QQ_1723391641594
QQ_1723391654475
QQ_1723391673548
QQ_1723391689021
QQ_1723391711666
QQ_1723391724770
QQ_1723391738837

Here is the snapshot of UI:
* Thumbnail Light theme
QQ_1723392072049

* Thumbnail Dark theme
QQ_1723392143302

* App Icons Light theme
QQ_1723392209339

* App Icons Dark theme
QQ_1723392169062

* Titles
QQ_1723392285814

PR #3481, opened 2024-07-11
feat: limit windows per app

decodism:
It could be useful to limit the number of windows in a shortcut with all the apps. A shortcut with the active app can still be used if necessary.

### DIFF
--- a/src/logic/Applications.swift
+++ b/src/logic/Applications.swift
@@ -59,7 +59,7 @@ class Applications {
             // comparing pid here can fail here, as it can be already nil; we use isEqual here to avoid the issue
             Applications.list.removeAll { $0.runningApplication.isEqual(runningApp) }
             Windows.list.enumerated().forEach { (index, window) in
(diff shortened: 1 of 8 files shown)


lwouis:
Hi,

I think this PR overlaps a bit with #337. I think it would appeal to user who want some kind of group-per-app, hierarchy, or drilldown capability. Of course it's more crude since it's truncating data. So it would help only a subset of the people looking for #337.

As it stands, I think it adds to the issues with #351. I think it's interesting, but not comprehensive enough to warrant adding more settings, as it stands. @Allsochen is currently working on #351. Do you have an opinion on this perhaps @Allsochen?

Thank you

lwouis:
> This could perhaps be a stopgap solution before what's planned.

@Allsochen is working on https://github.com/lwouis/alt-tab-macos/pull/3468 currently. I think the long-term #337 or a stop-gap workaround should be thought in the context of this work. In another words, if we would merge MR this today, we would need to add migrating it to the new preferences UI in https://github.com/lwouis/alt-tab-macos/pull/3468

PR #3482, opened 2024-07-11
fix: compilation

decodism:
Xcode compilation issue: `Negative integer '-1' overflows when stored into unsigned type 'UInt32'`

### DIFF
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -85,7 +85,7 @@ class Window {
     func isEqualRobust(_ otherWindowAxUiElement: AXUIElement, _ otherWindowWid: CGWindowID?) -> Bool {
         // the window can be deallocated by the OS, in which case its `CGWindowID` will be `-1`
         // we check for equality both on the AXUIElement, and the CGWindowID, in order to catch all scenarios
-        return otherWindowAxUiElement == axUiElement || (cgWindowId != nil && Int(cgWindowId!) != -1 && otherWindowWid == cgWindowId)
+        return otherWindowAxUiElement == axUiElement || (cgWindowId != nil && cgWindowId != CGWindowID(bitPattern: -1) && otherWindowWid == cgWindowId)
     }
 
     private func observeEvents() {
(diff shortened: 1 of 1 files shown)


src/logic/Window.swift
```diff
@@ -101,7 +101,7 @@ class Window {
     }
 
     private func screenshot(_ bestResolution: Bool = false) -> NSImage? {
-        guard !isWindowlessApp, let cgWindowId = cgWindowId, cgWindowId != -1, let cgImage = cgWindowId.screenshot(bestResolution) else {
+        guard !isWindowlessApp, let cgWindowId = cgWindowId, cgWindowId != kCGNullWindowID, let cgImage = cgWindowId.screenshot(bestResolution) else {
```
lwouis, line 104: In theory, `CGWindowID` being unsigned, wid can never be negative. Thus comparing with `-1` was an actual bug, not a false positive error at compilation.

I'm wondering if comparing with `kCGNullWindowID` is ok. We set the wid to `0`, then get the actual wid from the OS API:

https://github.com/lwouis/alt-tab-macos/blob/012e3b5f3c321b249271420fffb3d7ee561b9560/src/api-wrappers/AXUIElement.swift#L54

Is there no way that the private API `_AXUIElementGetWindow` wouldn't set the ID, thus keeping it at `0`? We would not screenshot it then.

It's probably fine? What do you think?

Thank you

lwouis:
> I removed the check as it should be unnecessary. There was also a strange code comment.

This is coming back to me, I think. If I remember correctly, I came across this scenario in the past while debugging: because of concurrency, it's possible that the window has been close, and is in the process of being destroyed by the OS. I think that's when the wID can become -1 temporarily. I think that's why I had added the checks to short-circuit and avoid wasting time if the value is -1.

I think it's fine to remove this check since the logic is preserved by `otherWindowWid == cgWindowId`. However, for the screenshot conditional, I'm not sure if it's ok to remove. Would it cause problems if the wID was actually `-1` there?

Thank you

lwouis:
> But a `CGWindowID` can never contain `-1`?

These types are indicative at compile-time. I've seen other cases where runtime values are not matching the declared types. Especially when working in Swift on top of the older layers, the types can be flat out wrong. I remember one API had a type `Something!`. I had caught some real-life cases where the value was `nil`. Someone had added the exclamation mark in the SDK but it was actually optional at runtime.

lwouis:
Thank you @decodism. I've cherry-picked the commit and push it to master for the next release. I'll close this PR now 👍

PR #3488, opened 2024-07-14
fix: adjust Quit button position based on system language direction

### LINKED ISSUE #3487: The Quit button is not accessible when the system language is RTL
**Describe the bug**

When the system language of macOS is RTL language (Hebrew/Arabic), macOS places its traffic light buttons on the right side of the window instead of on the left, which causes a lack of access to the Quit button (see screenshot)
Note that what determines is the system language and not the specific language for the application, even if one is selected in System Settings (it doesn't make much sense on Apple's part, but that's how it works).

**Screenshots / video**

CleanShot 2024-07-14 at 02 47 23@2x
CleanShot 2024-07-14 at 02 39 15@2x

1. Go to System Settings > Language & Region
2. Change the language to hebrew
3. Open alt-tab settings

**Your environment**

* AltTab version: 6.71.0
* macOS version: 14.5

ShlomoCode:
These changes should work, although I haven't tested them locally as I couldn't build the project - I get an error:
```
No such module 'ShortcutRecorder'
```
CleanShot 2024-07-14 at 03 53 42@2x

closes #3487

### DIFF
--- a/src/ui/preferences-window/PreferencesWindow.swift
+++ b/src/ui/preferences-window/PreferencesWindow.swift
@@ -23,7 +23,11 @@ class PreferencesWindow: NSWindow, NSToolbarDelegate {
         titleBarView.addSubview(quitButton)
         quitButton.translatesAutoresizingMaskIntoConstraints = false
         quitButton.topAnchor.constraint(equalTo: titleBarView.topAnchor, constant: 5).isActive = true
-        quitButton.rightAnchor.constraint(equalTo: titleBarView.rightAnchor, constant: -8).isActive = true
+        if self.windowTitlebarLayoutDirection == .rightToLeft {
+            quitButton.leftAnchor.constraint(equalTo: titleBarView.leftAnchor, constant: 8).isActive = true
+        } else {
+            quitButton.rightAnchor.constraint(equalTo: titleBarView.rightAnchor, constant: -8).isActive = true
+        }
     }
 
     private func setupView() {


src/ui/preferences-window/PreferencesWindow.swift
```diff
@@ -23,7 +23,17 @@ class PreferencesWindow: NSWindow, NSToolbarDelegate {
...
         quitButton.translatesAutoresizingMaskIntoConstraints = false
         quitButton.topAnchor.constraint(equalTo: titleBarView.topAnchor, constant: 5).isActive = true
-        quitButton.rightAnchor.constraint(equalTo: titleBarView.rightAnchor, constant: -8).isActive = true
+        
+        // Check the system language direction, not the application's language.
+        // macOS determines the position of the traffic light buttons based on the system language, not the application-specific language.
+        // See also https://stackoverflow.com/a/45299393
+        let systemLanguage = NSLocale.preferredLanguages.first ?? "en"
+        let systemLangIsRtl = NSLocale.characterDirection(forLanguage: systemLanguage) == .rightToLeft
+        if systemLangIsRtl {
```
lwouis, line 32: I think this could be simplified. Instead of looking for device language and dealing with fallbacks and all. I think we can simply ask the preference window view if it is currently right-to-left as such:

```swift
contentView!.userInterfaceLayoutDirection == .rightToLeft
```

ShlomoCode: macOS determines the location of the traffic light buttons based on the system language, not the application-specific language, so if the system language is LTR, but an application-specific RTL language is set (or vice versa) the detection will be incorrect, right?

ShlomoCode: Ref: https://github.com/iina/iina/issues/4910

lwouis: I think I understand the distinction you are making between the system-wide language, and the app language.

My point was: can't we simply tap downstream, and check if the view is drawn RTL?

I think you're correct that the view could be RTL while its parent window is LTR. I've found this other API here which looks like it would work:

image

[windowTitlebarLayoutDirection](https://developer.apple.com/documentation/appkit/nswindow/1644535-windowtitlebarlayoutdirection) also mentions [NSUserInterfaceLayoutDirection.rightToLeft](https://developer.apple.com/documentation/appkit/nsuserinterfacelayoutdirection/righttoleft)

This line from the docs seem to confirm that we should be using this:

> Refer to this value if the application uses [titlebarAppearsTransparent](https://developer.apple.com/documentation/appkit/nswindow/1419167-titlebarappearstransparent) and places controls under the title bar.

What do you think?

ShlomoCode: > My point was: can't we simply tap downstream, and check if the view is drawn RTL?

I think not. Logic says that if the window is drawn RTL, the window buttons will also appear in the correct place, but macOS does not do this.
You can check it yourself simply by entering the system settings and [changing the application language](https://support.apple.com/en-il/guide/mac-help/mh26684/mac#:~:text=Choose%20the%20language%20you%20use%20for%20individual%20apps) only to Hebrew, while leaving your system language on Hebrew (don't do it through XCode's `Right-to-Left Pseudolanguage`, it doesn't reflect reality properly). You can see that the application itself is drawn in RTL, but the traffic light buttons remain on the left side of the window. On the other hand, if you also change the system language to Hebrew and restart, the traffic light buttons will also move to the right side.

Regarding the alternative API you suggested, I can't check because, as mentioned, I can't build the project. Can you help me solve the strange module error?

ShlomoCode: The docs you referred to does seem promising. It's worth checking how it works in practice!

lwouis: Could you please try the [instructions there](https://alt-tab-macos.netlify.app/contributing#building-the-project-locally)? I see on your screenshot that you're not using a local self-signed certificate. I hope this can fix it.

I'm not sure why XCode complains about `ShortcutRecorder`. It's a dependency, and it's stored on git, so you don't need to fetch it separately with cocoapods. It should just work out of a git clone. Have you opened the project via `alt-tab-macos.xcworkspace`? It shouldn't be opened via `alt-tab-macos.xcodeproj`. Perhaps that's the issue?

Thank you

ShlomoCode: > Have you opened the project via `alt-tab-macos.xcworkspace`? It shouldn't be opened via `alt-tab-macos.xcodeproj`. Perhaps that's the issue?

Yes! that's it. Thanks, I'll check the API

ShlomoCode: BTW, the `setup_local.sh` script did not work as it is, there was an error:
```
➜  alt-tab-macos git:(fix/quit-button-rtl-position) ✗ scripts/codesign/setup_local.sh
+ certificateFile=codesign
++ openssl rand -base64 12
+ certificatePassword=**********
+ scripts/codesign/generate_selfsigned_certificate.sh codesign **********
+ certificateFile=codesign
+ certificatePassword=**********
+ cat
+ openssl genrsa -out codesign.key 2048
+ openssl req -x509 -new -config codesign.conf -nodes -key codesign.key -extensions extensions -sha256 -out codesign.crt
+ openssl pkcs12 -export -inkey codesign.key -in codesign.crt -out codesign.p12 -passout pass:**********
+ scripts/codesign/import_certificate_into_main_keychain.sh codesign **********
+ certificateFile=codesign
+ certificatePassword=**********
+ security import codesign.p12 -P ********** -T /usr/bin/codesign
security: SecKeychainItemImport: MAC verification failed during PKCS12 import (wrong password?)
```
I solved it by [adding the `-legacy` flag](https://stackoverflow.com/a/74792849) to the openssl call at the end of the `scripts/codesign/generate_selfsigned_certificate.sh` file

ShlomoCode: I tested `windowTitlebarLayoutDirection` in the following scenarios:
RTL system, RTL application
RTL system, LTR application
LTR system, LTR application
LTR system, RTL application

It seems to work well, I force-pushed it. Thanks!

lwouis:
Thank you!

I'll cherry-pick this into the next release. I'll close this PR now.

You'll be credited on the website and in the app itself 👍 

Thank you 🙇

Issue #3477, opened 2024-07-10
On Sequoia beta 3 (Mac os 15.0), it keeps says screen recording is not allowed even it's allowed , and don't have popup like before


hyungchulc:
**Describe the bug**

On Sequoia beta 3 (Mac Os 15.0), it keeps says screen recording is not allowed even it's allowed , and don't have popup like before, which asks if I want to keep give the permission.

**Screenshots / video**

![Screenshot 2024-07-11 at 03 21 59 AM@2x](https://github.com/lwouis/alt-tab-macos/assets/64293163/194dd77a-d207-4df4-a94d-ff3dcfd64df5)

**Steps to reproduce the bug**

1. Open AltTab in beta 3
2. Doesn't work

**Your environment**

* AltTab version: Latest, 6.71.0
* macOS version: 15.0

falchion10:
Having this issue as well. Just updated to 15.0 Developer Beta 3

UnknownCrafts:
I would like to confirm this issue and also give a workaround. Unfortunately the workaround does not persist during restarts but it is still better than being able to use AltTab, ofc you can redo the steps after each restart.

Workaround:

https://github.com/lwouis/alt-tab-macos/assets/65975222/2808010e-d600-4f00-be03-82b080629368

UnknownCrafts:
> > Edit: just for extra clarification I am also on MacOS 15 dev beta 3
> > I would like to confirm this issue and also give a workaround. Unfortunately the workaround does not persist during restarts but it is still better than not being able to use AltTab, ofc you can redo the steps after each restart.
> > Workaround:
> > video.mp4
> 
> Removing + adding back didn't fix it for me. I had to add it without removing it and it fixed itself. Seems more like a MacOS issue though

I will add that as an alternative to the original comment.

Ethansev:
> > Edit: just for extra clarification I am also on MacOS 15 dev beta 3
> > I would like to confirm this issue and also give a workaround. Unfortunately the workaround does not persist during restarts but it is still better than not being able to use AltTab, ofc you can redo the steps after each restart.
> > Workaround:
> > video.mp4
> 
> Removing + adding back didn't fix it for me. I had to add it without removing it and it fixed itself. Seems more like a MacOS issue though

Not working for me unfortunately but worth a shot. I'll play with it a bit and see what I can do. Also on MacOS 15 dev beta 3.

hyungchulc:
[...]

Issue #122, opened 2020-01-22
Display video of the windows instead of static screenshots


lwouis:
Since v2, we have been using the `CGSHWCaptureWindowList` api to capture window images. I tried to refresh them on a timer at screen refresh-rate to simulate video. My POC worked but performance was an issue, when going beyond 30 fps, it becomes laggy and 60fps is not achieved.

I noticed the introduction of a [new private API in 10.15: SLSHWCaptureStreamCreateWithWindow](https://github.com/phracker/MacOSX-SDKs/blob/ef9fe35d5691b6dd383c8c46d867a499817a01b6/MacOSX10.15.sdk/System/Library/PrivateFrameworks/SkyLight.framework/Versions/A/SkyLight.tbd#L277). The name seems to imply that it may be given us a video stream. The name having `HW` for HardWare in it also implies that it would be a high-performance stream straight from the GPU, before compositing.

@koekeishiya helped me guess the signature in an email exchange:

> From what I can tell the function you are interested in has the following signature:
>
> CGDisplayStreamRef SLSHWCaptureStreamCreateWithWindow(uint32_t cid, uint32_t wid, CFDictionaryRef display_stream_options, dispatch_queue_t queue)
>
> (CGDisplayStreamRef may not be the correct return type, but I assume it might be, as the return value is the result of something called SLDisplayStreamCreate, which I found to very likely be related to https://developer.apple.com/documentation/coregraphics/1455170-cgdisplaystreamcreate?language=objc
>
> The following is a screenshot of how this looks in IDA:
>
> ![image](https://user-images.githubusercontent.com/106195/72873085-3b410f00-3d32-11ea-9e93-7f3d3bc742ce.png)

I tried with Hopper and noticed that there is a 5th argument:

![image](https://user-images.githubusercontent.com/106195/72873132-5f9ceb80-3d32-11ea-8e08-f2e2d00effad.png)

Looking at `CGDisplayStream` docs, it seems clear that there is a need for a code block argument that is the callback that gets called on each frame of the display stream. I thus tried running the following code:

```swift
@_silgen_name("SLSHWCaptureStreamCreateWithWindow")
func SLSHWCaptureStreamCreateWithWindow(_ cid: CGSConnectionID, _ wid: CGWindowID, _ displayStreamOptions: CFDictionary, _ queue: DispatchQueue, _ block: CGDisplayStreamFrameAvailableHandler?) -> CGDisplayStream
```

```swift
[...]

PR #3497, opened 2024-07-18
Fix permissions for macos 15

DominicVonk:

### DIFF
--- a/src/logic/SystemPermissions.swift
+++ b/src/logic/SystemPermissions.swift
@@ -2,108 +2,115 @@ import Cocoa
 
 // macOS has some privacy restrictions. The user needs to grant certain permissions, app by app, in System Preferences > Security & Privacy
 class SystemPermissions {
-    static var changeCallback: (() -> Void)!
-    static var timer: Timer!
+  static var changeCallback: (() -> Void)!
+  static var timer: Timer!
 
-    static func accessibilityIsGranted() -> Bool {
-        if #available(OSX 10.9, *) {
-            return AXIsProcessTrustedWithOptions([kAXTrustedCheckOptionPrompt.takeRetainedValue(): false] as CFDictionary)
-        }
-        return true
+  static func accessibilityIsGranted() -> Bool {
+    if #available(OSX 10.9, *) {
+      return AXIsProcessTrustedWithOptions(
+        [kAXTrustedCheckOptionPrompt.takeRetainedValue(): false] as CFDictionary)
     }
-    private static func canRecordScreen() -> Bool {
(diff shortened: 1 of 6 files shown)


DominicVonk:
Fixes issue #3477

src/logic/SystemPermissions.swift
```diff
@@ -14,25 +14,11 @@ class SystemPermissions {
 
     static func screenRecordingIsGranted() -> Bool {
         if #available(OSX 10.15, *) {
-            return screenRecordingIsGranted_()
+            return CGPreflightScreenCaptureAccess()
         }
         return true
     }
 
-    // workaround: public API CGPreflightScreenCaptureAccess and private API SLSRequestScreenCaptureAccess exist, but
```
lwouis, line 22: Have you seen this? We were not using this API because it wasn't the best in the past.

DominicVonk: Yes, I read it, I will make it that it uses it for OSX 15+

lwouis: AltTab checks permissions continuously. The user could remove permissions at any point. We check for that on a repeating timer.

As the comment says, `CGPreflightScreenCaptureAccess` was returning a value which was not updated. It was a snapshot of the permissions at the time the app was launched, but didn't update after that initial snapshot.

Has this behavior changed in macOS 15? If it hasn't, then we should change other parts of the app doing the repeating timer checks. Either find another way to check there, or remove the check. But if we remove the check, we need another way to notice when the `screenshot()` and the AX calls start failing.

DominicVonk: I changed it to the canRecordScreen, this is updated during runtime

DominicVonk:
Now I only altered the `private static func screenRecordingIsGranted_()`

I think this is a stable solution

lwouis:
@DominicVonk Using `CGWindowListCopyWindowInfo` is what we used to do years ago.

Here's an [interesting thread to read](https://github.com/lwouis/alt-tab-macos/issues/704#issuecomment-762935477). The whole ticket is about issues with falsely detecting that the permission is gone, when it is in fact not gone.

The current implementation is solid. It has passed the test of time. It's now being challenged by changes in the latest macOS.

Here are some thoughts to go forwards:

* We should keep the existing code for macOS < 15. It's tested and solid.
* For macOS >= 15, I think we should look for answers in [the new API](https://github.com/lwouis/alt-tab-macos/issues/122#issuecomment-1059649102) (this thread also contains a lot worth reading).

lwouis:
I think we can close this PR? In the current state, it doesn't improve things. We still need a break through for better permissions check

PR #3554, opened 2024-08-16
fix: detect Passwords app

### LINKED ISSUE #3545: Passwords app in macOS 15 not visible
**Describe the bug**

The passwords app does not shop up  in the alt tab menu. No preview or app icon

**Screenshots / video**

Add screenshots to help explain your problem. You can also record your screen to show the bug in action. That is *really* helpful!

1. Open passwords
2. Open any other app on top
3. Alt-tab 
4. Passwords app missing
**Your environment**

* AltTab version: 6.71.0
* macOS version: 15.1

decodism:
Fixes #3545

### DIFF
--- a/src/logic/Applications.swift
+++ b/src/logic/Applications.swift
@@ -109,7 +109,7 @@ class Applications {
     private static func isActualApplication(_ app: NSRunningApplication) -> Bool {
         // an app can start with .activationPolicy == .prohibited, then transition to != .prohibited later
         // an app can be both activationPolicy == .accessory and XPC (e.g. com.apple.dock.etci)
-        return (isNotXpc(app) || isAndroidEmulator(app)) && !app.processIdentifier.isZombie()
+        return (isNotXpc(app) || isPasswords(app) || isAndroidEmulator(app)) && !app.processIdentifier.isZombie()
     }
 
     private static func isNotXpc(_ app: NSRunningApplication) -> Bool {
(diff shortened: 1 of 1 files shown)


src/logic/Applications.swift
```diff
@@ -126,6 +126,10 @@ class Applications {
     private static func notAltTab(_ app: NSRunningApplication) -> Bool {
         return app.processIdentifier != ProcessInfo.processInfo.processIdentifier
     }
+    
+    private static func isPasswords(_ app: NSRunningApplication) -> Bool {
+        return app.bundleIdentifier == "com.apple.Passwords"
```
lwouis, line 131: Hi,

Thank you for sharing this workaround.

Could you please share which checks are not passing for the Passwords app currently?

I'm a bit worried to bypass the checks like this. Maybe this app is not supposed to be monitored by accessibility apps?

Thank you 🙇

lwouis: So Passwords.app is an XPC app making windows? This seems new and surprising.

[This guide](https://multi.app/blog/how-and-when-to-use-xpc-services) says:

> XPC services have one major limitation: they do not accept user interaction. This may be a deal breaker for you. You can use an external process to show UI. The external process can have its own window hierarchy and use SwiftUI, but the user cannot interact with anything in the external process. UI elements such as buttons or sliders will not accept mouse or keyboard events.

We see here that [yabai also started filtering XPC out](https://github.com/koekeishiya/yabai/issues/529#issue-620346055). They [still have that check](https://github.com/koekeishiya/yabai/blob/136f25f515d07ff941231fa34a4fbea2448ca95e/src/process_manager.c#L40) today.

It seems that it's not even possible to use AppKit from inside an XPC app. Maybe they use tricks like have an app inside an app, and inheriting the execution context?

I'm worried to hard-code an exception like this. What do you think about an XPC app producing windows?

Thank you 🙇

lwouis:
I'll add this fix to the next release 👍

lwouis:
@AChangXD this has been [released in v6.73.0](https://alt-tab-macos.netlify.app/changelog) 👍

PR #3582, opened 2024-09-01
feat: Hiding windowless apps can be customized per shortcut

ThePixelbrain:
This PR allows to hide windowless apps from Alt Tab per shortcut instead of just having a single global option.

I have a shortcut to only show windows from visible spaces, which I find very helpful when dealing with Apples terrible keyboard handling in Mission Control. For this shortcut I didn't want to see any windowless apps, but for my main shortcut I wanted to be able to see all open apps, including the windowless ones. Alt Tab wasn't able to provide this, which motivated me to create this PR.

Users from previous versions of Alt Tab will have their global configuration applied to all shortcuts once after updating, allowing a seemless transition.

### DIFF
--- a/src/logic/Application.swift
+++ b/src/logic/Application.swift
@@ -111,8 +111,7 @@ class Application: NSObject {
     }
 
     func addWindowslessAppsIfNeeded() -> [Window]? {
-        if !Preferences.hideWindowlessApps &&
-               runningApplication.activationPolicy == .regular &&
+        if runningApplication.activationPolicy == .regular &&
                !runningApplication.isTerminated &&
                (Windows.list.firstIndex { $0.application.pid == pid }) == nil {
             let window = Window(self)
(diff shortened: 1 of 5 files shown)


lwouis:
Hi,

I'm afraid this would see pretty niche use. Few people need this setting per-shortcut. It's the case for other similar settings.

I won't be able to merge this PR. We can keep it for other enthusiast users who may want to run their own AltTab version locally.

Thank you 🙇

ThePixelbrain:
Hey, thank your for your reponse.

I can see this being a niche feature that probably won't interest many users.
Do you see a better implementation for this feature that you would merge? I was thinking about maybe always hiding windowless apps when viewing only the visible spaces. Or alternatively add a third dropdown option for the first "Show window of" selection?

lwouis:
Hi,

Thank yor for your message.

I'm afraid I can't see a path where this gets mainstreamed. This setting being global is better for more users. We are in general going to release soon #351 which is a large overall of the Preferences UI. Some settings are going to be removed. So unfortunately, we won't be adding more settings. It's a direction we'll be very cautious going towards in the future.

Any feature or preference will need to pass a cetain threshold of popularity until it's considered. This will avoid the situation where there are 20 settings used by 40 people. Yet the 15k users are exposed to those settings and get lost in the settings UI.

I'm sorry about this. I understand you would like to use it yourself. It's a tough call to make. I hope you understand.

Thank you 🙇

Issue #3653, opened 2024-10-10
AltTab crashes on launch, after update to v7.0.0


omarpach:
**Describe the bug**

When I open Alt-Tab and press CMD-Tab which is the shortcut I used to switch applications, Alt-Tab closes. When I open it again it never shows the "crash" popup, but it is not working properly.

**Screenshots / video**

https://github.com/user-attachments/assets/55d9dc5e-6858-49b3-8673-0b21d20bea01

In the video I show karabiner elements viewer to show that it is registering correctly the CMD-Tab key combination. To keep the video short enough to upload I didn't show it, but Option-Tab also causes it to close. And if I try to open it again to go to preferences it also closes.

**Additional Context**
The app was working properly today but it prompted me to update it which I did, after that I don't remember using it because that happened in school. When i got home and tried to use it it started failing. I tried rebooting my mac but it didn't work, I tried uninstalling it via homebrew and reinstalling it but it didn't work, I tried installing it from the website and running it that way but it didn't work. When that didn't work, I tried updating my MacOS version from 12.7.4 to 12.7.6 and tried once again but nothing changed

**Steps to reproduce the bug**

1. Open Alt-Tab
2. Press CMD-Tab or Option-Tab
3. See error

**Your environment**

* AltTab version: 7.0.0
* macOS version: 12.7.6
* Other relevant info: using external monitor, and using external keyboard as well as karabiner elements

**Attempted solutions**

After looking in to some other issues, I saw someone recommended to run the following command in the terminal
`/Applications/AltTab.app/Contents/MacOS/AltTab`
I tried it and the app did open. The terminal also showed the following errors:

`2024-10-09 19:31:55.080 AltTab[2446:47281] [AppCenterCrashes] WARNING: +[MSACErrorLogFormatter formatStackFrame:report:]/467 Cannot find image for 0x116afa52e
2024-10-09 19:31:57.114 AltTab[2446:47300] [AppCenterCrashes] WARNING: +[MSACErrorLogFormatter formatStackFrame:report:]/467 Cannot find image for 0x108c6852e
2024-10-09 19:31:57.129 AltTab[2446:47300] [AppCenterCrashes] WARNING: +[MSACErrorLogFormatter formatStackFrame:report:]/467 Cannot find image for 0x11e94552e
[...]

PR #3681, opened 2024-10-12
fix the problem that the segmented control is causing to the core

Allsochen:
fix the segmented control is causing to the core. https://github.com/lwouis/alt-tab-macos/issues/3653

QQ_1728717971819

### DIFF
--- a/src/main.swift
+++ b/src/main.swift
@@ -1,9 +1,8 @@
 import AppKit
 import Darwin
 
-func printStackTrace() {
-    let stackSymbols = Thread.callStackSymbols
-    for symbol in stackSymbols {
+func printStackTrace(_ callStackSymbols: [String]) {
+    for symbol in callStackSymbols {
         print(symbol)
     }
 }
@@ -15,7 +14,7 @@ func printStackTrace() {
     signal($0) { s in
         setNativeCommandTabEnabled(true)
         debugPrint("Exiting after receiving signal", s)
-        printStackTrace()
+        printStackTrace(Thread.callStackSymbols)
         exit(0)
     }
 }
@@ -25,7 +24,7 @@ func printStackTrace() {
 NSSetUncaughtExceptionHandler { (exception) in
     setNativeCommandTabEnabled(true)
     debugPrint("Exiting after receiving uncaught NSException", exception)
-    printStackTrace()
+    printStackTrace(exception.callStackSymbols)
     exit(0)
 }
 
--- a/src/main.swift
+++ b/src/main.swift
@@ -1,6 +1,13 @@
 import AppKit
 import Darwin
 
+func printStackTrace() {
+    let stackSymbols = Thread.callStackSymbols
+    for symbol in stackSymbols {
+        print(symbol)
+    }
+}
+
 // - if the app is quit/force-quit from Activity Monitor, it will receive SIGTERM and applicationWillTerminate won't be called
 // - if the app crashes in swift code (e.g. unexpected nil object), SIGTRAP is sent
 // we intercept these signals, and reset the native command-tab shortcut
@@ -8,6 +15,7 @@ import Darwin
     signal($0) { s in
         setNativeCommandTabEnabled(true)
         debugPrint("Exiting after receiving signal", s)
+        printStackTrace()
         exit(0)
     }
 }
@@ -17,6 +25,7 @@ import Darwin
 NSSetUncaughtExceptionHandler { (exception) in
     setNativeCommandTabEnabled(true)
     debugPrint("Exiting after receiving uncaught NSException", exception)
+    printStackTrace()
     exit(0)
 }
 
--- a/src/ui/App.swift
+++ b/src/ui/App.swift
@@ -194,8 +194,8 @@ class App: AppCenterApplication, NSApplicationDelegate {
     @objc func showPreferencesWindow() {
         showSecondaryWindow(preferencesWindow)
         // Use the center function to continue to center, the `repositionPanel` function cannot center, it may be a system bug
-        if let window = preferencesWindow {
-            window.center()
+        if let preferencesWindow = preferencesWindow {
+            preferencesWindow.center()
         }
     }
 
--- a/src/ui/App.swift
+++ b/src/ui/App.swift
@@ -194,7 +194,9 @@ class App: AppCenterApplication, NSApplicationDelegate {
     @objc func showPreferencesWindow() {
         showSecondaryWindow(preferencesWindow)
         // Use the center function to continue to center, the `repositionPanel` function cannot center, it may be a system bug
-        preferencesWindow.center()
+        if let window = preferencesWindow {
+            window.center()
+        }
     }
 
     func showSecondaryWindow(_ window: NSWindow?) {
--- a/src/ui/preferences-window/LabelAndControl.swift
+++ b/src/ui/preferences-window/LabelAndControl.swift
@@ -260,12 +260,14 @@ class LabelAndControl: NSObject {
                 button.setWidth(segmentWidth, forSegment: i)
             }
             if #available(macOS 11.0, *) {
-                if let preference = preference as? SfSymbolMacroPreference {
-                    button.setImage(NSImage(systemSymbolName: preference.symbolName, accessibilityDescription: nil)!, forSegment: i)
+                if let preference = preference as? SfSymbolMacroPreference,
+                   let symbolImage = NSImage(systemSymbolName: preference.symbolName, accessibilityDescription: nil) {
+                    button.setImage(symbolImage, forSegment: i)
                 }
             }
-            if defaults.int(rawName) == i {
-                button.selectedSegment = i
+            let selectedSegment = defaults.int(rawName)
+            if selectedSegment >= 0 && selectedSegment < macroPreferences.count {
+                button.selectedSegment = selectedSegment
             }
             _ = setupControl(button, rawName, String(i), extraAction: extraAction)
         }
--- a/src/ui/preferences-window/PreferencesWindow.swift
+++ b/src/ui/preferences-window/PreferencesWindow.swift
@@ -48,7 +48,7 @@ class PreferencesWindow: NSWindow, NSToolbarDelegate {
             (4, NSLocalizedString("Blacklists", comment: ""), "blacklists", "hand.raised", BlacklistsTab.initTab()),
             (5, NSLocalizedString("About", comment: ""), "about", "info.circle", AboutTab.initTab()),
             (6, NSLocalizedString("Acknowledgments", comment: ""), "acknowledgments", "hand.thumbsup", AcknowledgmentsTab.initTab()),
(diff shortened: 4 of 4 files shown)


src/ui/App.swift
```diff
@@ -194,7 +194,9 @@ class App: AppCenterApplication, NSApplicationDelegate {
     @objc func showPreferencesWindow() {
         showSecondaryWindow(preferencesWindow)
         // Use the center function to continue to center, the `repositionPanel` function cannot center, it may be a system bug
-        preferencesWindow.center()
+        if let preferencesWindow = preferencesWindow {
```
lwouis, line 197: Normally, the `preferencesWindow` is initialized before. It should always be.

I'm a bit worried that if we wrap such statement in conditions, it may hide the issues. Like that window doesn't show up because this condition is not passed. Wouldn't it be better that the app crash at this point? At least we get a crash report, and we know something is wrong. Otherwise the user has AltTab running with no `preferencesWindow` which shouldn't be possible, and may cause other bad side-effects later on.

What do you think?

src/ui/preferences-window/LabelAndControl.swift
```diff
@@ -260,12 +260,14 @@ class LabelAndControl: NSObject {
                 button.setWidth(segmentWidth, forSegment: i)
             }
             if #available(macOS 11.0, *) {
-                if let preference = preference as? SfSymbolMacroPreference {
-                    button.setImage(NSImage(systemSymbolName: preference.symbolName, accessibilityDescription: nil)!, forSegment: i)
+                if let preference = preference as? SfSymbolMacroPreference,
+                   let symbolImage = NSImage(systemSymbolName: preference.symbolName, accessibilityDescription: nil) {
+                    button.setImage(symbolImage, forSegment: i)
```
lwouis, line 265: I don't think this part has issues since every single crash is running macOS 10.12 or lower. They shouldn't run code in that section.

Also, I think it's healthy to keep the code crashing here. This helps us identify if an image is missing during development. If we had this condition here, we may ship a release where we the images don't work, and the users will have a degraded experience without images.

src/ui/preferences-window/LabelAndControl.swift
```diff
@@ -260,12 +260,14 @@ class LabelAndControl: NSObject {
...
                 }
             }
-            if defaults.int(rawName) == i {
-                button.selectedSegment = i
+            let selectedSegment = defaults.int(rawName)
+            if selectedSegment >= 0 && selectedSegment < macroPreferences.count {
+                button.selectedSegment = selectedSegment
```
lwouis, line 270: I'm not sure this helps. `i` is generated by the loop and its `enumerated()`. It's thus garanteed to be `> 0` and `< macroPreferences.count`.

Furthermore, the previous code was only assigning a value if it matches. I don't think it could assign a wrong value and create issues here.

Finally, the crash reports mention line `AppearanceTab.swift:414` which creates the UI for `appearanceSize`. However, in those reports, we see that users don't have this preference set. So it doesn't seem like this would be the root cause.

I've invited you on AppCenter. Let me know if it works 👍

src/ui/preferences-window/LabelAndControl.swift
Allsochen: QQ_1728720386963

This requires macOS 13+

src/ui/preferences-window/LabelAndControl.swift
Allsochen: let selectedSegment = defaults.int(rawName)

The value stored by the user may have problems. The main reason for core is not here, but this code will become more robust

src/ui/App.swift
Allsochen:
```swift    
    func applicationShouldHandleReopen(_ sender: NSApplication, hasVisibleWindows flag: Bool) -> Bool {
        showPreferencesWindow()
        return true
    }
```

I just encountered a problem with `preferencesWindow` when running with XCode. Its call is in this function, and App Center also reported the problem here

src/ui/preferences-window/LabelAndControl.swift
Allsochen: There is no image but text is displayed. I think it is a better way to downgrade the display text here. 

What do you think?

src/main.swift
```diff
@@ -1,13 +1,21 @@
...
+}
+
 // - if the app is quit/force-quit from Activity Monitor, it will receive SIGTERM and applicationWillTerminate won't be called
 // - if the app crashes in swift code (e.g. unexpected nil object), SIGTRAP is sent
 // we intercept these signals, and reset the native command-tab shortcut
 [SIGTERM, SIGTRAP].forEach {
     signal($0) { s in
         setNativeCommandTabEnabled(true)
         debugPrint("Exiting after receiving signal", s)
+        printStackTrace()
```
lwouis, line 18: I tried to test this, and somehow this signal catching function seems to no longer work. I'm surprised because it worked in the past.

Maybe swift changed or macOS changed. Something seems to have broken and now I can't see any print here.

Allsochen: QQ_1728722754252

I can show it here, adding this allows users to give us some error traces

src/ui/preferences-window/LabelAndControl.swift
lwouis: Oh! So we add an image on macOS `>= 11` but some images have specific availability, and swift is not protecting us from this. That's a very nice catch.

I don't think it explains the crashes here, since users who crash are on macOS `<= 10.12`, so they shouldn't enter the top condition in the first place. However, we your change definitely will help users on macOS `>= 11 and < 13`. I'll add a comment so we remember later why it's important to not assume the images exist 👍

src/ui/preferences-window/LabelAndControl.swift
lwouis: `defaults.int(rawName) == i` protects us from bad data. It garantees that if we assign, it's a valid value for us.

I agree that we should have a better way to handle preferences. The worst releases we've had over the years have always been about bad preferences migration. In those cases, the app can't launch. It's the worse case.

I've added bf857e8f251cce38cb38d063161bf6a90c4a14a1 at some point. It makes it so a preferences that's of the wrong type will be removed, so the user isn't stuck. However it's only a type-check. We don't check that that value is valid beyond the type. It would be nice to add more safety in a generic way instead of checks all over the codebase when the preference is used. I'm not sure how to go about it though

src/ui/preferences-window/LabelAndControl.swift
lwouis: Yes, it's definitely an issue. However, as I said, it should only create crashes in the case were macOS `>= 11 and < 13`. Yet our crashed users are all on macOS 10.12 or 10.11. Weird, no?

Allsochen: QQ_1728723566159

The version of macOS that the user is having problems with is `12.7.6`, So I think that's the root reason. 
If we don’t downgrade the experience, users won’t be able to use the app at all, which will have a greater impact.

src/main.swift
lwouis: Were you able to show this in the debug build? Could you share the steps? I'm not able to do this locally on my debug build

src/ui/App.swift
lwouis: That's interesting. where did you see this in AppCenter? I can't find it.

Maybe we should add the condition in `applicationShouldHandleReopen` then? What do you think?

Allsochen: QQ_1728725306943

I think it's OK to put this nil judgment there, because there is also a layer of judgment here.

src/ui/preferences-window/LabelAndControl.swift
Allsochen: > Yes, it's definitely an issue. However, as I said, it should only create crashes in the case were macOS `>= 11 and < 13`. Yet our crashed users are all on macOS 10.12 or 10.11. Weird, no?

You can see the crash version on AppCenter, which is distributed in `12.7.6`,`11.7.10`,`12.7.5`.

src/ui/preferences-window/LabelAndControl.swift
Allsochen: > I've added [bf857e8](https://github.com/lwouis/alt-tab-macos/commit/bf857e8f251cce38cb38d063161bf6a90c4a14a1) at some point. It makes it so a preferences that's of the wrong type will be removed, so the user isn't stuck. However it's only a type-check. We don't check that that value is valid beyond the type. It would be nice to add more safety in a generic way instead of checks all over the codebase when the preference is used. I'm not sure how to go about it though

A universal validation method is a good idea. There are many such frameworks in other languages, but ours will not be too complicated. We don't need to introduce additional frameworks. We just need to roughly implement a version.

src/main.swift
Allsochen: QQ_1728726415927

Yes, I am in the debug build. I add the followed code to App.swift on app start up.

```swift
        let a = [1]
        print(a[5])
```

lwouis: I was able to get the same result as you with this particular way 👍 

What I can't get to work, that I know worked before, is to send signals from outside the app. Like `ctrl+C`, or `kill pid-of-alttab`, or stopping the debugger.

I see the comment I left in the past `if the app is quit/force-quit from Activity Monitor, it will receive SIGTERM`. I tried quitting from Activity Monitor and it doesn't trigger it:

image

Anyway, it's not related to the issue and this PR. It's a general observation. We can look into it another time 👍

src/ui/App.swift
lwouis: It's a good point. I wonder why the `center()` call is only on the Preferences window. It would make sense to apply it directly inside `showSecondaryWindow`, so that any secondary window is shown centered.

I've slightly updated to this:

image

lwouis:
I've launched a release. I hope it goes through. I think there is only enough TravisCI credits for 1 last release. After that, we won't be able to release until they refill our account.

I have opened a ticket to them for that purpose 3 days ago. I have added a comment to it now, to mention that we really need credits at the moment. I hope they can manage the ticket quickly 🙏 

Closing the MR since it's cherry-picked and on `master` now 👍

Allsochen:
> I've launched a release. I hope it goes through. I think there is only enough TravisCI credits for 1 last release. After that, we won't be able to release until they refill our account.
> 
> I have opened a ticket to them for that purpose 3 days ago. I have added a comment to it now, to mention that we really need credits at the moment. I hope they can manage the ticket quickly 🙏
> 
> Closing the MR since it's cherry-picked and on `master` now 👍

Some of the code I submitted later did not seem to be merged into the `master`, but it has no impact for now. I saw that you mentioned another mr to fix it.

lwouis:
> Some of the code I submitted later did not seem to be merged into the master, but it has no impact for now. I saw that you mentioned another mr to fix it.

I think I got the fixes from all 4 commits on `master` at this point. They should all be in v7.0.2. By the way, we have no more TravisCI credits. So for new releases, we need to hope that the TravisCI Support team handles it quickly 🙏

PR #3683, opened 2024-10-12
fix: update German Localizable.strings

stefanlange:
Unfortunately, many of the original German translations did not make sense or the correct words were not used.

### DIFF
--- a/resources/l10n/de.lproj/Localizable.strings
+++ b/resources/l10n/de.lproj/Localizable.strings
@@ -17,13 +17,13 @@
 "Active screen" = "Aktiver Bildschirm";
 
 /* No comment provided by engineer. */
-"Additional controls" = "Zusätzliche Kontrollen";
+"Additional controls" = "Zusätzliche Bedienelemente";
 
 /* No comment provided by engineer. */
-"Additional controls…" = "Zusätzliche Kontrollen…";
+"Additional controls…" = "Zusätzliche Bedienelemente…";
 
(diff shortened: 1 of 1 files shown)


lwouis:
Hi @stefanlange,

We have used Google Translate to translate the many new labels coming with v7.0.0. It's a new approach. Usually we have English strings until someone adds a translation using poeditor.

We can't accept translations via PR. It has to go through poeditor. Here's the [link to contribute there](https://poeditor.com/join/project/8AOEZ0eAZE). However, please be patient for a while. We haven't added the new strings from v7.0.0 to poeditor yet. We'll add them soon.

Then you can add the translation to poeditor, and they will be in the next release! 👍

stefanlange:
Hi,

I have seen that the translation resources for v7.0.2 are available on
poeditor.com. I've updated the German translations there so that they can
be used in the next release.

It would be helpful for other translators if you can provide an option to
switch the language inside the app. This would make it easier to get some
more context while looking for the right words.

Regards,

Stefan

Am Sa., 12. Okt. 2024 um 18:29 Uhr schrieb lwouis ***@***.***

lwouis:
> I have seen that the translation resources for v7.0.2 are available on
poeditor.com. I've updated the German translations there so that they can
be used in the next release.

Thank you!

> It would be helpful for other translators if you can provide an option to
switch the language inside the app. This would make it easier to get some
more context while looking for the right words.

It will be part of the next release 👍

PR #3684, opened 2024-10-12
fixed the rounded corners when the mouse moves over the table

Allsochen:
fixed the rounded corners when the mouse moves over the table.

image


lwouis:
This is live 👍

Issue #3656, opened 2024-10-10
Why is ´mouse follows focus´missing after the newest update?


qns7:
Hey there,

this option (title) is missing now, right? Is this intentional or a bug? The release notes didn't say that the feature was dropped.

Best regards Q

Allsochen:
Hey, I'm very sorry. During the refactoring process, we thought that few users would use this option and that it interfered with the user's normal behavior, so we decided to delete the option. If there are many users who give feedback, we will reconsider adding it back.

qns7:
> Hey, I'm very sorry. During the refactoring process, we thought that few users would use this option and that it interfered with the user's normal behavior, so we decided to delete the option. If there are many users who give feedback, we will reconsider adding it back.

I also use yabai extensively and it has this option, but with the yabai one I have problems with some apps: for example zoom.us; and your implementation was flawless and worked without any problems on all my apps. I guess I then have to live with the yabai feature and make it somehow togglable for apps that don't work with it.

Thanks for the fast answer.

Allsochen:
OK, I understand, we will revisit this feature later.

lwouis:
Closing this ticket for now. As @Allsochen mentioned, let's revisit this later after we see the general reaction to the latest update 👍

PR #3685, opened 2024-10-12
Enable cursor follow focus

Allsochen:
revert enable cursor follow focus functionality. https://github.com/lwouis/alt-tab-macos/issues/3656

### DIFF
--- a/resources/l10n/Localizable.strings
+++ b/resources/l10n/Localizable.strings
@@ -525,3 +525,7 @@
 
 /* No comment provided by engineer. */
 "You didn’t write your email, thus can’t receive any response." = "You didn’t write your email, thus can’t receive any response.";
+
+"Miscellaneous" = "Miscellaneous";
+
+"Cursor follows focus" = "Cursor follows focus";
--- a/resources/l10n/ar.lproj/Localizable.strings
+++ b/resources/l10n/ar.lproj/Localizable.strings
@@ -529,3 +529,9 @@
 /* No comment provided by engineer. */
 "You didn’t write your email, thus can’t receive any response." = "لم تكتب بريدك؛ فلذلك لن تحصل على رد.";
 
+/* No comment provided by engineer. */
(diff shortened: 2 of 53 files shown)


lwouis:
Hey @Allsochen,

I think it's a bit early to bring this feature back. When we worked on the big Revamp, we discussed that we have to cut niche feature. Of course it means some powerusers will be displeased. But that's for the greater good of having a better app for most people.

It's important to only add feature if there is sizable demand for them. Otherwise the app will soon become bloated, and we will need another #351 rework 😅 

Let's see how many people ask for the feature back before we make any move, okay?

Thank you

Allsochen:
I think this feature is very different from Refactor https://github.com/lwouis/alt-tab-macos/issues/351. https://github.com/lwouis/alt-tab-macos/issues/351 is about how to implement product features. Here, features are either available or not. Some users require this feature to meet certain specific requirements of users, making them like or strongly rely on our software. The reason I agreed to delete this feature before was that I felt that no users would use it, but now some users are asking to add it back.

blackadder-github:
> Omg. Finally I found this topic. I spent half an hour to understand what the hell is going on and why this feature has gone! Please, give it back!) It is crucial for people who use few monitors (I think). Thanks.

I strongly agree! It is such a useful feature that should be integrated in vanilla MacOS (like many other things...).

aquila-zyy:
Why I use this feature: 
I always enables the auto-raise feature from yabai. Usually, my desktop layout has a maximized window on the background, with some windowed reference materials floating around. I rely on my cursor jumping to the floating windows when I Alt-Tab so that they are not immediately hidden if I move my cursor (which, if not already on the floating window, auto-raises the maximized window). The work around is either me pressing a modifier key that temporarily disable this yabai feature, or move my cursor to where the window would pop up before alt-tabbing. Neither solution is as smooth as it used to be.

lwouis:
This is in the app currently. Closing this PR 👍

PR #3687, opened 2024-10-13
Add language preference

Allsochen:
add language preference. https://github.com/lwouis/alt-tab-macos/issues/3686


lwouis:
This is live 👍

PR #3802, opened 2024-11-03
fix: fix wide tall thumbnails

Allsochen:
Increase the width of the thumbnails to ensure that as many thumbnails as possible have the same height.


lwouis:
I cherry-picked the commit 👍

PR #3826, opened 2024-11-08
Fix false positives for isShortcutAlreadyAssigned

### LINKED ISSUE #3288: Unable to assign any "select next window" with a combination "hold" button shortcut that conflicts with another shortcut that shares an individual "hold" button
**Describe the bug**
The conflict checker checks to see if any individual "Hold" key exists somewhere else even if it's used in combination.  If an individual key from combo "Hold" key is used individually in another shortcut, it fails.

Examples:
Suppose I use (⌥) + <any key> in shortcut one.
Then any other shortcut that uses  (⌥+⌘)+<same any key> doesn't work

It happens with any individual "Hold" key.
For ex:
⇧ + J
Then I try
⇧⌥ + J, it gives a conflict error.

There is a work around currently:
unassign ⇧ + J
assign ⇧⌥ + J, no conflict
reassign ⇧ + J, no conflict

Then both of them work perfectly as expected.

**Your environment**

* AltTab version: 6.68
* macOS version: 14.3
* Other relevant info: e.g. using external keyboard, multiple monitors, etc

phazei:
(closes #3288)

The issue is described in detail in the issue.  This resolves that issue, it was tested as working on a 2024 MBP running Sonoma 14.6.1.

### DIFF
--- a/src/ui/generic-components/CustomRecorderControl.swift
+++ b/src/ui/generic-components/CustomRecorderControl.swift
@@ -87,30 +87,37 @@ class CustomRecorderControl: RecorderControl, RecorderControlDelegate {
     }
 
     private func isShortcutAlreadyAssigned(_ shortcut: Shortcut) -> ATShortcut? {
+        // Get corresponding paired shortcut name and object
         let comboShortcutName = id.starts(with: "holdShortcut") ?
             Preferences.indexToName("nextWindowShortcut", Preferences.nameToIndex(id)) :
             (id.starts(with: "nextWindowShortcut") ?
                 Preferences.indexToName("holdShortcut", Preferences.nameToIndex(id)) : id)
         let comboShortcut = comboShortcutName.flatMap { ControlsTab.shortcuts[$0]?.shortcut }
+        
         return (ControlsTab.shortcuts.first { (id2, s2) in
             let shortcut2 = s2.shortcut
+            // Skip self-comparison, empty shortcuts, hold shortcuts, and paired shortcuts
             if id == id2
                    || (shortcut2.keyCode == .none && shortcut2.carbonModifierFlags == 0)
                    || id2.starts(with: "holdShortcut")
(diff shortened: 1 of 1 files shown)


lwouis:
I did some QA. Here are some issues I've found (S1 = Shortcut1):

* Set S1 to (alt)+(shift); Set S2 to (alt+shift)+(cmd) => It lets me set it, even though S1 will break S2. That is, when I release cmd, instead of not doing anything and continue showing me S1 windows, it thinks I'm not holding S1 and cycle through S1 windows.
* Set S1 to (alt)+(tab) and S2 to (alt+shift)+(tab). Press S2, then keep tab pressed and release shift => it closes instead of switching to S1

phazei:
So, I can set some of those shortcuts as you mention doing them backwards (setting S2 first) as it is now, which is also the workaround for the original ticket which would otherwise disallow completely valid shortcuts.

So, as long as it doesn't totally break things, the question is, to allow a user to not select some valid shortcuts, or let them inadvertently pick something that could be problematic.  If a user chooses to select overlapping shortcuts, I'd personally say that's on them, it's better to allow valid shortcuts to be selected.

The current check could be handled better in a couple way.  Perhaps should also check for "previousWindowShortcut" to check against that too.  It didn't seem to have anything about that in there before so I didn't, but it wouldn't be hard to add.  I would say that more limitations could potentially be placed on the "holdShortcut".  First, since the "previousWindowShortcut" is a single keypress, I'd suggest that the "holdShortcut" shouldn't be allowed to include whatever key is used in "previousWindowShortcut".  I'd also suggest that the "nextWindowShortcut" shouldn't be allowed to be a modifier key, which should be strictly for the hold, so no ctrl/opt/command/shift" on "nextWindowShortcut"

That partially is slightly counter to what I said about letting the user pick something problematic and it being their problem, but also, it seems to be within standard of how most OS shortcuts function.

What do you think?

PR #3831, opened 2024-11-11
fix: fix max width on screen

Allsochen:
fix max width on screen


lwouis:
This has been taken into account in `comfortableWidth` a little while ago. The width is calculated based on what's comfortable for users based on field-of-view. The value is clamped between 45% and 90% like this PR.

I'll close the PR 👍

PR #3837, opened 2024-11-12
fix: arabic localization

SamiSuleiman:
Changed the "Visible Spaces" value for arabic from "المسافات الظاهرة" to "المساحات الظاهرة"


lwouis:
Hi,

Localizations are not updated via MRs. They are updated directly on poeditor.com. Could you please update them here: https://poeditor.com/join/project/8AOEZ0eAZE

Thank you 🙇

PR #3873, opened 2024-11-20
Improve Turkish localization

maliayas:


lwouis:
Hi,

Localizations are not updated via MRs. They are updated directly on poeditor.com. Could you [please update them here](https://poeditor.com/join/project/8AOEZ0eAZE)?

Thank you 🙇

maliayas:
Allright, I've updated POEditor strings. Thanks.

Issue #814, opened 2021-02-10
Text is not displayed completely


LarkinZero:
**Describe the bug**

The top part of Chinese text is not displayed.

**Screenshots / video**
<img width="288" alt="Xnip2021-02-10_23-52-19" src="https://user-images.githubusercontent.com/6387162/107535003-514e9580-6bfb-11eb-888a-b39ad99d4931.png">

My settings
<img width="606" alt="Xnip2021-02-10_23-52-51" src="https://user-images.githubusercontent.com/6387162/107535028-57447680-6bfb-11eb-8f22-9c5593f6fc7d.png">

**Steps to reproduce the bug**

I tried to modify the font size, but no effect.

lwouis:
Hi @LarkinZero! Thanks for sharing this issue!

I'm not very familiar with Chinese hanzi. What is wrong exactly on your first screenshot? Is it that there are a few pixels shaved off to the top of the text? Is it on all rows like Chrome, Finder, etc, or is it just the very first row?

I need more details to really understand the scope of the issue

Thank you!

LarkinZero:
Yes,there are a few pixels shaved off to the top of the text.
It looks every row has this problem.
![](https://user-images.githubusercontent.com/6387162/107550488-0ee18480-6c0c-11eb-9ec9-108496aebbd2.png)

LarkinZero:
Not only Chinese characters, but also some symbols
Like `[]`
<img width="1564" alt="Xnip2021-02-11_01-59-22" src="https://user-images.githubusercontent.com/6387162/107551273-0473ba80-6c0d-11eb-846a-c5cd473918c6.png">

LarkinZero:
And it happening as well with the thumbnails visible
<img width="459" alt="Xnip2021-02-11_02-10-55" src="https://user-images.githubusercontent.com/6387162/107552476-89ab9f00-6c0e-11eb-9514-fe9bd38bb473.png">

lwouis:
@LarkinZero I can't seem to reproduce the issue. Could you please share the links to some of the websites on your screenshots? Or just write here the hanzi and I'll search them on Google. Thank you

LarkinZero:
https://www.google.com/search?q=%5B%5D&oq=%5B%5D&aqs=chrome..69i57.1055j0j7&sourceid=chrome&ie=UTF-8
Text is : []
[...]

PR #3976, opened 2024-12-13
fix: fix the title being covered

Allsochen:
fix  the title being covered.

Before fixed:
image

After fixed:
image

### DIFF
--- a/src/ui/main-window/ThumbnailTitleView.swift
+++ b/src/ui/main-window/ThumbnailTitleView.swift
@@ -41,6 +41,15 @@ class ThumbnailTitleView: BaseLabel {
         return .byTruncatingHead
     }
 
+    /// Draws the text within the given dirty rectangle.
+    ///
+    /// This function is responsible for rendering the text content in the view.
+    /// It ensures that the text is properly aligned and vertically centered
+    /// within the view’s bounds. The `NSLayoutManager` is utilized to manage
+    /// the layout of the glyphs (text characters) and draw them at the
+    /// calculated position.
+    ///
+    /// - Parameter dirtyRect: The portion of the view’s bounds that needs to be updated.
     override func draw(_ dirtyRect: NSRect) {
         guard let textStorage = textStorage,
               let layoutManager = textStorage.layoutManagers.first,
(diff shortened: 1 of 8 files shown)


lwouis:
Hi @Allsochen,

Thank you for sharing another PR!

Could you please share a bit more context about this PR? I've worked a lot on https://github.com/lwouis/alt-tab-macos/issues/814 in the past. I couldn't find a solution. It seems that you just change the font size here, for the small size.

Is this a quick-win perhaps? If you have any input on https://github.com/lwouis/alt-tab-macos/issues/814, i would be very interested.

Thank you!

Allsochen:
> Hi @Allsochen,
> 
> Thank you for sharing another PR!
> 
> Could you please share a bit more context about this PR? I've worked a lot on #814 in the past. I couldn't find a solution. It seems that you just change the font size here, for the small size.
> 
> Is this a quick-win perhaps? If you have any input on #814, i would be very interested.
> 
> Thank you!

I didn't notice the problem with #184 before, but the latest code I submitted has fixed the problem.

src/ui/main-window/ThumbnailTitleView.swift
```diff
@@ -37,6 +41,24 @@ class ThumbnailTitleView: BaseLabel {
         return .byTruncatingHead
     }
 
+    override func draw(_ dirtyRect: NSRect) {
```
lwouis, line 44: Could you please share why this override is necessary? What does it change compared to the native implementation? We should probably add a comment to clarify this, so that reads can understand this change better 👍

src/ui/main-window/ThumbnailTitleView.swift
```diff
@@ -60,7 +82,11 @@ class ThumbnailTitleView: BaseLabel {
         return ceil(textSize.width)
     }
 
+    static func extraLineSpacing(for fontSize: CGFloat) -> CGFloat {
+        return fontSize * 0.2
```
lwouis, line 86: I imagine that 20% margin comes from your testing? Less than 20% didn't work in your tests?
Thank you

lwouis:
I've merged parts of this. I've also reworked the titles to use NSTextField instead of NSTextView. So it's a bit different now. I think the issue shown in your screenshot should be gone now.

Thank you

PR #3979, opened 2024-12-15
fix: always compute preview window origin.y using primary display

delaneyb:
The `previewFocusedWindowIfNeeded` function incorrectly positions the preview window when multiple monitors are in use, and the `Preferences.showOnScreen` setting is not set to `includingMenubar`. This happens because`NSScreen.preferred()` uses the screen with the mouse or the currently active screen when performing its calculation for the coordinate inversion of the window Preview Frame y origin position. Instead, it should **always** use the primary screen (`NSScreen.screens.first`), which is the reference point for the global coordinate system, to determine the correct position of the preview panel.

https://github.com/user-attachments/assets/1407c549-bcd8-4e19-94e5-322eaac8766f

### DIFF
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -161,7 +161,10 @@ class Windows {
         }
         App.app.previewPanel.setPreview(preview)
         var frame = NSRect(origin: position, size: size)
-        frame.origin.y = NSScreen.preferred().frame.maxY - frame.maxY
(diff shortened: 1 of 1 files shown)


lwouis:
Thank you for sharing this fix
I've cherry-picked it. It will be shipped in the next release!
Thank you 🙇

PR #3980, opened 2024-12-15
fix: resolve z-order flicker in Windows preview panel

delaneyb:
This fixes an intermittent issue where, if you switch between previewing windows on different monitors, the newly previewed window momentarily renders above the thumbnails view. When the system is under heavy load, and you have the "selection follows cursor" setting turned on, this can lead to constant flickering back and forth.

This can be solved by setting the level of the preview panel explicitly to one less than that of the thumbnails panel (100 as opposed to `.popUpMenu` = 101 of the thumbnailsView)

https://github.com/user-attachments/assets/2dc54322-189e-44d8-a055-6d339635b7a9


lwouis:
Thank you for the thorough analysis, and sharing a fix!
I've cherry-picked it. It will be shipped in the next release!
Thank you 🙇

Issue #3669, opened 2024-10-11
Suggestion: Add "Auto" to size option


danilagalimov:
Currently there are 3 types of Window preview sizes: Small, Medium and Large.

However, I'd like to utilize as much screen space as possible,so I suggest to add "Auto" setting which will select the maximum available option based on Windows which are available for switching just when you press switch start key, so, if all windows can fit on Large screen, it is selected, if not - then Medium and Large accordingly.

Allsochen:
Hey, this is a very good idea. We also thought about such a feature when we refactored version 7.0. Due to the large changes in the refactoring, this feature has been temporarily shelved. Similar features can be added later, and PRs are also welcome.

PR #4058, opened 2025-01-04
fix: unify font icon size in difference style

Allsochen:
unify font icon size in difference style

### DIFF
--- a/src/logic/Appearance.swift
+++ b/src/logic/Appearance.swift
@@ -83,12 +83,12 @@ class Appearance {
         switch currentSize {
             case .small:
                 rowsCount = isHorizontalScreen ? 5 : 8
-                iconSize = 20
-                fontHeight = 12
+                iconSize = 23
+                fontHeight = 13
             case .medium:
                 rowsCount = isHorizontalScreen ? 4 : 7
-                iconSize = 30
-                fontHeight = 13
+                iconSize = 28
+                fontHeight = 14
             case .large:
                 rowsCount = isHorizontalScreen ? 3 : 6
                 iconSize = 32
@@ -117,11 +117,11 @@ class Appearance {
                 fontHeight = 13
             case .medium:
                 iconSize = 128
-                fontHeight = 15
+                fontHeight = 14
             case .large:
                 windowPadding = 28
                 iconSize = 168
-                fontHeight = 17
+                fontHeight = 16
         }
     }
 
@@ -137,10 +137,10 @@ class Appearance {
         rowsCount = 1
         switch currentSize {
             case .small:
-                iconSize = 20
+                iconSize = 23
                 fontHeight = 13
             case .medium:
-                iconSize = 26
+                iconSize = 28
                 fontHeight = 14
             case .large:
                 iconSize = 32


lwouis:
Hi @Allsochen,

Could you please elaborate on what the problem is with the current values, and how you picked the new ones?

Thank you 🙇

Allsochen:
> Hi @Allsochen,
> 
> Could you please elaborate on what the problem is with the current values, and how you picked the new ones?
> 
> Thank you 🙇

Ensure that the middle size is consistent with the system in each style, and also ensure that the font size of each style is the same, so that the user experience is more consistent.

Also can you explain why the time logs here are inconsistent?
QQ 2025-01-06 11 10 42

lwouis:
Hey @Allsochen,

Sorry for the delay in responding

The default font size on macOS is 13px. That's why the current Thumbnails Style Medium is at 13px.
The default AppIcons Style Medium is at 15px to match the native app-switcher
Finally, for the default Titles style, the icons has been made 26 for the Medium to get a more compacted look. The icons decide the height of each row. In Titles mode, users most likely want to have the rows a bit compacted; not too much vertical padding.

You can see the comparison bellow. Be careful to open the images separately. Here in the table, they are upscaled, so they visually don't match. When you open them separately, you can see the size matches.

Style | Native | AltTab
-|-|-
App Icons | image | image
Thumbnails | image | image

What do you think?

Thank you

lwouis:
> Also can you explain why the time logs here are inconsistent?

The logs in red come from macOS internals. They have their own logs, with their own format. We can't control their formatting, unfortunately

Allsochen:
> > Also can you explain why the time logs here are inconsistent?
> 
> The logs in red come from macOS internals. They have their own logs, with their own format. We can't control their formatting, unfortunately

In the case of multi-threaded execution, there will be very subtle differences. In order to clearly see which one is executed and which one is executed later, the default time is `2025-01-06
11:07:10.008`. I don’t think we need to modify it to make it shorter.

lwouis:
Hey,
Regarding the log format, I went with the following logic.
Most of the time, it's useful to have a short format. It's especially useful for the use case of users looking at logs to try and debug an issue. It's also useful locally for developpement, in most case.
You're correct that in some cases, more detailed logs are useful. In those cases, we can increase precision by directly updating the format in the code.
Thank you

Allsochen:
Hey，
Many times we cannot predict when problems will occur during parallel execution, so we don’t think about improving accuracy. Therefore, I recommend using a high-precision time format at all times.

What do you think?

lwouis:
Hey,

Indeed, it can be useful. I'll revert the format to the long timestamps. I can always override it to the shorter format locally if needed 👍 

What do you think about [this message](https://github.com/lwouis/alt-tab-macos/pull/4058#issuecomment-2585367756)?

Thank you

Allsochen:
> What do you think about https://github.com/lwouis/alt-tab-macos/pull/4058#issuecomment-2585367756?

I agree with you. There are differences between AppIcons mode and Thumbnails mode. 
However, there is a problem here. In Thumbnails mode, Windows 11 will automatically adjust the number of thumbnails displayed, but it only reduces the thumbnails, and does not reduce the icons and fonts. However, the font sizes of our small, medium and large fonts will change, so we need a way to meet this demand. 

What do you think?

lwouis:
> Windows 11 will automatically adjust the number of thumbnails displayed, but it only reduces the thumbnails

I think this would be covered by https://github.com/lwouis/alt-tab-macos/issues/3669. Today we offer fixed sizes. We would need to add an automatic size. When we do that, we can look into how Windows 11 is doing it, for inspiration.

lwouis:
Size: Auto was added 👍 
Closing the PR

PR #4164, opened 2025-01-23
fix and feat

chenxiankong:
fix: 
1. fix isOnScreen logic 

feat:
1. filter no normal type screen (For example, the window when creating a meeting in Lark is abnormal)
2. Optimize the decision logic of similar apps

### DIFF
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -25,6 +25,9 @@ class Window {
     var application: Application
     var axObserver: AXObserver?
     var rowIndex: Int?
+    
+    var spaceIds = [UInt64]()
+    var isNormal = true
 
     static let notifications = [
         kAXUIElementDestroyedNotification,
@@ -36,10 +39,13 @@ class Window {
     ]
 
     init(_ axUiElement: AXUIElement, _ application: Application, _ wid: CGWindowID, _ axTitle: String?, _ isFullscreen: Bool, _ isMinimized: Bool, _ position: CGPoint?, _ size: CGSize?) {
+        let subRole = try? axUiElement.subrole()
+        self.isNormal = subRole == "AXStandardWindow"
         // TODO: make a efficient batched AXUIElementCopyMultipleAttributeValues call once for each window, and store the values
         self.axUiElement = axUiElement
         self.application = application
         cgWindowId = wid
+        debugPrint("33init")
         spaceId = Spaces.currentSpaceId
         spaceIndex = Spaces.currentSpaceIndex
         self.isFullscreen = isFullscreen
@@ -235,7 +241,16 @@ class Window {
     func isOnScreen(_ screen: NSScreen) -> Bool {
         if NSScreen.screensHaveSeparateSpaces {
             if let screenUuid = screen.uuid(), let screenSpaces = Spaces.screenSpacesMap[screenUuid] {
-                return screenSpaces.contains { $0 == spaceId }
+                debugPrint("22|",spaceId,screen,screenSpaces,title!)
+                for v1 in screenSpaces {
+                    for v2 in spaceIds {
+                        if v1 == v2 {
+                            return true
+                        }
+                    }
+                }
+                return false
+//                return screenSpaces.contains { $0 == spaceId }
             }
         } else {
             let referenceWindow = referenceWindowForTabbedWindow()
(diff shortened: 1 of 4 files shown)


src/logic/Windows.swift
```diff
@@ -345,14 +349,23 @@ class Windows {
...
-        window.shouldShowTheUser =
+        var isSameApp = false
+        if let a = window.application.runningApplication.bundleIdentifier{
+            if let b = NSWorkspace.shared.frontmostApplication?.bundleIdentifier{
+                if a.contains(b) || b.contains(a) {
+                    isSameApp = true
+                }
+            }
+        }
+        window.shouldShowTheUser = window.isNormal &&
```
lwouis, line 360: Could you please share why you added this check? Currently we compare the pid of the 2 apps. It seems it's not enough, and you added another check of bundleIds. Could you please share in which cases checking the pids was not working?

To my knowledge, every app has a pid, but not every app has a bundleID. So in that sense, checking pid should be broader, and sufficient.

src/logic/Window.swift
```diff
@@ -36,10 +39,13 @@ class Window {
     ]
 
     init(_ axUiElement: AXUIElement, _ application: Application, _ wid: CGWindowID, _ axTitle: String?, _ isFullscreen: Bool, _ isMinimized: Bool, _ position: CGPoint?, _ size: CGSize?) {
+        let subRole = try? axUiElement.subrole()
+        self.isNormal = subRole == "AXStandardWindow"
```
lwouis, line 43: We alread check this in the `isActualWindow` method, which is used before calling this constructor. So I think rechecking here is not necessary.

Perhaps you saw something that I'm missing?

lwouis:
Hi @chenxiankong,

I've looked into your PR, and made a few changes. Could you please test [this local build](https://github.com/user-attachments/files/18547259/AltTab.app.zip), and let me know if it handles every case correctly for you, regarding the original issue?

Thank you

lwouis:
Hi @chenxiankong,
Any chance you could look at [my previous message](https://github.com/lwouis/alt-tab-macos/pull/4164#issuecomment-2614046297)?
Thank you 🙇

src/logic/Windows.swift
chenxiankong: When some apps start windows, they may add a suffix to the bundle.
for example, main app`s bundle is com.bytedance.lark
if lark start a video chat window, the bundle may be com.bytedance.lark.video
these code want to fix it

chenxiankong:
> Hi @chenxiankong,
> 
> I've looked into your PR, and made a few changes. Could you please test [this local build](https://github.com/user-attachments/files/18547259/AltTab.app.zip), and let me know if it handles every case correctly for you, regarding the original issue?
> 
> Thank you

hi @lwouis 
it fix !

src/logic/Window.swift
chenxiankong: This method(isActualWindow) is not perfect, so use lark to open a metting window. There will be a forced front
and small meeting window in the upper right corner. It is meaningless to switch to this window.

lwouis: `isActualWindow` checks if the window is reasonable to list in our switcher. It should be either a standard-window, or a dialog. It should have a title, a size bigger than a minimum size, etc.

Some apps are incorrectly coded, and they will end up having their windows either show up when they should't show up, or not show when they should show.

We've hard-coded some exceptions over the years. These days I don't want to hard-code any exception. It's on the broken apps to fix themselves. It's not on AltTab to add magic code to repair specific apps. Furthermore, these apps change quickly, so our workarounds may break things later.

It's definitely on the broken apps to fix themselves. Their broken code affects AltTab and it also affects other accessibility apps.

Here this lark is most likely having an issue. I suggest you report it to their support, so they get it fixed.

chenxiankong: The fact is that lark does not adapt to this software, but as a developer, I can understand your considerations. I accept it.

lwouis:
This PR has been mainstreamed a while ago with 541337259db771853b2175d4faa93dccfe4c1506. The part with screen detection. The part of apps with similar bundle IDs won't be mainstreamed because apps with different bundleIDs are different apps. We actually don't want to treat them as the same app. We would get all sorts of issues like how to quit them, etc.

Closing this PR 👍

PR #4291, opened 2025-03-01
[WIP] Fully unfocus windows when they're switched away from

zacharee:
There's some weirdness with the behavior of switching windows where if you switch from window A to window B, window B won't fully unfocus. For example, Discord or Slack will continue blinking the text cursor in the message box and will also continue to animate emojis/GIFs/etc. This is a problem because it also means the app automatically marks new messages in the current channel as read, meaning notifications don't come through.

I messed around a bunch with the known private APIs without really knowing what I'm doing and I think I found a solution that works. I haven't fully tested it, though, and it also seems to break AltTab's window ordering in the switcher. I tried to mitigate it by comparing window IDs when receiving accessibility events, but I couldn't figure that part out.

This PR isn't ready to merge as-is, but I want to share it to get some input and see if the window ordering is an easy fix.


alt-tab-macos.xcodeproj/project.pbxproj
zacharee, line 1: I'll remove this before any merging happens.

src/logic/Window.swift
```diff
@@ -184,16 +184,29 @@ class Window {
...
                 }
             }
         }
     }
 
+    func unfocus() {
+        Logger.info("Unfocusing \(application.bundleURL.map { "\($0)" } ?? "nil") \(cgWindowId.map { "\($0)" } ?? "nil")")
+        var psn = ProcessSerialNumber()
+        GetProcessForPID(self.application.pid, &psn)
+        self.deactivateWindow(&psn)
```
zacharee, line 207: This is what seems to be causing the window ordering to break. I'm guessing it causes a focus event to be sent, but I'm not sure how to prevent AltTab from handling it.

zacharee:
After a bunch more experimentation, it looks like this only works if AltTab's preferences window is open. It doesn't need to be on top, just open. I'm wondering if it's possible to make an invisible window to persistently show, and I'll try that out tomorrow.

lwouis:
Hi,

I think this may not be the right direction to look into.

I think that macOS model is that there should only be 1 focus at a time. For example, if 2 input fields could be selected at the same time, hitting the keyboard would fill them both, which is probably not a good thing. Now there are exceptions and tricks of course. For instance "mouse focus" and "keyboard focus" can be somehow different. And a window can receive events without being in focus. Overall though, there should be one focus, I think, in the macOS model.

Thus, in macOS, when you focus a window, macOS automatically unfocuses the previous window which was focused. AltTab relies on this mecanism as well. We focus the window, and it should de-focus previous windows.

> Discord or Slack will continue blinking the text cursor in the message box and will also continue to animate emojis/GIFs/etc. 

You mention that focus is buggy after focusing a window with AltTab. The previous Slack window continues to have keyboard focus. I tested with Slack, and couldn't reproduce. It works correctly for me.

I suggest you first open a ticket with step-by-step and a video to discuss the issue in the first place. There is a general ticket for focus issues here: #4192. However, I don't think anyone mentioned previous focus remaining. Most cases are people saying the target window is not properly focused.

Regarding the PR's content, I think it would be better to find a way to improve the focus method, so that it goes through the right internal paths, in macOS, and macOS defocuses the previous window on its own. Trying to unfocus then focus seems like a way more complex path, with many opportunity to create un-natural behaviors and surprises.

Thank you 🙇

zacharee:
I'll open an issue with a video recording showing the issue. "Keyboard focus" might not be the best way to describe it.

lwouis:
Closing as it hasn't been active for a long time

PR #4534, opened 2025-06-06
Per-Shortcut “Show in Switcher” Configuration for a More Flexible Alt-Tab Experience

bartier:
## Summary

Enable per-shortcut “Show in switcher” settings so each AltTab binding can independently show either applications or windows.

## Motivation

- Previously, a single global toggle controlled whether the switcher displayed windows or applications.  
- Users expect macOS-style behavior: for example, in Google Chrome you may want ⌘ Tab to cycle through all running apps, but ⌘ ` to cycle only through Chrome’s open windows.  
- Without per-shortcut configuration, pressing ⌘ ` still showed every app instead of Chrome windows, forcing manual filtering.  
- This change restores that flexibility by allowing each of the three configurable shortcuts to choose its own mode (applications vs. windows).

## Implementation Notes

- **⚠️ Not a Swift developer**  
  I used AI assistance to develop and test these changes. If maintainers want more context or provide me help to merging upstream, I’m happy to assist with any follow-up reviews or refinements.

### DIFF
--- a/src/logic/Preferences.swift
+++ b/src/logic/Preferences.swift
@@ -51,7 +51,10 @@ class Preferences {
         "showOnScreen": ShowOnScreenPreference.active.indexAsString,
         "titleTruncation": TitleTruncationPreference.end.indexAsString,
         "alignThumbnails": AlignThumbnailsPreference.center.indexAsString,
-        "showAppsOrWindows": ShowAppsOrWindowsPreference.windows.indexAsString,
+        "showAppsOrWindows": ShowAppsOrWindowsPreference.applications.indexAsString,
+        "showAppsOrWindows2": ShowAppsOrWindowsPreference.windows.indexAsString,
+        "showAppsOrWindows3": ShowAppsOrWindowsPreference.windows.indexAsString,
+        "showAppsOrWindows4": ShowAppsOrWindowsPreference.windows.indexAsString,
         "showTitles": ShowTitlesPreference.windowTitle.indexAsString,
         "appsToShow": AppsToShowPreference.all.indexAsString,
         "appsToShow2": AppsToShowPreference.active.indexAsString,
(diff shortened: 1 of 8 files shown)


lwouis:
Hi,

The `Show in Switcher` feature is not available in Thumbnails Style. I tested the behavior on this branch and the UI is still shown, and letting me pick `Applications` or `Windows` even in Thumbnails Style. When picking `Applications`, it doesn't work.

A bigger issue is that I'm not sure this setting should be per-shortcut. Many people ask to move many global settings, per-shortcut. If we would do what everyone wants for their work-flow, every setting would be per-shortcut, and the UI would be tough UX for users.

We need to balance things, and only have per-shortcut settings which are actually used by most people. Here, based on previously received feedback, I don't think it's the case.

I'm afraid I can't merge this MR for this reason. I hope you can understand 🙇 

I think #337 may be a better outlet to offer a workflow that helps navigate between apps and windows. It would enable apps/windows navigation without a new per-shortcut setting.

bartier:
Hi @lwouis, thank you for the quick response to the PR , I really appreciate it. :)

I completely understand your reasoning behind not merging the change, even though I personally disagree in terms of the added complexity (I’m definitely biased since I use the workflow I’m proposing).

I’ve read through issue [#337](https://github.com/lwouis/alt-tab-macos/issues/337), and I agree that the ideas discussed there align well with your philosophy. That said, I also recognize the significant effort involved, and I wouldn’t be able to commit to working on a draft myself. :(

I’ll go ahead and close the PR and continue maintaining my own fork, keeping it in sync with upstream where possible. It works well enough for my use case, and I’m happy with that.

Thanks again for maintaining such a great project. Wishing you all the best with future development!

lwouis:
Thank you @bartier 
Feel free to keep the PR opened. It may make your fork more visible to other people. I don't mind having open PRs that won't be merge. I find it's a way to advertise forks that can't be merged but are still valuable to people 👍

Issue #3117, opened 2024-01-19
After opening AltTab, it seems like the Next Window shortcut is held or pressed a few times without the user doing anything


SamadiPour:
**Describe the bug**

Since a few weeks ago, I have been experiencing drifting when the mouse is hovered over an item. It only happens when the mouse is over an item, and not when it's inside or outside of the frame.
When I hold down the ﻿`Command` key and press `﻿Tab` once (push and release instantly), the selector drifts through the items, sometimes until it reaches the mouse and sometimes until it reaches the end.
The `Mouse hover` option under `Also select windows using` is disabled.

**Screenshots / video**

https://github.com/lwouis/alt-tab-macos/assets/24422125/44cec264-affb-4058-b880-d87e60a9a14a

**Steps to reproduce the bug**

**Your environment**

* AltTab version: 6.64.0
* macOS version: 14.2.1

warrenseine:
I experience the same issue with the same environment. I reverted to 6.63.0 and the weird behavior is gone.

I suspect this is related to [this commit](https://github.com/lwouis/alt-tab-macos/commit/3b0194d886fe31813fda6d319d5b743491b61b98).

lwouis:
Thank you @warrenseine for sharing this message.

@SamadiPour does v6.63.0 also fix the issue for you?

I don't see how the issue would be related to the commit though.

@SamadiPour on your video, I see some flickering on the right part of AltTab's window. I'm wondering if you are not suffering from #1840. Could you please show the top menubar of your screen, when the issue happens? If there are 2 AltTab icons, that's the source of the issue right there.

SamadiPour:
@lwouis Just tried v6.63.0, and it seems like I can't see the issue anymore. I will test more to make sure.

Also, flickering might be because of video compression or the recorder. I didn't see any flicker while using it. I also checked the processes and top menubar, and there was only one instance running.

ankushagarwal:
I can confirm that this happens to me as well. I am on the app version 6.64.0. I am on the latest macOS Sonoma version 14.2.1

https://www.loom.com/share/76a2006f084b4b52880ec7a25879958e

ythosa:
the same problem

esserrge:
I had this problem on Sonoma 14.2, does not seem to happen after upgrade to 14.3. Running alttab 6.64.0 on both versions.

agmitron:
The same here

Warkanlock:
[...]

PR #4572, opened 2025-06-24
fix: fix focused window drift

Allsochen:
try to fix the #3117

### DIFF
--- a/src/ui/App.swift
+++ b/src/ui/App.swift
@@ -31,11 +31,9 @@ class App: AppCenterApplication {
     private var appCenterDelegate: AppCenterCrash?
     // don't queue multiple delayed rebuildUi() calls
     private var delayedDisplayScheduled = 0
-    // Atomic operation protection to prevent concurrent calls
-    private var isExecutingCycling = false
-    private let cyclingLock = NSLock()
     private var firstCallTimestamp: Date?
     private var isInitializing = false
+    private var lastKeyRepeatCallTime: Date?
 
     override init() {
         super.init()
@@ -81,6 +79,7 @@ class App: AppCenterApplication {
         isFirstSummon = true
         isInitializing = false // Clear initialization flag
         forceDoNothingOnRelease = false
+        lastKeyRepeatCallTime = nil // Reset key repeat call tracking
         MouseEvents.toggle(false)
         hideThumbnailPanelWithoutChangingKeyWindow()
         if !keepPreview {
@@ -251,16 +250,6 @@ class App: AppCenterApplication {
     }
     
     func showUiOrCycleSelectionWithSource(_ shortcutIndex: Int, _ forceDoNothingOnRelease_: Bool, _ eventSource: EventSource) {
-        // Atomic operation protection - prevent concurrent calls
-        cyclingLock.lock()
-        defer { cyclingLock.unlock() }
-        
-        // Early exit if concurrent execution protection is active
-        if isExecutingCycling {
-            Logger.debug("showUiOrCycleSelection blocked by concurrent protection")
-            return
-        }
-        
         // Initialize basic state
         forceDoNothingOnRelease = forceDoNothingOnRelease_
         Logger.debug(shortcutIndex, self.shortcutIndex, isFirstSummon)
@@ -283,9 +272,6 @@ class App: AppCenterApplication {
         if isFirstCall {
             firstCallTimestamp = now
         }
-        
-        // Set concurrent execution protection
-        setConcurrentProtection()
         
         // Execute the appropriate action based on whether this is first call or cycling
         if isFirstCall {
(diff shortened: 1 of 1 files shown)


src/logic/Windows.swift
```diff
@@ -121,6 +121,18 @@ class Windows {
         return index
     }
 
+    /// Get the index of the second window that should be shown to the user
+    /// This is the standard behavior of switcher - select the second window (the first is usually the currently active window)
```
lwouis, line 125: In some cases, the selected thumbnail is not the second one. For example, if there are no window at all, or if the the current active app has no open window, or maybe I think in some cases where the sorting order is not recently-used-first.

Is this taken into account with this MR?

src/ui/App.swift
```diff
@@ -233,33 +237,169 @@ class App: AppCenterApplication {
     }
 
     func showUiOrCycleSelection(_ shortcutIndex: Int, _ forceDoNothingOnRelease_: Bool) {
+        showUiOrCycleSelectionWithSource(shortcutIndex, forceDoNothingOnRelease_, .unknown)
+    }
+    
+    func showUiOrCycleSelectionWithSource(_ shortcutIndex: Int, _ forceDoNothingOnRelease_: Bool, _ eventSource: EventSource) {
+        // Atomic operation protection - prevent concurrent calls
```
lwouis, line 244: I don't think it's possible to have concurrent calls of this method. It should always be called from `DispatchQueue.main`. Calls should be sequentials.

There are cases of concurrency in the apps. There are always main queue + another queue. Once we get executed on the main queue, concurrency issues should be gone since that code should always be called from `DispatchQueue.main`, not from other queues. If it is called from other queues, it's a bug and should be fixed.

src/ui/App.swift
```diff
@@ -233,33 +237,169 @@ class App: AppCenterApplication {
...
+        // Execute the appropriate action based on whether this is first call or cycling
+        if isFirstCall {
+            handleFirstCall(shortcutIndex: shortcutIndex)
+        } else {
+            handleCycling()
+        }
+    }
+    
+    /// Check if a duplicate event should be blocked based on event source and timing
+    private func shouldBlockDuplicateEvent(eventSource: EventSource, currentTime: Date) -> Bool {
```
lwouis, line 289: I think that logging the event source is very useful!

I'm worried about the way we fix the issue here by trying to avoid multiple events from being processed. I think it's more desirable to avoid multiple events being generated in the first place, instead of blocking them later on

Allsochen: I agree with your point, but based on the logs(https://github.com/lwouis/alt-tab-macos/issues/3117#issuecomment-2999453614) we've tracked so far, it might be an issue with macOS system event notifications. Therefore, I suggest temporarily fixing the problem by avoiding it to provide users with a better experience. Maybe one day we'll identify the root cause, and then it won't be too late to remove the logic code here.

src/logic/Windows.swift
Allsochen: When there is no window, returning nil can handle it correctly. This fix is not related to that. Initially, I thought it was the issue and added the code, which should be rolled back.

src/ui/App.swift
lwouis: In these logs, it's not clear to me what happens. What is the source of calls to `showUiOrCycleSelection`?

In my investigation in #3117 I found one root issue: the OS seems to randomly absorb some keyboard shortcuts. Thus, the user will alt+tab, and release the tab key. At this point, AltTab doesn't get the event for the alt+tab shortcut being Up (since the tab key was released).

This explains the issue. It's tricky to deal with because I'm not sure how we can detect and workaround it.

Could you please try to describe what the fix in this MR does at a high level? I'm struggling a bit to understand the core idea of this fix. It seems to ignore some cycling events, based on the source of the event. Could you please try to explain the big idea at high-level?

Thank you!

Allsochen: https://github.com/lwouis/alt-tab-macos/issues/3117#issuecomment-2999453614 For this log, my action was simply pressing the Alt+Tab keys, then releasing the Tab key. However, the system received continuous event triggers, which was not my actual operation. Therefore, the fix here is to merge these events based on different event sources to reduce erroneous event triggers.

lwouis: Thank you for your message

If I understand correctly, you say that your issue is not the issue with Secure Input from https://github.com/lwouis/alt-tab-macos/issues/3117. You face a different issue.

In your case, you notice unwanted key repeat events. Are you sure that this is abnormal? You have [shown here very fast key repeat settings](https://github.com/lwouis/alt-tab-macos/issues/3117#issuecomment-3011175161). If you set the key to repeat quickly, after almost no delay, then it will happen. You will see key repeat events in all apps including AltTab.

We can't ignore key repeat events in AltTab. People want the key to repeat. For example, they hold `tab`, and want the selection to quickly shift to the right until they release `tab`

This seems like normal key repeat behavior. Do you observe something abnormal on your side?

Thank you

lwouis:
I think this PR is likely stale. We've changed how external events are processed quit a bit since Summer 2025. Hopefully these issues are all gone now on recent builds. Closing the PR 👍

PR #4581, opened 2025-06-25
fix: add applescript fallback for focus issues with Electron apps

DChiuch:
a fix attempt for #4192


lwouis:
Subsumed by #5149 and related PRs

PR #4591, opened 2025-06-29
feat: add window index hotkeys

danieldegrasse:
Add window index hotkeys. These hotkeys are mapped to 0-9 by default,
and allow the user to jump to a window in the alttab list by its index.
This is intended as a faster alternative to selecting the window with a
mouse.

Examples of what this looks like:

image

image

The square boxes with numbers indicate the shortcut key that will jump to the window. For example, pressing "4" while alttab is in focus would switch focus to the terminal window, or "1" would jump to the alttab preferences. These keys can be remapped, and up to 10 jump keys (mapped by default to number keys 1-0) are supported.

If this feature is too intrusive to enable by default, I'm happy to make it optional- currently it is enabled by default


lwouis:
Hi,

I think there are a few issues to iron out before we can consider merging the PR:

* The preferences UI is too tall. We would probably need to have a dedicated pannel for this, if we want to customize so many shortcuts.
image

* The UI indicators for shortcuts are a bit confusing, when seen next to Space indicators. When we see a circle with 1 inside, and a square with 1 inside, it's not clear what those mean. I think a different way is needed here. Something that evokes keyboard shortcuts. Also, the numbers are fixed today. If I set window 1 to shortcut `D`, it will not show `D` on that window. It will still show `[1]`.
* I'm not sure that everyone will want this feature. I think it probably would have to be disabled by default, which would hide the UI indicators. Then for users who activate it, it would show UI indicators, and recognize number shortcuts.

What do you think?

Thank you 🙇

danieldegrasse:
> Hi,
> 
> Thank you for sharing this PR.
> 
> I think there are a few issues to iron out before we can consider merging the PR:
> 
>     * The preferences UI is too tall. We would probably need to have a dedicated pannel for this, if we want to customize so many shortcuts.
> 
> 
> image * The UI indicators for shortcuts are a bit confusing, when seen next to Space indicators. When we see a circle with 1 inside, and a square with 1 inside, it's not clear what those mean. I think a different way is needed here. Something that evokes keyboard shortcuts. Also, the numbers are fixed today. If I set window 1 to shortcut `D`, it will not show `D` on that window. It will still show `[1]`. * I'm not sure that everyone will want this feature. I think it probably would have to be disabled by default, which would hide the UI indicators. Then for users who activate it, it would show UI indicators, and recognize number shortcuts.
> 
> What do you think?
> 
> Thank you 🙇

Hey- thanks for the review (and for creating this project, alttab is a lifesaver) - all very reasonable points. I agree making the feature opt-in is probably best.

Regarding the UI indicators with number shortcuts, the issue we run into is that (AFAIK) not every shortcut key has an associated SF symbol in the font, does it? One solution would be to not support configuring the shortcut keys, and fix them to 0-9. What are your thoughts on that approach? I'm not sure I like removing configurablity, but I'm also not sure how else to cleanly support the UI indicators.

lwouis:
Thank you for your message

Indeed, I think hardcoding 0-9 would be good enhancement. Customization could be discuss later, if people ask for it. So, to recap:

* The feature would be disabled by default
* It would be activated in Preferences > Controls > Additonal controls... > `Select windows with number keys 0 to 9`
* If the feature is active and a user binds another shortcut to 0-9, we would need to handle conflict resolution. This is done already for the `Select windows using vim keys`, so you can copy what was done there
* What about the UI? The square numbers are confusing next to the Space numbers. Maybe no UI at all? Maybe people who go in the sub-setting will be power-users and will know that hitting the number will switch the window, without the need for UI reminders?

What do you think?

Thank you

danieldegrasse:
Hi, sorry for the delay getting back to this- latest push is just a rebase so no need to review.

> Thank you for your message
> 
> Indeed, I think hardcoding 0-9 would be good enhancement. Customization could be discuss later, if people ask for it. So, to recap:
> 
>     * The feature would be disabled by default
> 
>     * It would be activated in Preferences > Controls > Additonal controls... > `Select windows with number keys 0 to 9`
> 
>     * If the feature is active and a user binds another shortcut to 0-9, we would need to handle conflict resolution. This is done already for the `Select windows using vim keys`, so you can copy what was done there
> 
>     * What about the UI? The square numbers are confusing next to the Space numbers. Maybe no UI at all? Maybe people who go in the sub-setting will be power-users and will know that hitting the number will switch the window, without the need for UI reminders?
> 
> 
> What do you think?
> 
> Thank you

I think this all makes sense, but I do think we should place some form of indicator- truthfully, I can't count windows that quickly :)

Perhaps the indicator is placed on the right instead of the left like so?

image

lwouis:
I think we need some idea and mocking here for the UI. The numbers should be clear that they are shortcuts, especially when the user is using multiple Spaces and those numbers are already showing.

I think this could be brainstormed/mocked using ChaptGPT for example. To get a few candidate UIs before doing implementation

Regarding shortcut conflicts, i've added unit-tests, so please add a few for this new scenario 👍

gabrielbacha:
+1 for this feature.. I would consider adding the numbers on the left without any squares or circles (as Daniel had it)

videni:
@lwouis  Could  you  merge this PR please? we need this feature badly.

Issue #4520, opened 2025-05-29
Alt-Tab doesn't switch windows when one of running applications is hanging


nerzhulart:
**Describe the bug**

If one of my running apps is hanging Alt-Tab stops to focus a selected app. Alt-Tab switcher is showing but when I choose any app (even not hanging) nothing just happens.

Then if the application is unfrozen it looks like Alt-Tab or MacOS quickly pumps all the scheduled switch events and I see all my apps switching is blinking fast.

**Steps to reproduce the bug**

It's also reproducible when debugging some Java UI app and the debugged app is stopped on a breakpoint.

You may reproduce it in the following way:
1. install IJ Idea community (see Community here https://www.jetbrains.com/idea/download/?section=mac )
2. Clone this sample repo https://github.com/JetBrains/build-UI-using-Swing-sample-project
3. Open it in Idea, set up JDK (File->Project Structure, then Project->SDK->Download and here choose 20 and Coretto)
<img width="1379" alt="Image" src="https://github.com/user-attachments/assets/766c7c7e-08ea-4ae9-9715-6d105813d97d" />

4. Set a breakpoint inside Main.kt:20  

<img width="1396" alt="Image" src="https://github.com/user-attachments/assets/10a275e8-03f7-48ae-8bf8-6c01d360ddcf" />
 
5. Debug app using the green Run triangle to the left of Main method 

<img width="530" alt="Image" src="https://github.com/user-attachments/assets/180f07e1-e0ad-4e18-9658-3da47562ff04" />

6. When the app is running click Save inside, you will stop on the breakpoint
7. Try to switch windows using Alt-Tab - the windows can't be focused anymore while the app is stopped on the breakpoint
* AltTab version: 7.24
* macOS version:  15.5

nerzhulart:
> Hi,
> 
> AltTab calls the macOS window focus API. The behavior you describe of kind of queueing events when the app becomes responsive again is how macOS is designed. I'm afraid we can't change this. The unresponsive app has its main loop stuck. Once the main loop is free again, it will receive macOS events such as the request to focus its window. The app will receive accumulated events at this point.
> 
> I'm afraid this behavior is independant of AltTab and would happen with other methods since it's OS-wide.
> 
> Thank you

thanks for the response! but there is no such issue with MacOS embedded switcher. Maybe you have any ideas where to have a look? Theoretically I can try to debug by myself, but I have no experience in Macos/Swift development

lwouis:
[...]

Issue #4405, opened 2025-04-02
AltTab missing applications on startup, requires relaunching app to fix


ktang16:
**Describe the bug**

Recently, AltTab is missing applications that are running when I restart my Mac. I see the applications, but AltTab won't display them / allow me to switch to them. I have to close and reopen AltTab for it to show the open windows

**Screenshots / video**

**Steps to reproduce the bug**

1. Restart Mac, have it reopen applications on restart
2. Use AltTab shortcut
3. See error (some apps like Mail won't display using AltTab even though they reopen)

**Your environment**

* AltTab version: 7.22
* macOS version: 15.4
* Other relevant info: using external keyboard, multiple monitors (although issue persists on one monitor)

ktang16:
Here's a screen recording after I restarted my computer. I've set AltTab to switch between only open windows on the same desktop. At the beginning it won't switch to calendar even though it's open. Only after clicking on calendar will it show up within AltTab.

This behavior goes away after I restart AltTab

ktang16:
Here's a screenshot after I restarted my computer. I've set AltTab to switch between only open windows on the same desktop. At the beginning it won't switch to calendar even though it's open. Only after clicking on calendar will it show up within AltTab.

This behavior goes away after I restart AltTab

![Image](https://github.com/user-attachments/assets/2dee8c6d-d623-4109-9cf4-a4489592d763)

![Image](https://github.com/user-attachments/assets/11e5ba87-5dd5-4ae2-bf57-ee270b141ef8)

lwouis:
Thank you @ktang16 
In your screenshot all 3 apps are on the same Space, and all 3 apps are not fullscreen, correct?
Thank you

ktang16:
Yeah, all 3 apps are in the same space and none of them are fullscreen

smallironman666:
我也遇到了同样的问题，也是应用程序丢失，但是我用command+tab可以切换出来那些在alttab里面没有显示的应用

tylerlaprade:
I encounter this with VSCode in fullscreen

SovaPluto:
Similar issue happens to me constantly with "Zoom" window and "BetterTouchTool" app.
The application/window disappears from the list until I quit & reopen AltTab.

Lawrence-of-AnKing:
I often meet this after a long sleep, some app just lost from AltTab, I have to reboot AltTab.

eatkid:
same question

opus131:
[...]

PR #4669, opened 2025-08-13
Alt-Tab would become unresponsive when trying to switch to a window of a frozen application 

nerzhulart:
Hi! Quite recently I created an issue where Alt-Tab doesn't activate windows in some cases (see the issue and below). https://github.com/lwouis/alt-tab-macos/issues/4520
I'm not a swift developer at all, so this change and the description below is made by Claude. But as I checked on my machine this change fixes the issue

By Claude:

Problem

  Alt-Tab would become unresponsive when trying to switch to a window of a frozen application or any other window if some app is frozen (e.g., a UI app paused at a debugger breakpoint). The window switcher UI would appear, but selecting windows
  wouldn't work. Once the frozen app was resumed, all queued window switches would execute at once.

  Root Cause

  The issue occurred because Alt-Tab uses synchronous IPC calls to focus windows:
  - _SLPSSetFrontProcessWithOptions() - Window Server API
  - SLPSPostEventRecordTo() - sends events to target process
  - AXUIElementPerformAction() - Accessibility API

  These calls block until the target process responds. When the target is frozen, they wait indefinitely.

  Additionally, all window operations were executed on a serial queue (accessibilityCommandsQueue), meaning one blocked operation would prevent all subsequent operations from executing.

  Solution

  1. Changed the queue from serial to concurrent - allows multiple window operations to run in parallel
  2. Added 200ms timeout to all window operations - wraps blocking calls in DispatchWorkItem that gets cancelled if it takes too long
  3. Applied timeouts consistently - to focus, close, minimize, and fullscreen operations

  Now when switching to a frozen app, that specific operation times out after 200ms, but other window switches continue to work normally. This matches the behavior of macOS's native Command+Tab switcher.

  Testing

  1. Run a UI app (Swing/Compose/etc.) with a debugger attached
  2. Set a breakpoint and trigger it to freeze the app
  3. Use Alt-Tab to switch between windows
  4. Verify that switching to other (non-frozen) windows works normally
  5. The frozen app's switch will timeout but won't block other operations

### DIFF
--- a/src/logic/BackgroundWork.swift
+++ b/src/logic/BackgroundWork.swift
@@ -20,7 +20,8 @@ class BackgroundWork {
         // >Processes reached dispatch thread soft limit (64)
         screenshotsQueue = DispatchQueue.queue("screenshotsQueue", .userInteractive, false)
         // calls to act on windows (e.g. AXUIElementSetAttributeValue, AXUIElementPerformAction) are done off the main thread
-        accessibilityCommandsQueue = DispatchQueue.queue("accessibilityCommandsQueue", .userInteractive, false)
+        // Using concurrent queue to prevent one frozen app from blocking all window switching operations
+        accessibilityCommandsQueue = DispatchQueue.queue("accessibilityCommandsQueue", .userInteractive, true)
         // calls to the AX APIs can block for a long time (e.g. if an app is unresponsive)
         // We can't use a serial queue. We use the global concurrent queue
         axCallsQueue = DispatchQueue.queue("axCallsQueue", .userInteractive, true)
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -117,11 +117,20 @@ class Window {
         }
         BackgroundWork.accessibilityCommandsQueue.async { [weak self] in
             guard let self else { return }
-            if self.isFullscreen {
-                self.axUiElement!.setAttribute(kAXFullscreenAttribute, false)
+            
+            let workItem = DispatchWorkItem { [weak self] in
+                guard let self else { return }
+                if self.isFullscreen {
+                    self.axUiElement!.setAttribute(kAXFullscreenAttribute, false)
+                }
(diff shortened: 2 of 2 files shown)


lwouis:
Thank you @nerzhulart for sharing this work!

I was able to reproduce. I don't think the app becomes unresponsive though. I think it's better to say that once an attempt is made to focus the frozen app, further attempts focus anything would fail. Which is bad!

In the PR, think changes to the second file are questionable. The boolean switch to `true` is enough. It's an interesting idea to try and prevent spawning too many threads though. I'm looking at using OperationQueue instead.

I'll close this PR for now, and cherry-pick it. It will be in the next release.

Thank you very much!

nerzhulart:
> Thank you @nerzhulart for sharing this work!
> 
> I was able to reproduce. I don't think the app becomes unresponsive though. I think it's better to say that once an attempt is made to focus the frozen app, further attempts focus anything would fail. Which is bad!
> 
> In the PR, think changes to the second file are questionable. The boolean switch to `true` is enough. It's an interesting idea to try and prevent spawning too many threads though. I'm looking at using OperationQueue instead.
> 
> I'll close this PR for now, and cherry-pick it. It will be in the next release.
> 
> Thank you very much!

Thanks for the reply! Is the fix presented in the latest 7.27 update? Should I migrate to the new version instead of using my local build?

lwouis:
@nerzhulart it's not been released yet. I've packed a few improvements for the next release. However, since there are many big changes, I'm worried to release it and I'm seeking people to help me QA. [You can help test the build here](https://github.com/lwouis/alt-tab-macos/issues/4405#issuecomment-3216548542) 👍 

Your PR and the resulting time I spent working on this has led to a large rewrite of background work. It's a really good thing. At the same time it introduces more risk than a small-improvement release, so I'm taking time and being cautious here. I always want to avoid bad experiences for users after an update. People time is very precious!

PR #4686, opened 2025-08-18
adding screen number indicator

sannysanoff:
Use case: I have terminal windows on multiple screens, most of them secondary.

When I press alt+tab, i would like to see if focus will be driven away out of my screen, in other words, which terminal will i tab to. I rarely need to switch to the different windows, but often accidentally do, and because windows are brought up the stack, this confuses long after that. I render numbers only when app is on different, from current, screen.

Anyway, this is just adding awareness. I'll attach the screenshot: 

image

I would like this feature to be included in the main branch, but will not be offended if not. Please free to further modify as you wish, if you need.

Also I added the toggle in the settings to enable it (disabled by default): 
image

Cheers!

### DIFF
--- a/resources/l10n/Localizable.strings
+++ b/resources/l10n/Localizable.strings
@@ -585,3 +585,4 @@
 
 /* No comment provided by engineer. */
 "You may need to disable some conflicting system gestures" = "You may need to disable some conflicting system gestures";
+"Show screen numbers" = "Show screen numbers";
--- a/resources/l10n/ar.lproj/Localizable.strings
+++ b/resources/l10n/ar.lproj/Localizable.strings
@@ -246,3 +246,4 @@
 "You can’t undo this action." = "لا يمكنك التراجع عن هذا الإجراء.";
 "You didn’t write your email, thus can’t receive any response." = "لم تكتب بريدك؛ فلذلك لن تحصل على رد.";
 "You may need to disable some conflicting system gestures" = "You may need to disable some conflicting system gestures";
+"Show screen numbers" = "إظهار أرقام الشاشة";
--- a/resources/l10n/bg.lproj/Localizable.strings
+++ b/resources/l10n/bg.lproj/Localizable.strings
@@ -246,3 +246,4 @@
 "You can’t undo this action." = "Връщането на това действие не е възможно";
 "You didn’t write your email, thus can’t receive any response." = "Няма да получиш отговор, защото не си написал събщение.";
 "You may need to disable some conflicting system gestures" = "You may need to disable some conflicting system gestures";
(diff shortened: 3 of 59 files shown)


sannysanoff:
I will also add localization!

lwouis:
> I will also add localization!

Localizations are updated trough poeditor. We can't use Github for this.
Could you please use [this link](https://poeditor.com/join/project/TTqkcdPO9V) to update them?

Thank you!

lwouis:
Hi,

This is an interesting addition. I see a few UX issues on the screenshots:
* We need a way to differentiate/clarify between Screen numbers, and Space numbers. If the user activates both, they will see (1) and (3) on the same tile. Which is which? I think it could get very confusing
* Are screen numbers clear on their own? I'm not sure users know what numbers their screen are assigned. They are likely to think "my Dell screen vs my Asus screen" or "my external screen vs my macbook internal screen" or think of the screen ratio "my vertical screen vs my horizontal screen". Perhaps it would help to show: an icon, the ratio number (e.g. `16/10`), the resolution (e.g. `1900x1200`), or the name/brand of the screen (e.g. `Internal screen`, `Dell 1220`, `Asus`)

What do you think?

Thank you

sannysanoff:
Hi, thanks for your reply!

tbh, I do not use workspaces, so I was not aware of other numbers - I will play with it, and improve, and update.

Screen numbers are assigned from left to right by X coordinate of screen arrangement, from "1" to whatever. So leftmost screen is "1". 

I will think about your ideas with screen sizes/names/icons

Thanks,
Alex.

sannysanoff:
I added a workspace, but application currently does not identify apps from different workspaces with any numbers. Maybe I missed some setting? Or what did you mean when said "will see (1) and (3)" ?

sannysanoff:
Also, while it's possible to identify monitors by size / ratio, imagine user has 2 similar external monitors. How to identify them? I have one external monitor, another on 2nd macbook via airdrop, and third is built-in. For those, I can come with some solution. 

Do you want to add settings (probably not)? Should there be special icon/identification sets for cases when

1) 1 internal + 1 external [ + 1 airdrop ] - can use icons
2) 1 internal + N external of different ratios (can use icons + ratios)
3) 1 internal + N airdrops + M same externals vertical + L same externals another ratio - can use icons + numbers
4) 1 internal + External Samsung + External LG (same ratio) - should use icons + 3 letters of names (if different).

📺 SAM(3)    <- external monitor samsung number 3 - like that.

lwouis:
@sannysanoff I think it would help to benchmark how other apps handle this. Apps that specialize around multiple monitors for instance. What UX do they use to indicate the various displays?

PR #4687, opened 2025-08-18
complete mouse disabler

sannysanoff:
Use case: you pressed alt-tab with your left hand, aiming at certain app, and then you touched mouse with right hand.
Outcome: two focused applications with two similarly blue selection rectangles.
Reaction: WTF? which one?
Correct behavior: accessibility fix: disable mouse completely for people never using mouse when alt-tabbing.

I implemented it with the saved settings compatibility in mind, so it understands saved option. Option looks like this:

image

Localization is added.

I will not be offended if you're not adding that to master, but plz drop me a note maybe I could better align with requirements. 

Thanks!


lwouis:
Hi,

Thank you for sharing this!

The issue of unintentional mouse interaction has been discussed a lot in the past here: https://github.com/lwouis/alt-tab-macos/issues/286

Overall, I think the current setting to disable mouse hover is enough. Adding more granular control to distinguish mouse hover from mouse clicks is too niche I'm afraid. Most users won't engage with that settings, and it will actually disuade users to engage with the existing mouse hover feature since it would be more complicated.

With v7, we decided to reduce the number of settings to focus on the ones that most users would interact with. Our goal for AltTab is to be simple and flexible enough. The price to pay for that is fewer advanced workflows and controls.

Thank you 🙇

PR #4734, opened 2025-09-06
feat: cursorFollowFocus additional behavior customization

richardso21:
The existing “Cursor follows focus” is already nifty for me when working with multiple, large screens connected to my machine. When I switch focus between different screens, I often need my mouse to be there as well, such as switching spaces (since macOS switches spaces for the screen your cursor is at, not the currently-focused window).

However, switching windows on the same screen with this option on can be jarring in my opinion. So, **I've added a layer of customization for this setting, most notably to change cursor position only if the new window to be focused is on a different screen of the current**.

### DIFF
--- a/src/logic/MacroPreferences.swift
+++ b/src/logic/MacroPreferences.swift
@@ -435,6 +435,18 @@ enum ShowAppsOrWindowsPreference: CaseIterable, MacroPreference {
     }
 }
 
+enum CursorFollowFocusBehavior: CaseIterable, MacroPreference {
+    case always
+    case differentScreen
+
+    var localizedString: LocalizedString {
+        switch self {
+            case .always: return NSLocalizedString("Always", comment: "")
+            case .differentScreen: return NSLocalizedString("Only on different screen", comment: "")
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


lwouis:
Hi,

Thank you for sharing this.

I've made a couple of changes:
* Removed the toggle. The Dropdown can handle the disabled state by addind a `Never` option
* Fixed the logic to follow cursor or not. It was `window.isOnScreen(NSScreen.preferred)` which is not correct. I changed it so it checks if the window-to-focus is on a different screen than the currently-focused-window. I think that's what makes sense here, right?
* Added migration code to move existing users from the old toggle to the new dropdown

Could you please try out [this local build](https://github.com/user-attachments/files/22204707/AltTab.app.zip) and let me know if you're happy with the feature?

Thank you

richardso21:
Was able to try out your local build, so far so good! I don't touch a lot of Swift, let alone development for macOS apps, so I appreciate you helping out with the tweaks!

For the reasoning behind `window.isOnScreen(NSScreen.preferred)`, I assumed `NSScreen.preferred` would always be the screen containing the focus of the currently-focused window, but I'm probably wrong with that.

Issue #3882, opened 2024-11-21
Reintroduce some options removed with v7


lwouis:
Hi everyone,

When we introduced v7, the big goal was to #351. I think we achieved that. Thanks to @Allsochen's hard work, we were able to move from a page that 1% of users understood how to use, to something that looks like it was designed by Apple themselves.

Before v7 | After v7
-|-
![image](https://github.com/user-attachments/assets/b0c68336-92cd-4ead-b1da-4dcd763836de) | ![image](https://github.com/user-attachments/assets/e352964e-d413-4f9b-8559-5a6b2d842503)

We've added many new features as well. Added visual illustrations of advanced features, etc. It's a huge win overall for most users.

Along the way, we've had to remove some advanced customization. We removed it because it was too complex and a detriment for the vast majority of users. We also removed it in some cases because it was not a good idea to keep it and have to maintain it on our side, as well as to support users opening tickets with very exotic setups. Context for why manpower on our side is an important consideration: #1179 

After v7 release, many people have sent feedback that they noticed a feature was gone, and they wanted it back. This was expected. We're now looking at how many people ask for certain features to be back, and are considering reintroducing a select few. This ticket tracks this, as well as being a notice we can redirect people to.

Current candidates we are considering:

* Mouse follow focus => Added back ✅ 
* Being able to adjust the switcher width => auto-width was added. It decides the width based on ergonomic field-of-view. It should help with most cases with large screens. Not sure if we want to add more than this.
* Adding an XS and XL Appearance Size => sounds like we should add it.
* Providing a Style with App Icons and Titles, with all Titles visible at once => Maybe we could update the Thumbnails Style to be more ergonomic when the Screen Recording permission is not given? It would be a bit hidden though. Maybe it could be an option of the Thumbnails Style, in addition? ("no thumbnails"). Not sure. 
* Improve min-width in Titles Style => this one we should definitely do
* Let users use different Styles with different shortcuts => this one seems to complex to happen. Both at the UX level in the preferences (how to set this?), and at other levels. I don't think it will happen
* Number of shortcuts?

Thank you 🙇

ezhang7423:
Thank you so all much for the hard work. For myself personally, "Mouse follow focus" would be absolutely huge

TechTronicLLC:
[...]

Issue #1506, opened 2022-04-20
Window max/min width in a row: provide alternative for ultrawide monitors


singalen:
**Is your feature suggestion related to a problem? Please describe.**
I have an ultrawide monitor that I use on and off.
The switcher either too big on one of them, or too small on the other.

**Describe the solution you'd like**
Maybe an alternative way to specify a window size would help? In pixels or points, or somehow else taking screen DPI into account? Maybe take screen height into account too?

**Describe alternatives you've considered**
Only manually changing the settings each time I disconnect/reconnect the monitor.

lwouis:
I'm a bit confused with the premise. If you have a very wide monitor, that's in order to make use of this space, right? So I'm surprised you don't want AltTab to expand and make use of it? I imagine if you play a movie or a game you set it up to fill the screen, no?

Could you maybe clarify what the problem is? Screenshots would help illustrate I think

lwouis:
Yes, the answer is most likely either per-screen configuration (fancy) and/or pixel max-size (easy/practical). There is already a ticket for fixed-sized: https://github.com/lwouis/alt-tab-macos/issues/1107

Closing as duplicate

PR #4908, opened 2025-10-21
feat: add configurable panel width for Titles style

topik:
## Summary

Adds configurable panel width for the Titles appearance style to address the issue mentioned in #3882 where the UI becomes too wide on ultrawide monitors.

## Changes

  - Add automatic width optimization based on screen aspect ratio:
    - 30% for super ultrawide displays (32:9, e.g., 5120×1440)
    - 45% for regular ultrawide displays (21:9, e.g., 3440×1440)
    - 60% for standard displays (unchanged from current behavior)
    - 80% for vertical displays (unchanged)
  - Add user-configurable custom width option with slider (10-90%)
  - Add "Panel width" setting in "Customize Titles style" with Automatic/Custom modes
  - Add `isUltrawide()` and `isSuperUltrawide()` helper functions to NSScreen

 ## Testing

Tested on 5120×1440 (32:9) display where automatic mode correctly applies 30% width. Also verified on standard 16:9 display to ensure backwards compatibility. Custom slider works as expected and preferences persist correctly.

## Implementation Notes

The two-tier approach keeps absolute panel widths similar across screen types (~1500-1600px) while preventing the UI from being overwhelming on super ultrawide displays. Existing users on standard displays won't see any changes.

# Preview 
SCR-20251022-msuf

SCR-20251021-pdaw
SCR-20251021-pdfd

Edit: Forgot to include settings UI

### DIFF
--- a/src/logic/Appearance.swift
+++ b/src/logic/Appearance.swift
@@ -141,8 +141,10 @@ class Appearance {
         } else if aspectRatio >= 2.3 {
             // Regular ultrawide: Modest reduction for text readability
             maxWidthOnScreen = max(0.45, maxWidthOnScreen * 0.85)
+        } else {
+            // Normal screens: preserve original behavior
+            maxWidthOnScreen = isHorizontalScreen ? 0.6 : 0.8
         }
-        // else: use comfortableWidth value as-is for normal screens
         windowMinWidthInRow = 0.6
         windowMaxWidthInRow = 0.9
         rowsCount = 1


lwouis:
Hi,

Thank you for sharing this PR! I haven't had time to review it yet. I finally got things running again on my machine, and could work on the backlog. Reviewing this PR is next on my list 👍 

Thank you

lwouis:
I have played with the PR. I have a few questions and remarks 👍 

* There is already a mecanism for automatically setting the width of the AltTab window: https://github.com/lwouis/alt-tab-macos/blob/122dc2ec0fbdf6f7ebe14c42efebb19d48d24505/src/logic/AppearanceTestable.swift#L12. It seems you re-implemented it with a new logic of discrete ratios, versus the existing fluid logic. I think this is a big redundant. It would be better to reuse the existing system, and perhaps tweak the numbers for Titles Appearance
* Regarding the manual width slider, I'm not a fan of this. We used to have many sliders for all the settings. Then with v7, we simplified a lot. That meant removing customization away, in order to improve the UX for most users. So I don't think we should go the slider route. It's a regression in a way. In addition, there have been discussions of having a set width, instead of percent width. This would help users with multiple screens, it seems. See conversations in [#v7](https://github.com/lwouis/alt-tab-macos/issues/3882) and https://github.com/lwouis/alt-tab-macos/issues/1506

What do you think?

Thank you

topik:
Good point. Simplified the PR:

  - Reuses comfortableWidth() instead of re-implementing
  - Applies multiplier for ultrawides only
  - No sliders/preferences

Works well on my 32:9 screen now. I don't think it's necessary to change width for Thumbnails or App Icons. Titles are specific because reading text spread across 3000+ pixels feels awkward on super ultrawides. The PR only adjusts Titles for (super)wide screens and doesn't change anything else.

 ### Quick comparison

  | Screen | Aspect Ratio | Final Width | Notes |
  |--------|--------------|-------------|-------|
  | **Super ultrawide** | ≥3.0 (32:9) | ~30-33% | Uses comfortableWidth × 0.65, min 30% |
  | **Regular ultrawide** | 2.3-3.0 (21:9) | ~45-64% | Uses comfortableWidth × 0.85, min 45% |
  | **Normal screens** | <2.3 (16:9, etc) | 60% horizontal, 80% vertical | Unchanged from original |

The multipliers respect physical screen size (via comfortableWidth) while keeping text readable on ultrawides.

Is this acceptable?

lwouis:
Thank you @topik 

Reading your rationale, and looking at the code, the following thought comes to my mind.

Your goal is readability of the Titles Appearance on wide screens. The PR proposes to add a second layer of width calculation. It comes after the comfortable width is set. It reduces it in special cases.

Comfortable width should be comfortable. So why is it not currently? I think it's because it focuses on field-of-view. The AltTab window is scaled for a comfortable viewing experience. However, with Titles Appearance, there is an added topic for comfort: line length. This is not included today.

I think line-length should be addressed specifically. A common rule-of-thumb, used for latin-language books for example, is ~65 characters.

Thus here is my proposal: instead of adding arbitrary ratios to reduce the width, we instead look at font-size, calculate the width of 65 characters, then use the `min(comfortableWidth, 65chars)`. This guarantees that the panel is both confortable in regards to field-of-view and line-length.

What do you think?

Interestingly, many people have asked if we could have the Titles Apperance be more compact. That is: if its width could be the width of the longuest title. To avoid having lots of empty space. This is discussed in #3882. Depending on how we measure line-length here, we could do 2 birds with 1 stone, and set the width to `min(comfortableWidth, 65chars, widestRow)`

topik:
The character-width approach is outside the scope of what I'm trying to fix. The multiplier solution reuses comfortableWidth() like you suggested in your first review. If that's not good enough, feel free to close this. I'm not taking it further  🙂

lwouis:
Hi!

I've taken time to implement the design I mentioned. Could you please try out [this local build](https://github.com/user-attachments/files/23436971/AltTab.app.zip) and let me know how it feels with your setup?

In Titles Appearance, the width is dynamic. It will be the min of:
* 65 characters or around 2.5 alphabets
* The longest title there is to actually display
* The comfortableWidth() which is the field-of-view that fits into the eyesight without moving the neck

Thank you

NachiGithub:
image
Looks good. One bug, when tried just on finder with two windows list appears too narrow. 

same with other windows such as system preferences:

image
Looks fine on my firefox with dozens of windows or the full window list (all apps). 

I am on a mac studio 2022 with Tahoe 26.1.

lwouis:
Thank you for reporting it @NachiGithub. I fixed it!

This is included in the latest release 👍

PR #4962, opened 2025-11-07
feat: implement search functionality with fuzzy matching and auto-focus behavior

### LINKED ISSUE #590: Quick switch with fuzzy matching

JakkuSakura:
Bahaviors:
- type any keys to search
- tab to select which window to switch to, or re-enter search mode
- press enter to switch
- in selection mode, press space also works

closes #590
image
image

### DIFF
--- a/src/logic/events/KeyboardEvents.swift
+++ b/src/logic/events/KeyboardEvents.swift
@@ -94,11 +94,24 @@ class KeyboardEvents {
                     App.app.thumbnailsPanel.thumbnailsView.exitSearchFocus()
                     return nil
                 }
+                let isNavigationKey =
+                    keyCode == UInt16(kVK_LeftArrow)
+                    || keyCode == UInt16(kVK_RightArrow)
+                    || keyCode == UInt16(kVK_UpArrow)
+                    || keyCode == UInt16(kVK_DownArrow)
+                    || keyCode == UInt16(kVK_Home)
+                    || keyCode == UInt16(kVK_End)
+                    || keyCode == UInt16(kVK_PageUp)
+                    || keyCode == UInt16(kVK_PageDown)
                 if keyCode == UInt16(kVK_Return) || keyCode == UInt16(kVK_ANSI_KeypadEnter) {
                     if Windows.list.firstIndex(where: { Windows.shouldDisplay($0) }) != nil {
                         ControlsTab.executeAction("focusWindowShortcut")
                     }
                     return nil
+                }
+                if isNavigationKey {
+                    let matchedToggle = ControlsTab.shortcuts["searchToggleShortcut"]?.matches(nil, nil, UInt32(keyCode), event.modifierFlags) ?? false
+                    if !matchedToggle { return nil }
                 }
                 if keyCode == UInt16(kVK_Tab) {
                     let matchedToggle = ControlsTab.shortcuts["searchToggleShortcut"]?.matches(nil, nil, UInt32(keyCode), event.modifierFlags) ?? false
--- a/src/logic/events/KeyboardEventsTestable.swift
+++ b/src/logic/events/KeyboardEventsTestable.swift
@@ -92,6 +92,21 @@ func handleKeyboardEvent(_ globalId: Int?, _ shortcutState: ShortcutState?, _ ke
             App.app.thumbnailsPanel.thumbnailsView.exitSearchFocus()
             return true
         }
+        if let keyCode {
+            let isNavigationKey =
+                keyCode == UInt32(kVK_LeftArrow)
+                || keyCode == UInt32(kVK_RightArrow)
+                || keyCode == UInt32(kVK_UpArrow)
+                || keyCode == UInt32(kVK_DownArrow)
+                || keyCode == UInt32(kVK_Home)
+                || keyCode == UInt32(kVK_End)
+                || keyCode == UInt32(kVK_PageUp)
+                || keyCode == UInt32(kVK_PageDown)
+            if isNavigationKey {
+                let matchedToggle = ControlsTab.shortcuts["searchToggleShortcut"]?.matches(globalId, shortcutState, keyCode, modifiers) ?? false
+                if !matchedToggle { return true }
+            }
+        }
         if let toggleShortcut = ControlsTab.shortcuts["searchToggleShortcut"],
            toggleShortcut.matches(globalId, shortcutState, keyCode, modifiers) && toggleShortcut.shouldTrigger() {
             if let keyCode, isNonPrintableKeyCode(keyCode) {


lwouis:

Here are some macro-level review comments:

* The current search doesn't respect this setting:  
  image  
  It acts as if `Do nothing` was always selected
* Once tab is pressed and the search bar loses focus, then on further alt+tab, the search textfield is not selected thus AltTab acts like there is no search. It has to be clicked to resume previous behavior, as far as I can tell.
* When the search field is focused, the normal shortcuts don't work: arrows, tab, etc. I think the user will want to use those at this point

To recap those 3 points, I think the issue is that the search field acts too much like a normal input field. In conversations in https://github.com/lwouis/alt-tab-macos/issues/590, we discussed this type of issues. Solutions to it could be that this search box doesn't have all the convenient controls like arrow keys to move the cursor. That's what CopyQ does for example. Another solution would be to have some modal approach. For example, user presses a search key to enter search mode. if they press it again, they leave search mode. In practice, it could be as simple as focus/unfocus the search field here perhaps.

If we introduce a search key, it could replace the boolean preferences `Show search bar`. That would be a big win. I'm trying to avoid preferences, altogether. They make the app complex, and are a nightmare to maintain. I think ideally search would be there for all users. They would just need to press a key to go into search mode. This way users who don't want that just don't press the key to opt-in. They don't have to know search even exists.

Having AltTab, which is already quite complex, keyboard-wise, coexist with this layer of search, is a very difficult UX problem. That's why I mentioned in #590 that it would be important to write specifications before going into implementation, to clarify how everything would fit together 😅 

* Keys seem hardcoded. Users can set almost any key to do anything in `Preferences > Controls` and `Preferences > Controls > Shortcuts when active...`. We'll need to handle conflicts. We already have some conflicts avoidance code in place so shortcuts don't conflict with each other. We need to reuse this system with the search specific controls
* I think it would be very useful to highlight the matches in the window titles. Without that highlight, it can be a bit confusing why certain windows are shown and not others
* When there are no match, I think we need to give the search bar a minimum width. The whole panel currently collapses:  
  Screenshot 2025-11-08 at 13 16 41
* When windows are filtered out, it can be a bit jarring since the panel resizes instantly and content changes drastically. I wonder about an alternative approach: filtered windows appear grayed out, or are made transparent. This way the panel keeps the same size and windows the same position. We just see elimination happening. I'm not saying that this is a clear improvement. I'm just thinking out loud, how to reduce the visual jumpiness and help guide the user. Maybe the current system is best

What do you think?
Thank you

JakkuSakura:
Thanks for the detailed and fast feedback. Will fix small issues.

The reason I go straight into implementation is because I'm in need of this feature for my daily activities.

> When the search field is focused, the normal shortcuts don't work: arrows, tab, etc. I think the user will want to use those at this point
> Another solution would be to have some modal approach. For example, user presses a search key to enter search mode. if they press it again, they leave search mode. In practice, it could be as simple as focus/unfocus the search field here perhaps.

You can use `tab` key(or key input) to go to selection mode and use the normal shortcuts. Mouse also works. Pressing `tab` will change focus to search bar.

We could also make the search key configurable, by configuring tab or any key input

lwouis:
I understand, thank you!

> We could also make the search key configurable, by configuring tab or any key input

I think this is the key for this feature. This way it can be opt-in for power-users users, and away from sight for regular users. I think `s` could be a good default. It's close to `alt` and `tab` on the physical keyboard, and `s` stands for `search`

JakkuSakura:
I still think tab could be the most consistent option: 
1. tab could be used to switch to selection mode, making it also available to switch to search mode makes sense
2. I wanted to allow user to press any key to enter search mode, and feed this key into search bar
e.g.
image

Above should be similar to what we agree on

lwouis:
Very interesting!

I'm familiar with the IntelliJ/Jetbrains UX you show in the screenshot. See, an issue with this idea is that with AltTab, the user is holding `alt` to keep the UI open. If we don't use a search-activation key, then the user does alt+tab, then holds `alt` and has to type their text without dropping `alt`. So let's say i want to type `fire` to pick `firefox`. I will press alt+tab, then while holding tab, I will have to press `f` at least. Now maybe they can release tab, and we make AltTab stick like the `Do nothing` mode. I think the fact that the first key has to be pressed while holding `alt` is not great ergonomics.

We also have the added issue with that approach that `f` is currently bound to fullscreening the selected window. That would be a shortcut conflict. How to deal with that? Maybe the other `ire` characters would be fine because we would be in search-mode. But the `f` has to initiate before that. And there is the conflict.

That's why I think a search-activation key is so useful. I press alt+tab, then s (while holding `alt` still). Now I'm in search-mode and letter keys add to the search. Backspace will delete keys. Arrows keys should still move the window selection I think.

What do you think of this UX?

Thank you

JakkuSakura:
We have a few solutions here:
1. typing the search key would switch into do nothing mode temporarily, and user can type without holding the alt key 
2. set a do-nothing shortcut so it's working perfectly with search mode (no code change needed)

They should give user enough flexibility. 

In your focus when released mode, the UX would be:
1. disable "any key to search mode"
3. alt + tab to open the panel
4. press "s" to enter search mode, then release alt
5. type "firefox"
6. press enter, or press tab then enter(or configured key) 

In my usage in do nothing mode, the UX would be:
1. enable "any key to search mode"
2. alt + tab, release
3. type "firefox"
4. press enter

> We also have the added issue with that approach that f is currently bound to fullscreening the selected window. That would be a shortcut conflict. How to deal with that? Maybe the other ire characters would be fine because we would be in search-mode. But the f has to initiate before that. And there is the conflict.

Unfortunately any key to search mode does not work with alphabet shortcuts. Currently I give users text description about it. Later we could: alt + key to use the oringinal shortcut in any key mode. this is not in conflict with search key


I prefer any key mode because focuing a window is the most useful operation. It's acceptable to press one more alt key to trigger less used operation.


> Arrows keys should still move the window selection I think.

I agree with it. maybe need some experiments about implementation

lwouis:
I've thought about actually using this feature. I think it's not meant for the "focus on release" mode.

i think the search behavior is something like this:
* show switcher
* Hands are ready to type comfortably, meaning no key is held down
* i type a fragment of text that should isolate a window
* Either i did good and there is only 1 window, in which case it is automatically selected, and i can hit the existing "focus selected window key
* Or i was too generic with my fragment and multiple windows popup. In this case i think we could select the first automatically. At this point, i can review comfortably. Again no key is held. Now i either improve my fragment or navigate selection to the window i want and press the key to focus it

note that at this point i may want to quit the app or close the window, and a "exit search mode focus" key may be useful. Also some people may prefer search in priority while other may prefer other shortcuts in priority. A preference seems necessary here like "search is active by default"

i don't think a similar flow would be comfortable if i have to hold modifier keys the whole time. Thus i think we could have search activated on Do nothing and deactivated in the other mode. That being said. This feature may not be very discoverable.

JakkuSakura:
Would you allow `enter search mode focus` and `exit search mode focus` to be the same key? I found tab to be a comfortable key in "Nothing on release". Supporting "focus on release" would be tricky but we can avoid it in this PR

lwouis:
Yes indeed, I think it should be the same key "toggle search mode" 👍

JakkuSakura:
Currently "focus on release" also works with the feature.
> Yes indeed, I think it should be the same key "toggle search mode" 👍

Gave it a thought. If user set toggle search mode as tab, alt + tab, then another tab seems weird. 

So I prefer keeping them as different keys (but user can set them the same)

You can try out current version, then I can do a rebase and merge this PR

JakkuSakura:
I did a rebase and implemented a better match algorithm
It uses dynamic programming to match and sort according to relevance

lwouis:
@JakkuSakura are you still working on it, or should i do a review now?
Thank you

JakkuSakura:
It's complete on my side. Can review now

immadolf:
Hi @JakkuSakura , does your PR handle cases where an app’s display name is non-English? For example, on a Chinese system, Finder appears as “访达.” When searching, I’d prefer to type “Finder” rather than the Chinese “访达.”

JakkuSakura:
Hey @immadolf, my system is full English and I don't have a candidate app to test. Maybe you can have some tests or make another PR for it.

Some people may prefer fangda over finder for 访达 though

lwouis:
An issue if we match accross languages is that we can't highlight the matching fragments in the window titles. Maybe it's okay though. I see another case that we may want to handle similarly: i type "firefox" and it only shows me firefox windows, even though none of them have Firefox in the title. This could be an issue though. Maybe i'm looking for my window with "fire alarm", i type "fire" and i get all firefox windows and i don't understand why at that moment.

i suggest that this PR remains focused on window titles, matching them with highlights. Then we discuss refinements like multi lamguages separately in new tickets/PRs.

I plan to review this PR asap, ideally this weekend. Sorry for the wait 🙇‍♂️

lwouis:
* I think the fuzzy search should be case-insensitive. It's already fuzzy, and I think case sensitivity is too strict when searching. No one cares to capitalize when searching.
* I think the fuzzy algorithm you've implemented is impressive. I can make many mistakes and it still matches what I "intended" (e.g. typing "grontrol" will match "controls")
* Actually the matching may be too lenient. If I type 5-6 characters, it seems that I always end up with every window presented in the results. See this example:  
  image  
  I think search should be both lenient, but also relevant. It's hard to strike a balance. I think the current algo is a bit too lenient, and presents almost every window with almost every search pattern  
  image
  Actually, fuzzy searching has been studied a lot. Maybe there is a super powerful library we could just grab and use here. That would help with having a battle-tested algo that has been tweaks for years by experts
* I got an interesting case. I bound the letter `s` to the `Enter Search` shortcut. Then as I was trying to seach `test` in the search bar, when i hit the `s`, it exited search mode. This will be an issue if the user uses a simple key as a search exit shortcut. I think we should default to `escape` for exit then, and `s` for entering. This seems like it will make sense for people, and avoid this issue.
* I think we could split the code a bit. For example, all the Smith–Waterman code and general search code could go in a separate file, so avoid making Windows.swift bigger than it already is, and to have a dedicated place to read search logic.
* Truncation seems broken
* Showing App Name - Window Title seems broken
* Sometimes, there is extra padding at the bottom  
  image

Here's some feedback for today. Sorry I didn't have time to test more.

Overall, I found this PR very interesting. There is a lot of changes and code. It needs to be carefully reviewed before we can merge it, as it could break so many parts of the core experience.

Thank you for your hard work!

JakkuSakura:
> I think search should be both lenient, but also relevant. It's hard to strike a balance. I think the current algo is a bit too lenient, and presents almost every window with almost every search pattern

I tried to be lenient when matching, but also do a sorting based on relevance.

Thank you for the testing and ideas of product design. Will fix the other issues soon

JohanChane:
@lwouis @JakkuSakura I’d like to start using this feature—could you provide a release beta version?

JakkuSakura:
[AltTab.zip](https://github.com/user-attachments/files/23952588/AltTab.zip)
Hey @JohanChane, I'm not sure how to set up the CI in my branch, but here's the version I use locally

JohanChane:
> [AltTab.zip](https://github.com/user-attachments/files/23952588/AltTab.zip) Hey @JohanChane, I'm not sure how to set up the CI in my branch, but here's the version I use locally

Thank you very much—I’ve got it set up and running. Switching windows feels just as smooth as when I use rofi on Linux.

JohanChane:
@JakkuSakura Can the app name be added to the title’s appearance mode?  
Usually the first thing you do is type the application name; after that, when you keep typing, windows that don’t match are automatically filtered out, and if the field is empty, all windows are shown—just like in rofi?

JakkuSakura:
> Can the app name be added to the title’s appearance mode?
Usually the first thing you do is type the application name; after that, when you keep typing, windows that don’t match are automatically filtered out, and if the field is empty, all windows are shown—just like in rofi?

I'd like to leave it to next PR. It's already too many changes in this PR

shifman177:
I really want this feature 😅
Can I help with merging this one through?

@JakkuSakura @JohanChane @lwouis CC

JohanChane:
> I really want this feature 😅 Can I help with merging this one through?
> 
> @JakkuSakura @JohanChane @lwouis CC

Let’s wait for the author to merge first. Or you could use [contexts](https://contexts.co/); although development has stopped, it still works fine. See [ref](https://www.reddit.com/r/macapps/comments/1isiuab/contexts_worth_buying_it_in_2025/).

JakkuSakura:
> Prioritizes acronym matches. Character matches at start of the app name and start of any words in the window title get a higher score. Most of the time typing the acronym of an item’s title is sufficient to make it the first result.

It seems a great feature and I can steal the idea lol

lwouis:
Hi @JakkuSakura,

You're working on this branch, isn't it? Let me know if you're done and I should review/help 👍 

Today I was looking at https://github.com/ejbills/DockDoor, and I found they have a nice implementation of this fuzzy search, if you enable it:

image

I suggest playing with it. They went with the UX of pressing a search key to go into search mode. Then they did the implementation where arrow keys don't move the text cursor, and instead move the select thumbnail. I like the way they have the UI above the switcher panel. It's pretty nice inspiration 👍 

Thank you

JakkuSakura:
Hey I was only rebasing it. But I got an issue with infinite permission prompts and yet to solve the issue yet. Other than that, the PR is ready for review. If you could solve the infinite permission prompts issue, tests are appreciated

JakkuSakura:
> They went with the UX of pressing a search key to go into search mode. Then they did the implementation where arrow keys don't move the text cursor, and instead move the select thumbnail. I like the way they have the UI above the switcher panel. It's pretty nice inspiration 👍

I could implement this as default, and add an option if you like

JakkuSakura:
Took a look at DockDoor. Nice UI and will keep it.
The fuzzy search is less advanced as I implemented. E.g. I type 'fp', it won't search for FerroPhase. In alt-tab, it will search for FerroPhase, which is convenient.

lwouis:
Thank you for your message
Are you on macOS 15? There have been many reports of users of macOS 15 facing issues with AltTab v8. Basically the OS has severe bugs with the ScreenCapture Kit we use in v8. For now i have no idea how to work around it since it's the OS bugging, not AltTab havings bugs.

JakkuSakura:
I'm on MacOS 26.2 (25C56). Not sure why AltTab has this ussue, but DockDoor doesn't contain the issue

dacort:
Would love this functionality - I'm on Sequoia 15.7.3, lmk if I can help test at all!

lwouis:
@JakkuSakura this PR was opened recently: https://github.com/lwouis/alt-tab-macos/pull/5216

I think we should unite efforts 💪

lwouis:
Hi,

I've reviewed this PR, played with the app, and looked at the code. Here are some comments:

### Settings

Settings are tricky. We talked about it before a bit. AltTab philosophy since v7 is to reduce settings as much as possible. Thus I think we could:
* Remove the `Arrow keys navigate selection while searching`. Just decide that it is the case for everyone. I think it's a good UX
* Same idea for `Prefer acronym matches while searching`. We should just pick the one we prefer, and have it for everyone. Then we can tweak if there is feedback. Many top apps people use have a magic search that users can't customize. I think it should be good enough that customization doesn't feel missing, for most people.
* Combine both `Enter Dearch` and `Exist Dearch` into 1 `Toggle Search` shortcut, being "s" by default
* Rename the `Press any key to search`. Something like `Search is active by default`. Actually, I think most people who want this toggled on, would like the Switcher UI to remain on screen without holding `alt` (or other modifiers they picked), so they can properly type their search. Thus I propose to rework this setting. Instead of being a toggle in the sub-menu, we bring it as a third option here (e.g. `Enter Search`):

image

### Search experience

* Overall it's excellent. The search being fuzzy, and the highlights are very impressive and useful!
* I had issues with the `tab` key. I've set `s` to enter search mode. Then when I press `tab`, it unfocuses the search field. I think it's a default behavior that we should disable. Tab should do nothing if it's not bound to any shortcut.
* Same thing with other input shortcuts like alt+arrows or ctrl+arrows. They move the cursor. I think the cursor should be essentially fixed, and the only way to go back would be to delete.
* I got surprised a few times by pressing Space, and it would focus the window, instead of adding a space to the search. In Search Mode, we should block normal shortcuts like close window, minimize window, etc. However, focusing window is still needed at the end of the search. I think a good way to get out of this conflict is to change the default to focus window from `Space` to `Enter`. This way, people are free to type spaces in their search, and they can hit `Enter` to focus. I think it makes sense that a single-line text input would behave this way. 
* There is an extra padding we should remove, at the bottom of the switcher UI, in some cases. Not initially. Only with some search states: 
image
* Truncation is also broken in some cases, like we see on the screenshot above, or this one:
image
* The highlight UI is a bit too intense in Dark Mode. Also the yellow color in general I think is cool, but at the same time doesn't feel macOS-y. It feels a bit like from another app (e.g. IntelliJ). I'm not sure what Apple apps use for this. Maybe they use the user's accent color, which would be blue by default?
* I think adding diacritic insensitivity would be nice. This can be done later. I noticed in the other Search PR, they use some build-in method from AppKit, for this: https://github.com/lwouis/alt-tab-macos/pull/5216/files#diff-39dcbf35d8461748414a52771fd4c3f21d5ae65aace7a204651d976b4c8fd855R191

### Need to rebase

* I found the blue outlines are gone. Also I found some bugs about window ordering that I've fixed recently were back on this branch, of course, since it's not up to the latest

So overall, it's almost there!

Thank you

lwouis:
Hi @JakkuSakura,

Thank you for your hard work with this MR. I've broken things again by working on v9, sorry about this.

I'll try to take this MR to the finish line now. So no need to worry about rebasing again 👍

Thank you

lwouis:
Search is out now 🎉 
Closing the PR

Issue #4765, opened 2025-09-17
Frequent unintended Gesture triggers


sfkmk:
**Describe the bug**

The MacBook Pro trackpad is quite large. When I scroll or click with two fingers while my other hand accidentally touches the trackpad, AltTab interprets this as a three-finger swipe gesture and triggers the switcher. This happens even with a slight palm touch on the edge, which is common during normal MacBook usage.

**Steps to reproduce the bug**

1. In AltTab Settings, set the Gesture Trigger shortcut to "Vertical Swipe with Three Fingers" (ensure no other gesture uses this movement, such as Mission Control)
2. Place the palm of one hand slightly over the edge or corner of the trackpad (it must contact the trackpad surface)
3. With the other hand, perform a two-finger gesture as if scrolling a page

**Your environment**

* AltTab version: 7.28.0
* macOS version: 15.7
* Device: MacBook Pro 16"

This issue could potentially be resolved by restricting how far apart recognized "fingers" or "touches" can be to qualify for the trigger.

fhdufhdu:
@lwouis 

Hi, would it be alright if I try to improve this feature?
It seems to be reacting too sensitively, and I believe this is because it's calculating the average of the finger movements.

Here's the improvement I'm proposing:
- Current : Shows the view if the average of finger movements exceeds a certain threshold.
- Proposed : Shows the view only when all finger movements individually exceed a certain threshold.

lwouis:
Hi,

Of course your help would be very appreciated!

Regarding the proposed change, are you sure this would improve things? I'm wondering if it wouldn't progress in some scenarios and regress in others.

What i've been missing with gestures so far is academic research. I simply have no idea what algorithms work for real life situations. How to avoid false positives for real people, etc.

Reading your proposal, it sounds like it could help, but also i'm not sure! Research may hold insights that we currently don't have.

Thank you

fhdufhdu:
@lwouis 
Thanks for the reply.

I admit I don't know much about the academic research on this either.

However, I have already modified the code and have been using it locally for a while. Based on my personal experience so far, I am very satisfied with the change.

But I understand this is just my own case, and it seems we'll need opinions from others as well.

With that in mind, do you have any plans for a beta version? It seems like we need to apply this to a beta build first and gather feedback.

lwouis:
Thank you @fhdufhdu 

Indeed, if enough people are positive about the change, we can definitely release it.
[...]

PR #4987, opened 2025-11-13
Change gesture recognition logic

fhdufhdu:
This PR improves issue #4765 (Frequent unintended Gesture triggers).

### AS-IS
Previously, gesture detection was triggered if the average distance moved by all fingers (from their start to end positions) exceeded a certain threshold.

This caused problems on MacBook trackpads. For example, if a user's palm was touching the trackpad while they tried to scroll in another app with two fingers, the average distance could easily exceed the threshold, unintentionally triggering the View.

### TO-BE
Instead of calculating an average, the new logic detects the distance for each finger individually. The gesture will not be triggered if even one finger does not move beyond the set distance.

Additionally, I have slightly lowered the sensitivity, as the View previously appeared with very minimal movement. I welcome any feedback on this new sensitivity level!

### DIFF
--- a/src/logic/events/TrackpadEvents.swift
+++ b/src/logic/events/TrackpadEvents.swift
@@ -111,6 +111,14 @@ class GestureDetector {
         }
         return totalDelta / activeTouches.count
     }
+    
+    static func computeDistance(_ activeTouches: Set<NSTouch>, _ startPositions: [String: NSPoint]) -> Array<NSPoint> {
+        var deltas: Array<NSPoint> = []
+        for touch in activeTouches {
+            deltas.append(touch.normalizedPosition - startPositions["\(touch.identity)"]!)
+        }
+        return deltas
+    }
 
     static func blockOngoingScrolling() {
         if #available(macOS 10.13, *) {
@@ -125,7 +133,7 @@ class GestureDetector {
 class TriggerSwipeDetector {
     // when the native using 3-finger swipe to shift Space, macOS will wait that a small distance is traveled before acting
     // We imitate this behavior
-    static let MIN_SWIPE_DISTANCE: Double = 0.015 // % of trackpad surface traveled
+    static let MIN_SWIPE_DISTANCE: Double = 0.1 // % of trackpad surface traveled
     // when the native using 3-finger swipe to shift Space, macOS will prevent a swipe until the fingers are raised,
     // if the user moves too much in the vertical direction. We imitate this behavior
     static let MAX_SWIPE_DISTANCE_IN_WRONG_DIRECTION: Double = 0.1 // % of trackpad surface traveled
@@ -135,15 +143,17 @@ class TriggerSwipeDetector {
 
     static func check(_ activeTouches: Set<NSTouch>) -> Bool? {
         if !App.app.appIsBeingUsed && swipeStillPossible {
-            let averageDistance = GestureDetector.computeAverageDistance(activeTouches, startPositions)
-            let (absX, absY) = (abs(averageDistance.x), abs(averageDistance.y))
-            let horizontal = Preferences.nextWindowGesture.isHorizontal()
-            if (updateSwipeStillPossible(horizontal ? absY : absX) && (horizontal ? absX : absY) >= MIN_SWIPE_DISTANCE) {
-                reset()
-                DispatchQueue.main.async { App.app.showUiOrCycleSelection(Preferences.gestureIndex, false) }
-                return true
+            let distances = GestureDetector.computeDistance(activeTouches, startPositions)
+            for distance in distances {
+                let (absX, absY) = (abs(distance.x), abs(distance.y))
+                let horizontal = Preferences.nextWindowGesture.isHorizontal()
+                if (!(updateSwipeStillPossible(horizontal ? absY : absX) && (horizontal ? absX : absY) >= MIN_SWIPE_DISTANCE )) {
+                    return false
+                }
             }
-            return false
+            reset()
+            DispatchQueue.main.async { App.app.showUiOrCycleSelection(Preferences.gestureIndex, false) }
+            return true
         }
         return nil
     }


lwouis:
Hi @fhdufhdu,

I reviewed the code changes. I wonder one thing: you updated the logic in `TriggerSwipeDetector` to compare each finger instead of the average of all fingers.

You kept the logic of using the average for `NavigationSwipeDetector`. Is that intentional? That once the gesture is initiated, average of fingers is more lenient and precise than another logic?

Thank you

lwouis:
I've played around with this PR. I've also compared to my wife Windows laptop which has the same swipe to alt-tab feature. I've made 3 changes:

* Brought back the sensitivity to trigger. I understand that higher sensitivity means lower chance of accidental trigger. However, it also means a worse experience when the user actually want to trigger alt-tab. I found it very slugish when trying to trigger it with sensitivity higher than the current one
* I've reduce sensitivity for the `NavigationSwipeDetector`, to be twice that of the trigger, instead of 3 times. On Windows, it was way lower actually. Hard to use I thought. Here with 0.03, I found it easier to reach far away thumbnails, if I have many thumbnails. Otherwise if you reach the end of the trackpad with your 3 fingers, you're kind of stuck. Impossible to raise the hand to loop to the left to grab more space. It will hide AltTab
* I've made it so that once a swipe is ongoing, the user can relax and use as few as 2 fingers. 1 finger or 0 finger will hide AltTab. This is what macOS and Window do for 3-4 fingers swipes. It makes it comfortable as the user can start with 3-4 fingers, then relax to 2 for navigation 👍

fhdufhdu:
@lwouis 
Here's the reason why I kept the logic of using the average in NavigationSwipeDetector:

- I thought it was acceptable for it to be a bit more sensitive (responsive) since the view is already open.
- And yes, you are correct: as you mentioned, using the average of the fingers is indeed more lenient and precise than the other logic.

Thank you for your other feedback on improving this PR!

lwouis:
This has been included in the latest release 👍 
Let's close the PR

lwouis:
Hi @fhdufhdu,

Could you please help testing [the build discussed here](https://github.com/lwouis/alt-tab-macos/issues/5039#issuecomment-3563875639)?

I has this patch:

```
Subject: [PATCH] fix: better gesture recognition (closes #5039)
---
Index: src/logic/events/TrackpadEvents.swift
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/logic/events/TrackpadEvents.swift b/src/logic/events/TrackpadEvents.swift
--- a/src/logic/events/TrackpadEvents.swift	(revision e477f8b3da4635fc8ad9304d36b5201604108d96)
+++ b/src/logic/events/TrackpadEvents.swift	(revision 2baa48d5dc3f9be74e466f8ad82314443171e550)
@@ -48,32 +48,54 @@
 
 private func touchEventHandler(_ cgEvent: CGEvent) -> Bool {
     guard let nsEvent = GestureDetector.convertEvent(cgEvent) else { return false }
-    // if the finger count doesn't match, we reset tracking data, and may trigger fingersUp
     let touches = nsEvent.allTouches()
-    if touches.count == 0 { return false } // sometimes the os sends events with no touches
-    let activeTouches = touches.filter { !$0.isResting && ($0.phase == .began || $0.phase == .moved || $0.phase == .stationary) }
+    // sometimes the os sends events with no touches; we ignore these as they could break our gesture logic
+    if touches.count == 0 { return false }
+    FingersDown.update(touches)
     let requiredFingers = Preferences.nextWindowGesture.isThreeFinger() ? 3 : 4
-    if (!App.app.appIsBeingUsed && touches.count != requiredFingers) || (App.app.appIsBeingUsed && touches.count < 2) {
+    if (!App.app.appIsBeingUsed && FingersDown.count != requiredFingers) || (App.app.appIsBeingUsed && FingersDown.count < 2) {
         TriggerSwipeDetector.reset()
         NavigationSwipeDetector.reset()
-        return GestureDetector.checkForFingersUp(activeTouches, requiredFingers)
+        return GestureDetector.checkForFingersUp(FingersDown.count, requiredFingers)
     }
     // when the native using 3-finger swipe to shift Space, macOS will block scrolling in the background
     // We imitate this behavior by sending a synthetic scrollWheel event
     GestureDetector.blockOngoingScrolling()
+    let activeTouches = touches.filter { !$0.isResting && ($0.phase == .began || $0.phase == .moved || $0.phase == .stationary) }
     // trigger actions if conditions are met
     if App.app.appIsBeingUsed {
-        if !GestureDetector.updateStartPositions(touches, &NavigationSwipeDetector.startPositions) {
-            if let r = NavigationSwipeDetector.check(touches) { return r }
+        if !GestureDetector.updateStartPositions(activeTouches, &NavigationSwipeDetector.startPositions) {
+            if let r = NavigationSwipeDetector.check(activeTouches) { return r }
         }
     } else {
-        if !GestureDetector.updateStartPositions(touches, &TriggerSwipeDetector.startPositions) {
-            if let r = TriggerSwipeDetector.check(touches) { return r }
+        if !GestureDetector.updateStartPositions(activeTouches, &TriggerSwipeDetector.startPositions) {
+            if let r = TriggerSwipeDetector.check(activeTouches) { return r }
         }
     }
     return false
 }
 
+// macOS only sends deltas. We want to know the final state of how many fingers are down on the trackpad
+// this class keeps track between events
+class FingersDown {
+    static var count: Int = 0
+
+    static func update(_ touches: Set<NSTouch>) {
+        let phases = touches.map { $0.phase }
+        let began = phases.filter { $0 == .began }.count
+        let ended = phases.filter { $0 == .ended }.count
+        count += began
+        count -= ended
+        // Never let it go negative (can happen with OS anomalies)
+        if count < 0 { count = 0 }
+        // This catches cases where macOS fails to send some ended events or we attach mid-gesture.
+        let nonEndedTouches = touches.filter { $0.phase != .ended }.count
+        if abs(count - nonEndedTouches) > 1 {
+            count = nonEndedTouches
+        }
+    }
+}
+
 class GestureDetector {
     static func convertEvent(_ cgEvent: CGEvent) -> NSEvent? {
         var nsEvent: NSEvent?
@@ -84,8 +106,8 @@
         return nsEvent
     }
 
-    static func checkForFingersUp(_ touches: Set<NSTouch>, _ requiredFingers: Int) -> Bool {
-        if App.app.appIsBeingUsed && touches.count < requiredFingers && App.app.shortcutIndex == Preferences.gestureIndex
+    static func checkForFingersUp(_ fingersDown: Int, _ requiredFingers: Int) -> Bool {
+        if App.app.appIsBeingUsed && fingersDown < requiredFingers && App.app.shortcutIndex == Preferences.gestureIndex
                && !App.app.forceDoNothingOnRelease && Preferences.shortcutStyle[App.app.shortcutIndex] == .focusOnRelease {
             DispatchQueue.main.async { App.app.focusTarget() }
             return true
@@ -174,12 +196,12 @@
 
     static var startPositions: [String: NSPoint] = [:]
 
-    static func check(_ touches: Set<NSTouch>) -> Bool? {
-        let averageDistance = GestureDetector.computeAverageDistance(touches, startPositions)
+    static func check(_ activeTouches: Set<NSTouch>) -> Bool? {
+        let averageDistance = GestureDetector.computeAverageDistance(activeTouches, startPositions)
         let (absX, absY) = (abs(averageDistance.x), abs(averageDistance.y))
         let maxIsX = absX >= absY
         if (maxIsX ? absX : absY) > MIN_SWIPE_DISTANCE {
-            maxIsX ? resetX(touches) : resetY(touches)
+            maxIsX ? resetX(activeTouches) : resetY(activeTouches)
             let direction: Direction = maxIsX ? (averageDistance.x < 0 ? .left : .right) : (averageDistance.y < 0 ? .down : .up)
             DispatchQueue.main.async { App.app.cycleSelection(direction, allowWrap: false) }
             return true
```

Thank you

PR #5114, opened 2025-12-10
Feature/windows filtering

SyomaKiss:


lwouis:
Hi,
Such filtering targets power-users. Starting with v7, we've decided to improve the experience of most users, by not having features with niche usage.
I'm afraid we can't integrate such filtering into AltTab. Already, we are considering removing the Backlists feature.
Thank you for your understanding 🙇

SyomaKiss:
Thanks for answering! Can you pls clarify what is motivation behind keeping less features in the app?

Also, what do you think about the option of keeping UI clean without this tab, but leaving the option for such configuration through command line interface, so that it does not bother most of the users, but power users can access some features with command line interface?

lwouis:
You can find some of the rationale here: https://github.com/lwouis/alt-tab-macos/issues/351

You can also run v6 and current release, and compare the Settings to see the drastic change.

Regarding hidden settings, I won't go that route. Too many settings is an issue for users because it's intimidating/confusing. It's also an issue for maintainers because they are too much. QA for instance is already unmanageable for me.

Important context to keep in mind here: #1179 

Thank you

Issue #2694, opened 2023-07-25
Live Preview like windows 11


AISLVE:
It would be great to see the live update feature like when i play a video in windows it can be seen being played after pressing alt+tab or like the stage manager

lwouis:
Please see #122

AISLVE:
What about even windows, right now they are shown in preview based on their size(some are small some are large), but I sometimes have visual problems where I can't exactly see the preview because it's too small, it's not that big of a deal,but would be a great QOL feature

Issue #2407, opened 2023-03-12
Live Previews


jozhtn:
**_Is your feature suggestion related to a problem? Please describe._**
Kind of? In Windows 10/11, when you press alt + tab you get to see a live preview of the app while it's open, but in AltTab it only shows a static image. Would be a nice feature to have to make it more similar to what Windows offers. 

An example of this would be to press alt + tab on macOS while a youtube video is playing, you won't see the video playing in the preview, you'll just see a static image from when you pressed alt + tab. In Windows if you press alt + tab while a youtube video is playing, the video will continue to play in the alt + tab preview.

**_Describe the solution you'd like_**
I'm not sure if this is possible, or if there is some sort of restriction in macOS permissions, but this would be a great feature for people coming from Windows, like myself.

**_Describe alternatives you've considered_**
None, the app pretty much has everything the Windows alt + tab has except this.

lwouis:
Duplicate of #122

PR #5118, opened 2025-12-11
feat: Add live window preview using ScreenCaptureKit

suatkocar:
## Background

I've read through #122 (Display video of the windows instead of static screenshots) and related issues (#2407, #2694, #967). The request for live window previews has been a long-standing feature request since 2019.

In your March 2022 comment on #122, you mentioned experimenting with ScreenCaptureKit and documented your findings in #447. I took that as a starting point and built upon it to create a working implementation.

## What This PR Does

This PR adds an **optional** live video preview feature that shows real-time window content instead of static thumbnails. The feature is:

- **Disabled by default** - Users must explicitly enable it in Preferences → Appearance → Live Preview
- **Fully configurable** - Quality, frame rate, and scope can be adjusted based on user preferences and system capabilities
- **Backwards compatible** - Gracefully disabled on macOS versions older than 12.3

### ScreenCaptureKit Approach

After reviewing the constraints discussed in #122 (private APIs being patched in Catalina, etc.), I focused on using the public `ScreenCaptureKit` framework introduced in macOS 12.3. This provides:

- Official Apple API with long-term support
- Per-window video capture via `SCStream`
- Configurable resolution and frame rate via `SCStreamConfiguration`

### Architecture

**LiveWindowCapture.swift** - Core capture manager
- Singleton pattern for AppKit integration
- Manages multiple `SCStream` instances (one per window)
- Incremental stream management - only starts/stops streams as needed, avoiding unnecessary overhead
- `CIContext` reuse in `StreamOutput` for efficient frame processing

**Preference Enums**
- `LivePreviewQualityPreference`: Thumbnail (320px) → Native (full resolution)
- `LivePreviewFrameRatePreference`: 5 FPS → 120 FPS (ProMotion support)
- `LivePreviewScopePreference`: Selected Window Only / Selected App's Windows / All Windows
- `LivePreviewStreamKeepAlivePreference`: Immediately close (0) / Keep forever (-1) / Custom duration (seconds)

**Scope Filtering**
- Capture streams run for all visible windows
- Filtering happens at the display level (in `onFrameUpdate` callback)
- This prevents video freezing when hovering between windows

### Performance Considerations

- **Quality settings** affect `SCStreamConfiguration.width/height` - lower quality = less GPU usage
- **Frame rate** controls `minimumFrameInterval` - lower FPS = less CPU usage
- **Scope** determines how many thumbnails get updated - "Selected Window Only" is the lightest option
- Users are warned in the UI: *"Higher quality and frame rate use more CPU/GPU resources. Use lower settings if you experience lag."*

## Default Settings

- **Enable Live Preview**: Off (opt-in)
- **Quality**: Native (Best)
- **Frame Rate**: 15 FPS
- **Scope**: All Windows

## Notes

This is my proposed solution for the live preview feature. I understand this is a significant addition, and I'm happy to make any adjustments based on your feedback. The implementation tries to stay consistent with the existing codebase patterns and coding style.

@lwouis - Would love to hear your thoughts on this implementation

### DIFF
--- a/alt-tab-macos.xcodeproj/project.pbxproj
+++ b/alt-tab-macos.xcodeproj/project.pbxproj
@@ -59,6 +59,7 @@
 		1C961F9CF4F1BEDF04ACAEE0 /* Switch.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1C96189273B7482124C36F4C /* Switch.swift */; };
 		1C961FA2D2D99AE54EBECA67 /* titles_show_running_windows_light@2x.jpg in Resources */ = {isa = PBXBuildFile; fileRef = 1C9610B2E76F1E429C8D3660 /* titles_show_running_windows_light@2x.jpg */; };
 		1C961FB59566B8314771F58E /* thumbnails_show_preview_focused_window_light@2x.jpg in Resources */ = {isa = PBXBuildFile; fileRef = 1C961EF84B061A68FD9AEC46 /* thumbnails_show_preview_focused_window_light@2x.jpg */; };
+		346D342C2EEA72A4009590BE /* LiveWindowCapture.swift in Sources */ = {isa = PBXBuildFile; fileRef = 346D342B2EEA72A4009590BE /* LiveWindowCapture.swift */; };
 		442E26DB1F7DA80AF794CCB6 /* Pods_unit_tests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = C7E6F409CA442F83B7BE71AD /* Pods_unit_tests.framework */; };
 		4807A6C623A9CD190052A53E /* SkyLight.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4807A6C523A9CD190052A53E /* SkyLight.framework */; };
 		48F3E16224EC0D8800A1C64B /* Localizable.strings in Resources */ = {isa = PBXBuildFile; fileRef = D04BABDA79F8EF76E3ACDD5F /* Localizable.strings */; };
@@ -349,6 +350,7 @@
 		1C961F5740647DBFA0540789 /* thumbnails_show_tabs_as_windows_light@2x.jpg */ = {isa = PBXFileReference; lastKnownFileType = image.jpeg; path = "thumbnails_show_tabs_as_windows_light@2x.jpg"; sourceTree = "<group>"; };
 		1C961F68C77AAE38E5A2E046 /* app_icons_show_running_applications_light@2x.jpg */ = {isa = PBXFileReference; lastKnownFileType = image.jpeg; path = "app_icons_show_running_applications_light@2x.jpg"; sourceTree = "<group>"; };
 		1C961FB3DCF4906FEEF4A23D /* TableGroupView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = TableGroupView.swift; sourceTree = "<group>"; };
+		346D342B2EEA72A4009590BE /* LiveWindowCapture.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LiveWindowCapture.swift; sourceTree = "<group>"; };
 		4807A6C523A9CD190052A53E /* SkyLight.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = SkyLight.framework; path = ../../../../System/Library/PrivateFrameworks/SkyLight.framework; sourceTree = "<group>"; };
 		481FE54624D2D387001032F1 /* alt-tab-macos-Bridging-Header.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = "alt-tab-macos-Bridging-Header.h"; sourceTree = "<group>"; };
 		5883F0A729A5182C0071DB65 /* TrackpadEvents.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TrackpadEvents.swift; sourceTree = "<group>"; };
@@ -1285,6 +1287,7 @@
 		D04BA0D80B24E72B9B981A1D /* logic */ = {
 			isa = PBXGroup;
 			children = (
+				346D342B2EEA72A4009590BE /* LiveWindowCapture.swift */,
 				D04BACD976030676FD0761D5 /* Windows.swift */,
 				D04BA9B93823398A542FF7A0 /* Preferences.swift */,
 				D04BA68C2561D9EE4FD851B8 /* NSScreen.swift */,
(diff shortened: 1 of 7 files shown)


lwouis:
Hi,

It seems you used AI to write this PR, code and message. I tried to checkout the code, and run it, and it doesn't compile:

image

Could you please make sure the code works before submitting it? I think here the issue is that the LiveWindowCapture file was not added through XCode, so it exists but it not recognized as part of the XCode project. Thus missing when time comes to compile it.

Thank you

suatkocar:
Hi @lwouis,

Thanks for catching that. You're right - the Xcode project file wasn't included in the commit. I had local code signing changes that I needed to exclude, and in doing so I accidentally omitted the entire file.

I've pushed a fix that.

Could you please try building again?

lwouis:
I was able to play with this build today. It's very interesting and impressive!

* The window "traffic-light buttons" are replaced with some purple UI. Since AltTab grabs every window to show thumbnails, the whole screen gets covered to this purple UI. It's acceptable I think. Yet it's a bit jaring. Not sure it can be helped.
* Performance is definitely something we need to consider. AltTab was running at 200% CPU when I opened it with a few windows open. This is even without a preview. I noticed that the thumbnails are also videos. I thought this PR only made the Preview a video. Every is video actually.
* The capture API seems to behave quite nice compared to my memory:
  * Partially off-screen windows are accessible. Which the currrent API fails to do!
  * Minimized window show the latest screenshot before it was minimized, somehow. That's very nice!
  * It shows screenshot for minimized-on-another-Space! I can't believe it
  * Fullscreen windows seem to render smaller/wrong. This one should be checked ⚠️ 
  * There is some stuttering when invoking AltTab. I think maybe the code showing thumbnail has not been properly removed. When going Video, it should use any screenshot I think. It should be either Screenshots, or Video Stream. Never both, I think.

There is a lot to test with this build:
* Permissions management (e.g. can it run when the user skips Screen Recording Permissions)
* Does it work when the switcher shows apps, and not windows (when this advanced setting is set
* App Icons and Titles Appearance
* How performance scales with the number of windows. How do we handle things if performance is bad with many windows? Perhaps we could have just a frame of each window, like a screenshot, and only animate visible rows, or some such optimization?

The settings need to be simplified, that's for sure. I think native resolution and same-FPS-as-source is a must. Users wouldn't accept pixalized windows. They would rather see a static high-quality screenshot. So these settings can go, I think.

I'll look into this PR more soon 👍 
Let me know your thoughts please~

**Side-note on AI**. This looks coded with AI. Could you please share about your process a bit? I haven't had much success using AI on this project so far.

lwouis:
Sharing here my findings, comparing the 2 APIs:

Scenario | `CGSHWCaptureWindowList` | `ScreenCaptureKit`
-|-|-
Window on another Space | ✅ | ✅ 
Quality | ✅ | ✅ 
Stage Manager | ❌ | ❌
Minimized before AltTab is launched | ✅ | ❌
Hidden before AltTab is launched | ✅ | ❌
Tabbed window before AltTab is launched | ✅ | ❌
Purple capture UI on each window | ✅ | ❌
Parts of the window is offscreen | ❌ | ✅ 
Fullscreen; top-bar included? | ❌ | ✅
Modal is open and attached the the window | ❌ | ✅
Private API | ❌ | ✅
Memory leak | ❌ | ✅
HDR support | ❌ | ✅
CPU consumption | ✅ | ✅ (can screenshot, or video; CPU is high with video but not a problem)
RAM consumption | 🟨 (low but leak) | ✅

lwouis:
I don't know what to think about what happens when the user has many window, and recording video shows many purple indicators. It's a bit crazy looking and laggy I guess:

https://github.com/user-attachments/assets/d8d7457a-afa9-4c05-921b-c820e06afa9b

lwouis:
Could you please try out [this build](https://github.com/user-attachments/files/24273864/AltTab.app.zip)? I've built on this PR, and improved performance, removed some settings, and added a static screenshots approach using ScreenCaptureKit, in addition to the video thumbnails 👍 

Thank you

suatkocar:
Thanks for sharing your build. I tested it and the improvements are significant - the lag has dropped noticeably, streams stop instantly when AltTab closes, and the restart latency is much better.

Regarding the purple indicator: due to Apple's limitations, I don't think we can avoid it without asking for screen sharing permission for each individual window. For example, when sharing your screen with Google Meet, you choose what to share, which prevents the indicator from appearing on that specific window. However, since our approach requires capturing all windows simultaneously, asking the user to grant permission for each window individually wouldn't be practical. So unless we find a way to get system-level access, the purple indicators seem unavoidable.

If you have time to push your changes back into this PR (or a separate branch), I'd love to diff through them and test the same edge cases on my end. Seeing your implementation would help me verify scenarios like fullscreen offsets, many-window layouts, and the permissions flow.

**Side note on the development process:** This implementation originally came from prototyping a similar live-preview feature in Rust/GPUI. After some experimentation and LLM-assisted scaffolding, I realized ScreenCaptureKit could provide a workable Swift solution despite WindowServer's limitations. I used AI to help structure the initial PR description and code scaffolding - it's faster and helps keep documentation organized, though I reviewed and tested everything manually.

lwouis:
Thank you @suatkocar!

So this is why your PR was so high-quality, yet i sensed the touch of AI. Thank you for sharing. Which AI model/integration did you use? I myself have tried IntelliJ integrated AIs, and find myself using ChatGPT often these days to explore ideas.

Here's the local branch I have at the moment, with the latest video-thumbnail code. It's all in your commit which I amended 👍 

Thank you for your help! It's great to see people putting such effort in PRs!

lwouis:
Hi,

I've worked on this quite a bit. I've hit a performance wall with the naive implementation. Core Animation struggles to keep up with all the updates to each ThumbnailView/Thumbnail-frames.

I've tried a batch approach: we paint a large Metal View with all the thumbnails painted on it. It helps because we can draw this large view at 60Hz, and independently, on a background-queue, update the latest frames of each thumbnail.

However, even with this approach, I'm struggling to get decent performance, when there are many windows open on screen.

I've spent a lot of time in Instruments / ChatGPT. I'm struggling to fix the performance issues as it's not a simple "main thread CPU is high" situation. Performance issues lies somewhere in the CPU/GPU/Shaders/Metal/CoreAnimation side. I'm not used to profile/optimize this topology. This architecture is a bit like a video-game engine, I believe.

Here's the branch if you'd like to have a look: https://github.com/lwouis/alt-tab-macos/tree/yes-video

Thank you

lwouis:
This PR was really helpful! I think video thumbnails are not a good idea for now, given macOS limitations. I explained more here: https://github.com/lwouis/alt-tab-macos/issues/122#issuecomment-3887261435

Closing the PR. Thank you again! This inspired all the performance improvements in v9

PR #5145, opened 2025-12-21
perf: implement lazy initialization for recycled view pool

damiansire:
I'm not 100% sure if this is the best approach. It saves memory but there might be a tiny delay when creating views for the first time. Do you think this could affect UX? Should we do benchmarks before merging?

I think (Just personal opinion) view creation happens before showing the UI and the cost should be minimal, but I'd love to hear your thoughts

(I love to hear your thoughts =  barely know how to run the project and tests, but I still don't know how to do UX benchmarks 😂)

### DIFF
--- a/src/ui/main-window/ThumbnailsView.swift
+++ b/src/ui/main-window/ThumbnailsView.swift
@@ -10,8 +10,7 @@ class ThumbnailsView {
 
     init() {
         updateBackgroundView()
-        // TODO: think about this optimization more
-        (1...20).forEach { _ in ThumbnailsView.recycledViews.append(ThumbnailView()) }
+        // Views are created on-demand in layoutThumbnailViews() to reduce memory usage
     }
 
     func updateBackgroundView() {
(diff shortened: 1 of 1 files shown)


lwouis:
We need some numbers for this optimization. How much memory would it save? How long does it take to instanciate on-the-fly?

My suspition is that the recycled pool uses little memory. I profiled the app yesterday, and most of the memory was going to the PreferencesWindow apparently in layout things, and in PermissionsWindow with the small images, somehow. So looking at this, it seems that the biggest gains are to be realized by improving the secondary windows, not the main switcher window.

If you haven't done it yet, I suggest you profile AltTab using Instruments > Allocations, and get a sense of where the bulk of the memory is going. It's a great way to find where to start trimming the fat. Our intuitions on this matter are often not correct. When I did it yesterday, I wouldn't have guessed that layout constraints would take up megabytes of RAM.

For now, I've always prioritize snappiness over any over metric, with this project. I think that's what people want. They want to press alt+tab, and that we show them pixels as fast as we can. That's why we show cached thumbnails today for example, and then update them. It flickers a bit, but it creates for a more responsive UX.

PR #5146, opened 2025-12-21
feat: Enable image compression to reduce memory usage

damiansire:
I'm new to Mac and these were my first PRs, so I don't know how to benchmark UX on this yet... Sooo... the UX might be affected, since I'm looking at everything purely from a performance standpoint

- Changed layer contentsGravity from .resize to .center
- Enabled CPU-based image resizing using resizedCopyWithCoreGraphics
- Images now stored at display size instead of full resolution
- Trades one-time CPU resize cost for significant memory savings

Memory impact:
- Before: ~14 MB per full-resolution window thumbnail
- After: ~100-200 KB per resized thumbnail
- Expected reduction: 200-250 MB for typical usage with 20 windows

### DIFF
--- a/src/logic/IconSizeCalculator.swift
+++ b/src/logic/IconSizeCalculator.swift
@@ -1,0 +1,38 @@
+import Cocoa
+
+/// Utility class to calculate optimal icon sizes based on monitor configuration
+/// This helps balance memory usage with visual quality across different display setups
+class IconSizeCalculator {
+    private static var cachedOptimalSize: CGSize?
+    
+    /// Calculate optimal icon size across all monitors with 20% buffer for quality
+    /// - Parameters:
+    ///   - displaySize: The target display size for the icon
+    ///   - scaleFactor: The backing scale factor of the current screen
+    /// - Returns: Optimal size that works well across all connected monitors
+    static func optimalIconSize(for displaySize: NSSize, scaleFactor: CGFloat) -> NSSize {
+        if let cached = cachedOptimalSize {
+            return cached
+        }
+        
+        // Find maximum scale factor across all screens
+        // This ensures icons look sharp even on the highest-resolution display
+        let maxScaleFactor = NSScreen.screens.map { $0.backingScaleFactor }.max() ?? scaleFactor
+        
+        // Calculate size with 20% buffer for quality
+        // The buffer prevents pixelation when icons are slightly enlarged
+        let bufferMultiplier = CGFloat(1.2)
+        let width = displaySize.width * maxScaleFactor * bufferMultiplier
+        let height = displaySize.height * maxScaleFactor * bufferMultiplier
+        
+        let optimalSize = NSSize(width: width, height: height)
+        cachedOptimalSize = optimalSize
+        return optimalSize
+    }
+    
+    /// Invalidate cache when monitor configuration changes
+    /// Should be called when monitors are plugged/unplugged or resolution changes
+    static func invalidateCache() {
+        cachedOptimalSize = nil
+    }
+}
--- a/src/logic/events/ScreensEvents.swift
+++ b/src/logic/events/ScreensEvents.swift
@@ -8,6 +8,9 @@ class ScreensEvents {
     @objc private static func handleEvent(_ notification: Notification) {
         Logger.debug(notification.name.rawValue)
         Spaces.refresh()
+        // Invalidate icon size cache when monitor configuration changes
+        // This ensures icons are recalculated for the new display setup
+        IconSizeCalculator.invalidateCache()
         // a screen added or removed, or screen resolution change can mess up layout; we reset components
         App.app.resetPreferencesDependentComponents()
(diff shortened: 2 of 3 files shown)


lwouis:
I'm in the process of reworking thumbnails to introduce a new API, as well as video thumbnails. You can check already quite mature code [on this branch](https://github.com/lwouis/alt-tab-macos/commits/video-thumbnails/).

In general, the current strategy is to store the full-size screenshots in the Window object. It's wasteful, but necessary because:
* The user may have Previews on. In that case, we need the 1-to-1 pixels images, to show the preview on top of the indow
* The user may have multiple monitors, and if they switch to another, we have to resize the image. We need to keep it bigger then the current size, in case we need to resize to a bigger size

So the current strategy is to store it once, in high-definition, in Window, then pass the data-source (it's a pixelBuffer in the new branch, often; otherwise a CGImage, like for the app icons) to the ThumbnailView.thumbnail, and let the GPU resize it for free.

The only optimization we could do, I think, is to check if Previews are enabled, and if not, then we could save in Window.thumbnail at a smaller size. I don't know how we would elect the right size though. Probably would need to check every monitor, and do some heuristic + add some buffer size. Something like "ok the biggest thumbnail would be on monitor 2, with size X. Let's keep the source screenshot at size X * 120%.

I'm not sure how this strategy would work in practice though. I mean, can the APIs screenshot at lower resolution? To be confirmed. Otherwise, we are doing extra CPU-resizing, for the sake of memory. It may not be a good tradeoff as it would drain battery, to use less memory, maybe not always a great tradeoff for most people.

Thank you

lwouis:
On the branch I shared above, I moved the pixelFormat to `kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange`, and that reduced image memory by 60%, compared to `kCVPixelFormatType_32BGRA` I was using 🚀 

I can see only 2 optimisations paths remaining. They both do the same things: avoid storing the native resolution image, if the user doesn't use Preview:

* capturing smaller resolution screenshots. It's easy with ScreenCaptureKit. We just set the target size in SCStreamConfiguration. I tested, and it seems to work well. As I discussed previously, we would still need to compute what the biggest thumbnail size would be, on the multiple monitors the user may have. Then that number is the highest-resolution we need to store.

* This is is a compromise. We could store the last thumbnail, and use that as the first-image-seen-when-the-UI-is-shown, before the live stream takes over. Upside is that we stop storing high-res images in Window objects, so our RAM footprint drops. We only store the small thumbnails. Downside is that if the user display the thumbnail on another monitor, then it may stretch it. So the user would see a flash of a stretched pixelized thumbnails, then it would become high-res. It must not be that common to have users change between multiple monitors with widely different resolutions, so I think I would like this optimization.

damiansire:
Hi @lwouis,

Apologies if I'm missing something, I'm still learning Mac development and the macOS platform, so I'm using these PRs to gain experience

Thank you for the detailed explanation! I understand now that your optimization focuses on **window thumbnails** (screenshots), while my changes in `feature/image-compression` address **app icons** specifically.

After reviewing your comments and the codebase, I've implemented a complementary optimization for **app icons** that follows the same principles you outlined:

1. **Conditional compression based on App Icons mode**
   - When `Preferences.appearanceStyle == .appIcons`: stores high-res icons (GPU scaling)
   - Otherwise: compresses to optimal size (CPU resizing, memory savings)

2. **Monitor-based sizing heuristic**
   - Created `IconSizeCalculator` that calculates optimal size across all connected monitors
   - Uses max scale factor + 20% buffer (similar to your "X * 120%" suggestion)
   - Caches the calculation for performance

3. **Monitor change detection**
   - Invalidates cache when monitors are plugged/unplugged via `NSApplication.didChangeScreenParametersNotification`

### Key differences from window thumbnails:

- **Target**: `appIcon.updateWithResizedCopy()` in `ThumbnailView.swift:320`
- **Does NOT affect**: `thumbnail.updateWithResizedCopy()` (lines 339, 343),  your window screenshot optimization
- **Tradeoff**: One-time CPU resize vs ongoing memory savings (acceptable for icons since they load once)

For users NOT in App Icons mode, this should reduce app icon memory from ~80-120 MB (1024x1024 icons) to ~5-25 MB depending on size preference, without affecting your window thumbnail optimizations. (*1)

The implementations are complementary and don't conflict. Let me know if you'd like me to adjust anything! 

(*1) These are theoretical estimates based on assumptions. Let me attempt to create a real benchmark to measure the actual impact

https://github.com/lwouis/alt-tab-macos/pull/5146/commits/aca71f94ea8309a2575a68d00a55cff1302fa8d6

lwouis:
Please see https://github.com/lwouis/alt-tab-macos/issues/4255#issuecomment-3708086398

lwouis:
I'm working on implementing it very similarly to this PR. I'll hopefully release it in the next release. There were more edge-cases than I thought. These app icons can end up in many places. I've shipped this for thumbnails. App Icons are next 🚀 

Thank you for pushing for it! It is a big win for users 👍

Issue #5149, opened 2025-12-24
$200 bounty: AltTab can't open Telegram


kodjima33:
Telegram is my top-2 used app

But Alttab doesn't open it due to issue #4192 

I'll give $200 cash bounty to whoever will fix it

https://github.com/lwouis/alt-tab-macos/issues/4192

thehaffk:
works for ayugram desktop for me

aybanda:
@kodjima33 
Can you review this and confirm if it resolves the issue?

lwouis:
@aybanda activating an app with bring forward its windows. I don't think your code change fits the bill, unfortunately. Today we use a private API to focus the window without bringing anything else to the front. We use a private API to do it

aliraza556:
@lwouis Could you please assign this to me when you get a chance? Thank you.

aliraza556:
@lwouis Thanks

aliraza556:
@kodjima33 @lwouis @lwouis Please review this PR https://github.com/lwouis/alt-tab-macos/pull/5152

lwouis:
@aliraza556 my remark here https://github.com/lwouis/alt-tab-macos/issues/5149#issuecomment-3691658571 applies to your PR as well. We can't activate apps. If they have 100 open windows, all those will be focused. It will mess the user visually, and it will mess the recently-focused order in AltTab. That is, AltTab will show all 100 windows on the left, since they were just recently focused.

dluffy56:
~~Hi @lwouis can i attempt this issue? (asking because it's assigned)~~

lwouis:
Anyone can work on this, of course

kodjima33:
@lwouis can you pls check this PR? 

Love your app so much but can't use it anymore cuz of this - neither telegram not even apple notes don't get switched sometimes 

Had to switch to shortcuts instead

lwouis:
I checked the PRs. See my comments above. They activate the apps instead of focusing the window. To me it's not doing what it needs to. Activating apps creates problems

ozgeozturkapple-max:
hey

there is um a feature of macos that reveals every open wondow so you can choise between thoose why would you not use this? i forgot the features name and etc but probably there is a issue with thoose codeing _use macos s feature get used to that_ 

and thats how you solve a issue!

ozgeozturkapple-max:
that is not a spam you ll be seeing this and then i ll be asking what is a spam to you?


deal?

see who you re

all

[...]

PR #5152, opened 2025-12-25
Fix: Focus switching not working for Teams, Telegram and other Electron apps (#4192)

### LINKED ISSUE #4192: AltTab doesn't shift focus to certain apps

* macOS version: 15.3 (24D60)

aliraza556:
## Problem

Apps like Microsoft Teams, Telegram, and other Electron-based apps don't receive focus when selected from the AltTab carousel. Focus stays on the previously active application instead of switching to the selected window.

Reported in #4192.

## Root Cause

The low-level SkyLight API calls (`_SLPSSetFrontProcessWithOptions`, `makeKeyWindow`, `focusWindow`) work correctly for most apps but are not sufficient for some Electron-based applications that have custom window management implementations.

## Solution

Enhanced the `focus()` method in `Window.swift` with the following improvements:

1. **Unminimize first**: If the window is minimized, unminimize it before any focus attempt with a 100ms delay for the animation
2. **Preserve existing behavior**: Keep the original SkyLight API calls for apps that work correctly with them
3. **Add fallback activation**: Call `runningApplication.activate(options: .activateIgnoringOtherApps)` after the low-level calls to ensure stubborn apps are brought to the foreground
4. **Retry AX raise action**: After activation, retry `focusWindow()` to ensure the *correct* window is focused

## Testing

- [x] Tested with Microsoft Teams
- [x] Tested with Telegram
- [x] Verified no regression with standard macOS apps (Finder, Safari, etc.)
- [x] Tested with minimized windows
- [x] Tested with multiple monitors

## Related Issues

Fixes #4192 , #5149

### DIFF
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -197,14 +197,43 @@ class Window {
             // macOS bug: when switching to a System Preferences window in another space, it switches to that space,
             // but quickly switches back to another window in that space
             // You can reproduce this buggy behaviour by clicking on the dock icon, proving it's an OS bug
+            //
+            // For some apps (e.g. Teams, Telegram, and other Electron-based apps), the low-level SkyLight
+            // API calls alone are not sufficient to properly focus the window. We need to also call
+            // NSRunningApplication.activate() to ensure the app is properly activated.
+            // See: https://github.com/lwouis/alt-tab-macos/issues/4192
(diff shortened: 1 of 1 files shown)


aliraza556:
@lwouis Please review this PR.

PR #5154, opened 2025-12-26
Fix: Fallback to app activation only when CGWindowID is invalid

### LINKED ISSUE #4192: AltTab doesn't shift focus to certain apps

aybanda:
Fixes #4192 and #5149

- Always try private APIs first to focus the specific window (no activation)
- Verify whether the focus actually landed on that window
- Only fallback to app activation if private APIs fail (or no CGWindowID)


kodjima33:
@lwouis can you pls check this PR? 

Love your app so much but can't use it anymore cuz of this - neither telegram not even apple notes don't get switched sometimes 

Had to switch to shortcuts instead

lwouis:
subsumed by #5149 and related PRs

Issue #3639, opened 2024-10-01
Browser Tab Support


scottkuhl:
Consider adding an option to display each open browser tab as its own selection.

lwouis:
https://github.com/lwouis/alt-tab-macos/issues/2957#issuecomment-1793576113

Issue #2957, opened 2023-11-04
Treating browser tabs as windows


craigmartell:
It would be great if browser tabs could be treated as windows - just like the way finder tabs are treated as windows when "Show standard tabs as windows" is checked.  If this is suppose to be the case already, then there is a bug using 6.64.0 on Sonoma.  

Thanks for an AMAZING project.   If there is a way for me to donate money, I would be happy to.

craigmartell:
Or, instead of treating it like a window, at least allowing me to alt-tab between tabs.  Often I need to quickly go back and forth between two tabs in my workflow.  Work around is to move the tabs to windows.  But being able to go back and forth like Witch (from ManyTricks) would be really cool.

lwouis:
Hi,

Web browsers use custom tabs. We can't handle them. We can't screenshot them for instance. It's technically not possible. This is why the preference says "Show **standard** tabs as windows". It only works if the app uses standard tabs. We can't show any custom drawn UI elements that looks like a tab. Unfortunately all major web browser today use non-standard tabs, Safari included.

Thank you

Issue #1790, opened 2022-07-17
See tabs of Safari in AltTab just like in Windows 10/11 with MS Edge


Arnav-arw:
**Is your feature suggestion related to a problem? Please describe.**
Quickly changing tabs in Finder is very good. It would be a great quality of life improvement to get that feature for Safari. 

**Describe the solution you'd like**
Basically have an option for showing tabs of Safari (just like for Finder) in the preferences page. Also to avoid clutter, we can option to choose from showing only 3 tabs, 6 tabs or custom tabs.

lwouis:
Please see https://github.com/lwouis/alt-tab-macos/issues/1755#issuecomment-1173111180

Issue #1755, opened 2022-07-03
Tabs!


mB-PiBox:
**Is your feature suggestion related to a problem? Please describe.**
Sucks not being able to cycle through tabs..

**Describe the solution you'd like**
Add the option tab display for safari, doesn't even have to include thumbnails.
Down arrow key on Safari thumbnail would jump into list of open tabs on that window.
Or for mouse/trackpad users, right clicking Safari thumbnail while Option + Tab is help would jump into list of open tabs on that Safari window.

lwouis:
AltTab already supports showing tabs as windows (Preferences > Appearance > Show standard tabs as windows), but it only works if the app is using standard tabs.

Standard tabs means the app uses the standard macOS/AppKit APIs to create tabs. Finder uses that, Terminal.app uses that. Safari, Chrome, Firefox don't use that. Instead they draw custom graphics. This means there is no way for AltTab to get information about these tabs. We could probably get at least the titles of the tabs, but we can't get the screenshots, that's for sure. And getting the titles would be custom work for each web browser, and it would break anytime they do a redesign of their UI. We won't do that.

it's unfortunate that web browsers are the most used apps these days and they don't use standard tabs, especially Safari being maybe by Apple not using their own API, but c'est la vie :/

mB-PiBox:
So then just do it for Safari .. don't care about Chrome (Malware) or any of the janky browsers based on it.. I mean, after all, if folks wanna use Chrome they should get a Chromebook 🤣

Witch gets the titles of the tabs; don't need screenshots of the tab. That seems like a waste of system resource anyways. Heck if hiding the thumbnails for the windows (apps) was an option, I would check that box real fast.. I don't see the point of the thumbnail anyways, at least on my MBP, they are to small to make out what's in them anyways, at least any text small images..

lwouis:
> Heck if hiding the thumbnails for the windows (apps) was an option, I would check that box real fast.. I don't see the point of the thumbnail anyways, at least on my MBP, they are to small to make out what's in them anyways, at least any text small images..

There is a preference you can check to not show thumbnails. There are also many preferences to change the size of many UI components including thumbnails, if you have issues seeing them clearly. All these are in Preferences > Appearance.

PR #5157, opened 2025-12-27
feat: show browser tabs from browsers as separate windows

rekcah5:
## Summary
Adds browser tab switching to AltTab. Individual tabs from supported browsers show up as separate entries in the switcher, with favicons and previews.
## How Tab Previews Work
Tabs initially show their favicon since there's no way to screenshot a background tab. When you select a tab through AltTab, a preview is captured after a short delay (to let the browser render) and cached. Next time you trigger AltTab, that tab will show its cached preview.
## Automation Permission
This feature uses AppleScript/ScriptingBridge to query browser tabs, which requires automation permission. macOS will prompt for permission the first time AltTab tries to access a browser's tabs. The prompt appears per-browser (once for Chrome, once for Arc, etc.) and only triggers when "Show browser tabs" is enabled and that browser is running.
Added `com.apple.security.automation.apple-events` entitlement and `NSAppleEventsUsageDescription` to handle this.
## Tested Browsers
- Chromium-based: Chrome, Chromium, Perplexity (other Chromium browsers like Brave, Edge should work but weren't tested)
- Safari
- Arc
- Dia
## Usage
Enable "Show browser tabs" in Preferences → Appearance → Customize Style
## Disclaimer
Vibe coded this for personal use. Arc and Dia will show ALL tabs (including pinned/sidebar tabs) since their APIs don't distinguish between tab locations. Your mileage may vary.

### DIFF
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -155,7 +155,7 @@ class Window {
     }
 
     func canBeClosed() -> Bool {
-        return !isWindowlessApp
+        return !isWindowlessApp && !isBrowserTab
     }
 
     func close() {
@@ -187,7 +187,7 @@ class Window {
     }
 
     func canBeMinDeminOrFullscreened() -> Bool {
-        return !isWindowlessApp && !isTabbed
+        return !isWindowlessApp && !isTabbed && !isBrowserTab
     }
 
     func minDemin() {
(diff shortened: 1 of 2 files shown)


mksoska:
Could someone please look at this PR? There have been multiple requests for this feature over the years: #3639 #5097 #4634 #2957 #1790 #1755. I'm sure many would benefit from this being merged.

lwouis:
Please see https://github.com/lwouis/alt-tab-macos/issues/5230#issuecomment-3783009993

I will not add this feature, because custom integration with other apps is too much to maintain for me. There is a lot more in thread I linked above. I'll also add an FAQ on the website to clarify this important human limitation.

Thank you

PR #5165, opened 2026-01-05
Fix multi-window Unity/similar app focus failure (#4192), Zoom/similar race condition, improve notification dots

### LINKED ISSUE #4192: AltTab doesn't shift focus to certain apps

chuttenh:
fix: Unity app focus (#4192 ), Zoom race condition, notification count cap

### Bug Fixes

* Swapping away from Unity/similar multi-window apps failed to change focus (should fix #4192 )
* Race condition during Zoom/Photoshop/similar apps with custom windows (revises previous fix)

### Features

* Improves notification dots (moves * cap up to >four digits)

Apologies that I have a very intense full time job and haven't had time to test under every circumstance
or match convention perfectly. I want to get this into circulation, though, since #4192 is breaking
for me, and the Zoom/Photoshop race condition has been a bear to track down. I've done extensive
testing on both of those prior to v7.39.0, which unfortunately stomped on several of the fixes.
I've done my best to re-test after a lot of merge conflicts, but I haven't had nearly as much time,
and I want to get this in over the holiday break before work takes over again. Apologies that I've
done my best, thanks for your work on this!

### DIFF
--- a/src/logic/events/AccessibilityEvents.swift
+++ b/src/logic/events/AccessibilityEvents.swift
@@ -48,6 +48,82 @@ class AccessibilityEvents {
             case kAXWindowResizedNotification,
                  kAXWindowMovedNotification: try windowResizedOrMoved(element, pid)
             default: return
+        }
+    }
+}
+
+private static func applicationActivated(_ element: AXUIElement, _ pid: pid_t) throws {
+    let initialFocusedWindow = try element.focusedWindow()
+    let initialWid = try initialFocusedWindow?.cgWindowId()
+
+    func applyFocusBump(_ axWindow: AXUIElement?, _ wid: CGWindowID?) {
+        DispatchQueue.main.async {
+            if let app = Applications.find(pid) {
+                if app.hasBeenActiveOnce != true {
+                    app.hasBeenActiveOnce = true
+                }
+                app.isHidden = false
+
+                var window: Window? = nil
+                if let axWindow, let wid {
+                    // Try to update recency immediately if we can identify the window
+                    window = Windows.updateLastFocus(axWindow, wid)?.first
+
+                    // If the window isn't tracked yet, mirror the focusedWindowChanged fallback to create it
+                    if window == nil {
+                        AXUIElement.retryAxCallUntilTimeout(context: "activation wid:\(wid) pid:\(pid)", pid: pid, callType: .updateWindow) {
+                            if let (title, role, subrole, isMinimized, isFullscreen) = try axWindow.windowAttributes() {
+                                let position = try axWindow.position()
+                                let size = try axWindow.size()
+                                let level = wid.level()
+                                DispatchQueue.main.async {
+                                    if let app2 = Applications.find(pid),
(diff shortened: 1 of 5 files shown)


lwouis:
Thank you very much for this PR! I'll review it soon 👍

src/logic/events/AccessibilityEvents.swift
```diff
@@ -87,62 +131,38 @@ class AccessibilityEvents {
...
-                                DispatchQueue.main.async {
-                                    if (!Windows.list.contains { $0.isEqualRobust(element, wid) }),
-                                       AXUIElement.isActualWindow(app, wid, level, title, subrole, role, size) {
-                                        let window = Window(element, app, wid, title, isFullscreen, isMinimized, position, size)
-                                        Windows.appendAndUpdateFocus(window)
-                                        App.app.refreshOpenUi([window], .refreshUiAfterExternalEvent)
-                                    }
+            Windows.updateLastFocusDebounced(element, wid, pid: pid, debounce: 60, requireFrontmost: false)
+            
+            // Also schedule a fallback to create the window if it doesn't exist yet after a short delay
```
lwouis: What does it mean "the window doesn't exist yet"? I thought that when this method is called `windowCreated(_ element: AXUIElement, _ pid: pid_t)`, we have a pid, and we have the AXUIElement of the created window.

Could you please clarify what we are waiting 120ms for? Which object or attribute is not ready initially?

src/ui/main-window/ThumbnailFontIconView.swift
```diff
@@ -33,20 +40,197 @@ class ThumbnailFontIconView: ThumbnailTitleView {
...
+    private func setSymbol(_ symbol: String) {
+        overlayText = nil
+        overlayLabel?.isHidden = true
+        needsDisplay = true
+        replaceCharIfNeeded(symbol)
     }
 
     func setStar() {
-        replaceCharIfNeeded(Symbols.circledStar.rawValue)
+        setSymbol(Symbols.circledStar.rawValue)
```
lwouis: I'm curious about why you picked a white dot in the red circle.

I looked at Apple guidelines, and I can't find clear guidelines as to what apps should do when they can show a straight label. I created a test app that sets a Dock badge. I tried to set `NSApplication.shared.dockTile.badgeLabel` to these values:
- `nil`
- `""`
- `"1"`
- `"10"`
- `"100"`
- `"1000"`
- `"10000"`
- `"test"`

It either shows the value litterally, or it shows no badge, for the `""` and `nil` values. Not an empty red disk, or a red disk with a white dot. No disk at all.

What do you think?

lwouis:
Something that I need, in order to evaluate this work, beyond the comments I've left on the code, would be to describe in this PR the scenarios that you've tested. You seem to be able to reproduce these focus issues deterministically.

If I can reproduce on my side, I can help with this fix. It I can't reproduce on my side, all I can do is to theorize about the code, and how it may help.

Thank you!

PR #5195, opened 2026-01-13
fix: mouse hover detection for hidden/windowless apps

emptiestplace:
Mouse hover to select wasn't working for hidden apps or apps without windows. Hovering over these apps didn't show the blue highlight, so you couldn't hover and release the modifier key to focus them - clicking still worked. (Tested in App Icons and Titles modes, but likely affected all modes.)

The issue was that `findTarget` iterated `Windows.list.count` times over `recycledViews`, but the actual displayed subviews can be fewer due to the `shouldShowTheUser` filter in `layoutThumbnailViews`. This mismatch meant hover detection was checking views that weren't in the view hierarchy.

Fixed by iterating directly over `documentView.subviews` instead.

### DIFF
--- a/src/ui/main-window/ThumbnailsView.swift
+++ b/src/ui/main-window/ThumbnailsView.swift
@@ -308,8 +308,7 @@ class ScrollView: NSScrollView {
     }
 
     private func findTarget(_ location: NSPoint) -> ThumbnailView? {
-        for i in 0..<Windows.list.count {
-            let view = ThumbnailsView.recycledViews[i]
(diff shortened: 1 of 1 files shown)


lwouis:
Thank you very much for sharing this fix! I could reproduce the issue, and confirm that the fix works.

It will be shipped in the next release 🎉

himudigonda (approved):
Thanks for this fix, @emptiestplace! I was running into this exact same issue with hidden apps, so I'm really glad to see this addressed.

lwouis:
Closing this PR as the fix is live.
Thanks again!

PR #5212, opened 2026-01-17
fix(controls-tab): make enter key focus selected window

AbdelrahmanHafez:
This PR add Enter key (both main keyboard Return and numpad Enter) as additional shortcuts to focus the selected window when AltTab is active, matching the existing Space key behavior. re last bit of https://github.com/lwouis/alt-tab-macos/issues/210#issuecomment-3763096245

I verified that it works properly on my machine, as well as unit-tests.

### DIFF
--- a/src/ui/preferences-window/tabs/controls/ControlsTab.swift
+++ b/src/ui/preferences-window/tabs/controls/ControlsTab.swift
@@ -10,6 +10,8 @@ class ControlsTab {
         "holdShortcut2": { App.app.focusTarget() },
         "holdShortcut3": { App.app.focusTarget() },
         "focusWindowShortcut": { App.app.focusTarget() },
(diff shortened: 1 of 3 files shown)


lwouis:
Hi,

Users can already customize the key to focus, to use the Enter key, instead of the Space key:

image

I see issues with this PR:
* Hard-coding the "enter" key to focus. What if the user already uses it for another shortcut they prefer? What if they don't want enter to focus the window? The current system allows them to select
* Conflicts with existing shortcuts is not handled. We have code to avoid 2 shortcuts with the same keys for example

I think the current status quo works. People who prefer Enter to Space can customize it. There is no hard-coded, un-customizable, mandatory, shortcuts.

For these reasons, I can't merge 🙇 

Thank you

Issue #5177, opened 2026-01-08
Lag when opening the switcher


drauziooppenheimer:
**Describe the bug**

I didn't find a way to consistently reproduce this, but sometimes, when switching between windows, the AltTab window area would freeze for a second or two after the window was switched, and the previous window would be cropped to the AltTab's location.

**Steps to reproduce the bug**

Quickly keep switching between windows using alt-tab, eventually it will freeze with the previous window being show in its location.

**Your environment**

* AltTab version: 8.0.0 (But this has been happening for a few versions at this point)
* macOS version: 26.2
* Other relevant info: e.g. using external keyboard, multiple monitors, etc

lwouis:
Hi,

I can't reproduce this. Any idea on how to reproduce? When does it happen for you? Any specific pattern like your system is under heavy load? etc

Thank you

SamuelGadiel:
In the video, I'm pressing cmd+tab to switch between two windows, basically just spamming the cmd+tab
I'm not holding it down, just hitting the keys and immediately releasing it

In my case, sometimes the alt tab pop-up remain visible even after the windows had already been switched

https://github.com/user-attachments/assets/9f482eaa-e0a4-4563-8b42-db4a9a2da19a

lwouis:
Thank you. Could you please share a video of your fingers as well hitting the keys? maybe using your phone to capture both fingers and screen

Thank you

baroca:
I also experiencing this problem. It seems to happen at random. Sometimes the alt tab screen freezes while fading out keeping the semi transparent window visible for a second or two.

I think it started after upgrading to MacOS Tahoe.

SamuelGadiel:
@lwouis I recorded me alt-tabbing for 47 seconds. Then another weird thing happened at the end of the video. The alt-tab completed, but instead of keep the pop up showing, it bugged the pop up for a second before hiding it.

In the first video i posted I was alt-tabbing just like in this new video
You can see that it took me a bunch of tries to make it happen.

https://github.com/user-attachments/assets/fa994f1c-2061-45dc-a160-2bb061a0a482

SamuelGadiel:
By the way, here's my specs

Macbook Pro M3 Pro, 18GB ram
macOS: 15.6.1 (24G90)
AltTab: 8.0.0 (this was happening on version 7 as well)

Jeordy:
[...]

PR #5251, opened 2026-01-27
CPU Hangs on macOS 26.2 (Tahoe)

cdbattags:
References:

https://github.com/lwouis/alt-tab-macos/issues/5177

### DIFF
--- a/src/logic/Spaces.swift
+++ b/src/logic/Spaces.swift
@@ -24,6 +24,25 @@ class Spaces {
         return CGSCopyWindowsWithOptionsAndTags(CGS_CONNECTION, 0, spaceIds as CFArray, options.rawValue, &set_tags, &clear_tags) as! [CGWindowID]
     }
 
+    /// Build a reverse mapping from window ID to space IDs
+    /// This is much faster than calling cgWindowId.spaces() for each window individually
+    /// Makes O(number of spaces) WindowServer calls instead of O(number of windows)
+    static func buildWindowToSpacesMap() -> [CGWindowID: [CGSSpaceID]] {
+        var windowToSpaces: [CGWindowID: [CGSSpaceID]] = [:]
+
+        // For each space, get all windows in that space
+        for (spaceId, _) in idsAndIndexes {
+            let windowsInThisSpace = windowsInSpaces([spaceId], true)
+
+            // Add this space to each window's space list
+            for windowId in windowsInThisSpace {
+                windowToSpaces[windowId, default: []].append(spaceId)
+            }
+        }
+
+        return windowToSpaces
+    }
+
     static func refresh() {
         let now = DispatchTime.now()
         let timeSinceLastRefresh = Double(now.uptimeNanoseconds - lastRefreshTime.uptimeNanoseconds) / 1_000_000_000
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -251,17 +251,25 @@ class Window {
         return application.localizedName ?? ""
     }
 
-    func updateSpacesAndScreen() {
+    func updateSpacesAndScreen(windowToSpacesMap: [CGWindowID: [CGSSpaceID]]? = nil) {
         // macOS bug: if you tab a window, then move the tab group to another space, other tabs from the tab group will stay on the current space
         // you can use the Dock to focus one of the other tabs and it will teleport that tab in the current space, proving that it's a macOS bug
         // note: for some reason, it behaves differently if you minimize the tab group after moving it to another space
-        updateSpaces()
+        updateSpaces(windowToSpacesMap: windowToSpacesMap)
         updateScreenId()
     }
 
-    private func updateSpaces() {
+    private func updateSpaces(windowToSpacesMap: [CGWindowID: [CGSSpaceID]]? = nil) {
         guard let cgWindowId else { return }
-        let spaceIds = cgWindowId.spaces()
+
+        // Use pre-computed mapping if available (much faster), otherwise fall back to individual API call
+        let spaceIds: [CGSSpaceID]
+        if let map = windowToSpacesMap, let cachedSpaceIds = map[cgWindowId] {
+            spaceIds = cachedSpaceIds
+        } else {
+            spaceIds = cgWindowId.spaces()
+        }
+
         self.spaceIds = spaceIds
         self.spaceIndexes = spaceIds.compactMap { spaceId in Spaces.idsAndIndexes.first { $0.0 == spaceId }?.1 }
         self.isOnAllSpaces = spaceIds.count > 1
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -88,12 +88,35 @@ class Windows {
             lazy var cgsWindowIds = Spaces.windowsInSpaces(spaceIdsAndIndexes)
             lazy var visibleCgsWindowIds = Spaces.windowsInSpaces(spaceIdsAndIndexes, false)
 
+            // Build window→spaces mapping once instead of querying each window individually
+            // This makes O(number of spaces) API calls instead of O(number of windows)
+            // Typically 2-10 spaces vs 64 windows = 6-30x fewer WindowServer API calls!
+            let windowToSpacesMap = section.step("Build window→spaces map") {
+                return Spaces.buildWindowToSpacesMap()
+            }
+
             section.step("Window loop") {
+                var detectTabbedTime: Double = 0
+                var updateSpacesTime: Double = 0
+                var refreshShouldShowTime: Double = 0
+
                 for window in list {
+                    let start1 = BaseLogger.preciseTimestamp()
                     detectTabbedWindows(window, cgsWindowIds, visibleCgsWindowIds)
-                    window.updateSpacesAndScreen()
+                    detectTabbedTime += BaseLogger.elapsedMilliseconds(from: start1)
+
+                    let start2 = BaseLogger.preciseTimestamp()
+                    window.updateSpacesAndScreen(windowToSpacesMap: windowToSpacesMap)
+                    updateSpacesTime += BaseLogger.elapsedMilliseconds(from: start2)
+
+                    let start3 = BaseLogger.preciseTimestamp()
                     refreshIfWindowShouldBeShownToTheUser(window)
+                    refreshShouldShowTime += BaseLogger.elapsedMilliseconds(from: start3)
                 }
+
+                section.log("  - detectTabbedWindows total: \(String(format: "%.2f", detectTabbedTime))ms")
+                section.log("  - updateSpacesAndScreen total: \(String(format: "%.2f", updateSpacesTime))ms")
+                section.log("  - refreshIfWindowShouldBeShownToTheUser total: \(String(format: "%.2f", refreshShouldShowTime))ms")
             }
 
             section.step("refreshWhichWindowsToShowTheUser") {
--- a/src/ui/main-window/PreviewPanel.swift
+++ b/src/ui/main-window/PreviewPanel.swift
@@ -33,7 +33,7 @@ class PreviewPanel: NSPanel {
         let newFrame = NSRect(origin: position, size: size)
         let frameChanged = newFrame != cachedFrame
         let idChanged = id != currentId
-        
+
         if idChanged || frameChanged {
             if frameChanged {
                 cachedFrame = newFrame
@@ -43,7 +43,7 @@ class PreviewPanel: NSPanel {
                 previewView.updateContents(preview, size)
             }
         }
-        
+
         if idChanged || !isVisible {
             if Preferences.previewFadeInAnimation && !isVisible {
                 alphaValue = 0
--- a/src/ui/main-window/ThumbnailsPanel.swift
+++ b/src/ui/main-window/ThumbnailsPanel.swift
@@ -34,9 +34,9 @@ class ThumbnailsPanel: NSPanel {
     }
 
     private var lastLayoutTime: DispatchTime = .now()
-    
+
     func updateContents() {
-        // Workaround for macOS 15+ hang: throttle full layout passes during rapid interaction
+        // Throttle full layout passes during rapid interaction
         // External events (window title changes, moves) can trigger expensive layouts while cycling
         // Skip if we just did a layout less than 100ms ago
         let now = DispatchTime.now()
@@ -45,7 +45,7 @@ class ThumbnailsPanel: NSPanel {
             return
         }
         lastLayoutTime = now
-        
+
         CATransaction.begin()
         defer { CATransaction.commit() }
         CATransaction.setDisableActions(true)


cdbattags:
~I'll update the commit convention shortly!~

Edit: Done

cdbattags:
This should be ready to go! And I have another 80% speed up coming shortly 😉

cdbattags:
Will need your opinion on leaving the performance logger (only debug builds)

cdbattags:
Summary from Cursor/Sonnet 4.5:

<details>

# Staged Changes Summary - macOS 15.2 Hang Fix

## 🎯 PRIMARY FIX: External Event Flood Throttling

### Root Cause Identified
While rapidly pressing Tab, external accessibility events (window title changes, moves, etc.) from background apps were triggering **hundreds of concurrent expensive operations** (`updatesBeforeShowing()` calls at 70-120ms each), causing 2-9 second hangs.

**Before Fix:** 275+ concurrent calls → 6-9 second cumulative hang  
**After Fix:** ~50 sequential calls with 150ms throttle → smooth, snappy performance

## 📝 FILE-BY-FILE BREAKDOWN

### 1. **src/ui/App.swift** (+45 lines)
**Purpose:** Core external event throttling mechanism

**Key Changes:**
- **Added serial dispatch queue** for external events
  ```swift
  private let externalEventQueue = DispatchQueue(label: "com.lwouis.alt-tab-macos.externalEventQueue")
  ```
- **Added 150ms throttle** on external event refreshes
  ```swift
  private var lastExternalEventRefreshTime: DispatchTime?
  private let externalEventRefreshThrottle: TimeInterval = 0.15
  ```
- **Thread-safe external event handling** in `refreshOpenUi()`
  - Prevents race conditions
  - Prevents concurrent execution
  - Reduces 275+ calls to ~50 calls

**Impact:** 🎯 **PRIMARY FIX** - Eliminates the hang completely

### 2. **src/api-wrappers/PerfLogger.swift** (NEW FILE, +34 lines)
**Purpose:** Performance logging infrastructure

**Key Features:**
- **Debug-only logging** using `#if DEBUG` directive
- Writes to `/tmp/alttab_perf.log` with timestamps
- Automatic `[PERF]` prefix on all messages
- Zero cost in Release builds

**Usage:**
```swift
PerfLogger.log("Operation took 50ms")
// Output: [14:30:45] [PERF] Operation took 50ms
```

**Impact:** Enables future performance debugging without affecting production

### 3. **src/logic/Spaces.swift** (+78 lines modified)
**Purpose:** Performance logging for WindowServer API calls

**Key Changes:**
- Added timing measurements for expensive operations:
  - `CGSCopyWindowsWithOptionsAndTags()` - WindowServer API
  - `CGSCopyManagedDisplaySpaces()` - Space enumeration
  - Cache hit/miss logging
  - Throttle logging (200ms throttle on space refresh)

**Example Output:**
```
[PERF] windowsInSpaces: cache HIT for 3 spaces
[PERF] Spaces.refresh: THROTTLED (last refresh 150ms ago)
[PERF] CGSCopyManagedDisplaySpaces took 5.67ms, found 3 spaces
```

**Impact:** Visibility into WindowServer performance, confirms APIs are fast

### 4. **src/logic/Windows.swift** (+61 lines)
**Purpose:** Performance logging for window list operations

**Key Changes:**
- Added comprehensive timing for `updatesBeforeShowing()`:
  - Window loop timing (70-120ms typical)
  - Sort timing (~1ms)
  - Individual operation breakdown
- Added timing for `cycleSelectedWindowIndex()` (Tab press)
- Added timing for `updateSelectedAndHoveredWindowIndex()`

**Example Output:**
```
[PERF] ===== updatesBeforeShowing START =====
[PERF] updatesBeforeShowing: Window loop took 85.90ms
[PERF] ===== updatesBeforeShowing TOTAL: 85.90ms =====
[PERF] ===== cycleSelectedWindowIndex TOTAL: 1.40ms =====
```

**Impact:** Identifies bottlenecks (window loop), confirms Tab press is fast

### 5. **src/ui/main-window/ThumbnailsPanel.swift** (+13 lines)
**Purpose:** Layout throttling (existing optimization, kept)

**Key Changes:**
- 100ms throttle on `updateContents()` to prevent layout thrashing
- Comments documenting the fix
- Reference to issue #5177

**Impact:** Secondary defense against external event layouts

### 6. **src/ui/main-window/ThumbnailView.swift** (+105 lines)
**Purpose:** Caching optimizations (existing, kept from previous fixes)

**Key Changes:**
- Image caching (skip GPU updates if unchanged)
- Label frame caching (skip recalculation if unchanged)
- Layer property caching (skip CALayer commits if unchanged)
- Comments documenting each optimization

**Impact:** Eliminates redundant GPU operations during hover/selection

### 7. **src/ui/main-window/ThumbnailTitleView.swift** (+30 lines)
**Purpose:** Constraint optimization (existing, kept from previous fixes)

**Key Changes:**
- Width caching (skip constraint updates when unchanged)
- Constraint reuse (update `.constant` instead of remove/recreate)
- Comments documenting the optimization

**Impact:** 90% reduction in constraint operations

### 8. **src/ui/main-window/PreviewPanel.swift** (+23 lines)
**Purpose:** Preview frame caching (existing, kept from previous fixes)

**Key Changes:**
- Cache preview frame to skip redundant repositioning
- Only update what actually changed (frame vs content)
- Comments documenting the optimization

**Impact:** Skips expensive window repositioning when preview unchanged

### 9. **scripts/install_local_build.sh** (NEW FILE, +95 lines)
**Purpose:** Local development build installation script

**Features:**
- Automated backup of existing AltTab.app
- Install Debug/Release builds to /Applications
- Clear quarantine attribute
- Optional launch after install
- Colored output for better UX

**Impact:** Faster development workflow

### 10. **scripts/build_app.sh** (+55 lines)
**Purpose:** Improved build script with launch capability

**Features:**
- Added `--launch` flag to automatically open app after build
- Better error handling
- Integrated with install_local_build.sh

**Impact:** One-command build-and-test workflow

### 11. **alt-tab-macos.xcodeproj/project.pbxproj** (+20 lines)
**Purpose:** Xcode project configuration

**Key Changes:**
- Added PerfLogger.swift to build
- Updated file references

**Impact:** PerfLogger included in builds

### 12. **src/logic/SystemPermissions.swift** (+12 lines)
**Purpose:** Minor logging additions

**Changes:**
- Added debug logging for permission checks

**Impact:** Better diagnostics

### 13. **src/api-wrappers/CGWindow.swift** (-20 lines)
**Purpose:** Code cleanup

**Changes:**
- Removed old/unused code

**Impact:** Cleaner codebase

### 14. **src/logic/Applications.swift** (-20 lines)
**Purpose:** Code cleanup

**Changes:**
- Removed old/unused code

**Impact:** Cleaner codebase

## 📊 PERFORMANCE RESULTS

### Before Fix
- **Initial trigger:** Smooth (first Tab press)
- **Rapid Tab presses:** 2-9 second hangs
- **External events:** 275+ concurrent `updatesBeforeShowing()` calls
- **User experience:** Frustrating, unusable during rapid interaction

### After Fix
- **Initial trigger:** Smooth (unchanged)
- **Rapid Tab presses:** Smooth, snappy, responsive ✅
- **External events:** ~50 sequential calls, 572 throttled (92% reduction)
- **User experience:** "Feels snappy now!" - User feedback

### Metrics
- **Reduction in operations:** 92% (275 → 50 calls)
- **Throttle effectiveness:** 572 calls prevented
- **No performance regressions:** Tab press still <2ms
- **No concurrent execution:** Serial queue prevents race conditions

## 🔍 WHAT'S NOT INCLUDED (Future Work)

The following untracked documentation files are NOT staged:
- `DIAGNOSTIC_LOGGING.md` - Investigation notes
- `DIAGNOSTIC_PLAN.md` - Investigation plan
- `FIX_PERMISSION_DETECTION.md` - Permission investigation
- `HANG_FIX_*.md` - Various fix documentation
- `OPTIMIZATION_SUMMARY.md` - Optimization notes
- `PERFORMANCE_TUNING_GUIDE.md` - Future optimization opportunities

**Reason:** These are internal investigation/planning documents. The actual code changes contain sufficient inline documentation.

## ✅ TESTING VERIFICATION

### Test Scenarios Passed
1. ✅ Open AltTab with 50+ windows
2. ✅ Rapidly press Tab 20-30 times (main test)
3. ✅ Verify smooth, responsive navigation
4. ✅ No lag even if background apps are updating
5. ✅ External events properly throttled
6. ✅ No race conditions or concurrent execution

### Performance Logging Verification
```bash
# Debug build - logging enabled
tail -f /tmp/alttab_perf.log

# Release build - zero logging overhead (compilation flag)
# No log file created, no NSLog calls
```

## 🚀 COMMIT READINESS

### Checklist
- ✅ All changes tested and working
- ✅ Performance verified (92% reduction in operations)
- ✅ User feedback positive ("feels snappy now!")
- ✅ Code properly documented with inline comments
- ✅ Debug-only logging has zero Release build cost
- ✅ No regressions in existing functionality
- ✅ Build scripts improved for development workflow

### Recommended Commit Message

Fix: Eliminate 2-9s hang during rapid Tab cycling on macOS 15.2

Root Cause:
External accessibility events from background apps (window title changes,
moves, etc.) were triggering hundreds of concurrent expensive window list
updates (updatesBeforeShowing() at 70-120ms each) during rapid Tab presses,
causing cumulative 6-9 second hangs.

Solution:
1. Serial dispatch queue for external event handling
2. 150ms throttle on external event refreshes
3. Thread-safe implementation prevents race conditions

Results:
- 92% reduction in operations (275+ → ~50 sequential calls)
- Smooth, responsive rapid Tab navigation
- Zero performance regression on normal usage
- Zero cost in Release builds (debug logging only)

Related optimizations kept from previous investigation:
- 100ms layout throttle in ThumbnailsPanel
- Image/label/layer caching in ThumbnailView
- Constraint reuse in ThumbnailTitleView
- Preview frame caching in PreviewPanel

Testing:
- Verified with 50+ windows across multiple spaces
- Rapid Tab press test: smooth and responsive
- External event flood: properly throttled
- Performance logging confirms 92% reduction

Fixes: #5177 (macOS 15.2 hang bug)

## 📝 NOTES FOR REVIEWER

1. **The PRIMARY fix** is in `App.swift` (external event throttling)
2. **PerfLogger** is debug-only (`#if DEBUG`) - zero cost in Release
3. **Existing optimizations** from previous investigation are kept and working
4. **Build scripts** improved for faster dev workflow (quality of life)
5. **Documentation** is in code comments, not separate MD files
6. **Testing** has been thorough - user confirmed it's working

</details>

cdbattags:
I'll go back through and add a cleaner system for the performance stuff so we're not holding onto all those timing vars.

cdbattags:
Actually, lemme make this 3 separate PRs that build on each other to make life easier for you:

1. `App.swift`
2. `PerfLogger`
3. Bunch of other optimizations

```
~98% overall performance improvement
Sub-3ms cycle times (was 100-200ms)
99.97% faster window loop (0.03ms vs 70-120ms)
99.2% faster frame updates (0.4ms vs 50ms)
95%+ cache hit rates
```

This is a true measurement of the next PR.

src/ui/main-window/ThumbnailView.swift
```diff
@@ -19,6 +19,15 @@ class ThumbnailView: FlippedView {
...
     var maximizeIcon = TrafficLightButton(.fullscreen, NSLocalizedString("Fullscreen/Defullscreen window", comment: ""))
+    
+    // Cache for performance - avoid redundant image updates
+    // see https://github.com/lwouis/alt-tab-macos/issues/5177
+    private var cachedThumbnailImage: CGImage?
+    private var cachedAppIconImage: CGImage?
+    private var cachedLabelFrame: NSRect = .zero
+    private var cachedBackgroundColor: NSColor?
+    private var cachedBorderColor: NSColor?
+    private var cachedBorderWidth: CGFloat = 0
```
lwouis: Is there a way to avoid recomputation, that doesn't involve caching every sub-part of the view, and comparing?

Also, I wonder, does reassigning the same values really trigger computing on AppKit side? Like if the title color is red, and I re-assign red, is it really going to recompute the title rendering? I would imagine that AppKit has the concept of dirty properties to mark what's changed. They certainly seem to support it for layout. We needs the needDisplays, needsLayout methods, etc.

I wonder how the maintenability of all these cached values and checks all over the place. And their effectiveness as well. Maybe there is a better way, more holistic/structure way?

scripts/build_app.sh
```diff
@@ -1,6 +1,55 @@
 #!/usr/bin/env bash
 
-set -ex
+set -e
 
-set -o pipefail && xcodebuild -workspace alt-tab-macos.xcworkspace -scheme Release -derivedDataPath DerivedData | scripts/xcbeautify
-file "$BUILD_DIR/$XCODE_BUILD_PATH/$APP_NAME.app/Contents/MacOS/$APP_NAME"
+# Build script for AltTab
```
lwouis: Is this file, as well as `install_local_build`, modified to suit AI development? I'm not yet at the forefront of AI development, so it's a bit mysterious to read for me. Could you please share some insights into your workflow, and why this change was necessary? What it enabled?

src/api-wrappers/PerfLogger.swift
```diff
@@ -0,0 +1,34 @@
+import Foundation
+
+class PerfLogger {
```
lwouis: Couldn't we use the Logger we already have to measure performance? We use SwiftBeaver. Maybe there is a way to leverage it for this. After all, it already tracks timing with high precision. And it can log to console, disk, etc. It's very flexible. Maybe we could create a Logger.perfStart, and Logger.perfStop for example, that we print the difference. Just reusing SwiftBeaver instead of making an ad-hoc logger?

Maybe it's not a good idea. I'm just wondering

cdbattags:
All great comments! I will make sure to reduce all the noise and make sure that the code changes accurately reflect what you're asking for! Give me til late this evening and it will absolutely be suitable for you!

cdbattags:
I will continue to be very clear about making sure this just solves the immediate need and then we will talk 1v1 about where I think we can go further.

src/logic/Spaces.swift
```diff
@@ -6,24 +6,61 @@ class Spaces {
...
+    private static var lastRefreshTime: DispatchTime = .now() - .seconds(1)
+    private static let refreshThrottle: TimeInterval = 0.2
 
     static func isSingleSpace() -> Bool {
         return idsAndIndexes.count == 1
     }
 
     static func windowsInSpaces(_ spaceIds: [CGSSpaceID], _ includeInvisible: Bool = true) -> [CGWindowID] {
+        let cacheKey = "\(spaceIds)_\(includeInvisible)"
+        if let cached = windowsInSpacesCache[cacheKey] {
```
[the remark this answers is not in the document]
cdbattags: Fixed!

src/ui/App.swift
```diff
@@ -199,13 +199,54 @@ class App: AppCenterApplication {
...
+                    Windows.previewSelectedWindowIfNeeded()
+                    guard self.appIsBeingUsed else { return }
+                    Applications.refreshBadgesAsync()
+                }
+            }
+            return
         }
-        guard appIsBeingUsed else { return }
+        
+        // Non-external event path continues synchronously
```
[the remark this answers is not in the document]
cdbattags: It's working ok without this now but you decide!

src/ui/main-window/ThumbnailView.swift
cdbattags: Fixed!

cdbattags:
Should be cleaned up to your liking now!

cdbattags:
From my testing, this still doesn't reduce _all_ of the hanging as is. But it's a big improvement.

cdbattags:
I FOUND IT!

<details>

Window→Space Mapping Optimization

## Problem Identified

Performance testing revealed that `updatesBeforeShowing()` was taking 50-173ms, with **98.5% of the time** spent in `updateSpacesAndScreen()`:

```
Window loop: 173.3ms (worst case)
├─ detectTabbedWindows: 0.72ms (0.4%)
├─ updateSpacesAndScreen: 170.60ms (98.5%) ← THE BOTTLENECK
└─ refreshIfWindowShouldBeShownToTheUser: 1.88ms (1.1%)
```

## Root Cause

The `Window.updateSpaces()` method was calling `cgWindowId.spaces()` for **each window individually**:

- 64 windows × individual WindowServer API call = **64 expensive API calls**
- Each call took ~2-2.5ms in worst case = 128-160ms total
- The WindowServer API `CGSCopySpacesForWindows()` was being called once per window

## Solution

Instead of querying each window individually, we now:

1. **Build a reverse index once** using `Spaces.buildWindowToSpacesMap()`
   - Makes O(number of spaces) API calls instead of O(number of windows)
   - Typically 2-10 spaces vs 64 windows = **6-30x fewer WindowServer calls**

2. **Reuse the cached mapping** for all windows in the loop
   - Pass `windowToSpacesMap` to each `window.updateSpacesAndScreen()`
   - Dictionary lookup is nearly instant vs 2ms API call

## Implementation

### Added to `Spaces.swift`:
```swift
/// Build a reverse mapping from window ID to space IDs
/// Makes O(number of spaces) WindowServer calls instead of O(number of windows)
static func buildWindowToSpacesMap() -> [CGWindowID: [CGSSpaceID]] {
    var windowToSpaces: [CGWindowID: [CGSSpaceID]] = [:]

    for (spaceId, _) in idsAndIndexes {
        let windowsInThisSpace = windowsInSpaces([spaceId], true)
        for windowId in windowsInThisSpace {
            windowToSpaces[windowId, default: []].append(spaceId)
        }
    }

    return windowToSpaces
}
```

### Modified `Window.swift`:
```swift
func updateSpacesAndScreen(windowToSpacesMap: [CGWindowID: [CGSSpaceID]]? = nil) {
    updateSpaces(windowToSpacesMap: windowToSpacesMap)
    updateScreenId()
}

private func updateSpaces(windowToSpacesMap: [CGWindowID: [CGSSpaceID]]? = nil) {
    guard let cgWindowId else { return }

    // Use pre-computed mapping if available (much faster)
    let spaceIds: [CGSSpaceID]
    if let map = windowToSpacesMap, let cachedSpaceIds = map[cgWindowId] {
        spaceIds = cachedSpaceIds
    } else {
        spaceIds = cgWindowId.spaces()  // Fallback
    }

    self.spaceIds = spaceIds
    self.spaceIndexes = spaceIds.compactMap { spaceId in
        Spaces.idsAndIndexes.first { $0.0 == spaceId }?.1
    }
    self.isOnAllSpaces = spaceIds.count > 1
}
```

### Modified `Windows.swift`:
```swift
// Build window→spaces mapping once instead of querying each window individually
let windowToSpacesMap = section.step("Build window→spaces map") {
    return Spaces.buildWindowToSpacesMap()
}

section.step("Window loop") {
    for window in list {
        // ... detectTabbedWindows ...
        window.updateSpacesAndScreen(windowToSpacesMap: windowToSpacesMap)  // ← Use cached map
        // ... refreshIfWindowShouldBeShownToTheUser ...
    }
}
```

## Performance Results

### Before Optimization:
- **Best case**: 10-50ms
- **Typical case**: 50-120ms
- **Worst case**: **173ms** 🐌
- Bottleneck: `updateSpacesAndScreen` took 170ms (64 individual API calls)

### After Optimization:
- **Best case**: 7-13ms (**3-7x faster** 🚀)
- **Typical case**: 8-17ms (**3-7x faster** 🚀)
- **Worst case**: 88ms (**2x faster** 🚀)
- New breakdown: `Build map` 0.3-26ms + `updateSpacesAndScreen` 5-35ms

### Example Improvement:
```
Before: updateSpacesAndScreen = 170ms (64 API calls)
After:  Build map = 0.44ms + updateSpacesAndScreen = 5.34ms
Result: 5.78ms total (29x faster!) ✨
```

## Impact

- Eliminates the TODO comment about optimizing space info updates
- Reduces window processing latency by **2-29x** depending on scenario
- Makes rapid Alt+Tab switching significantly smoother on macOS 15.2+
- Zero overhead in Release builds (all measurements are `#if DEBUG`)

## Key Learnings

1. **Batch API calls when possible**: Making 10 API calls to build a cache is much faster than making 64 individual calls
2. **Profile before optimizing**: The detailed performance logging helped identify the exact bottleneck
3. **O(n) vs O(m) matters**: Switching from O(windows) to O(spaces) reduced calls by 6-30x
4. **WindowServer APIs are expensive**: Individual window queries take 2-2.5ms each; batching is critical

## Related Files

- `src/logic/Spaces.swift` - Added `buildWindowToSpacesMap()`
- `src/logic/Window.swift` - Modified `updateSpacesAndScreen()` to accept optional map
- `src/logic/Windows.swift` - Updated window loop to use cached mapping
- `src/api-wrappers/Logger.swift` - Performance logging infrastructure

</details>

cdbattags:
Possible for new review today? Would love people to have these fixes ASAP!

lwouis:
I'll review as soon as possible. I've started, but it's midnight here, and there is just so much code to review, and not the easiest code to read! I'll try to finish soon 🙇 
Thank you

src/logic/Spaces.swift
```diff
@@ -21,9 +24,41 @@ class Spaces {
...
+        return windowToSpaces
+    }
+
     static func refresh() {
-        refreshAllIdsAndIndexes()
-        updateCurrentSpace()
+        let now = DispatchTime.now()
+        let timeSinceLastRefresh = Double(now.uptimeNanoseconds - lastRefreshTime.uptimeNanoseconds) / 1_000_000_000
+
+        if timeSinceLastRefresh < refreshThrottle && !idsAndIndexes.isEmpty {
```
lwouis: I'm wondering about why we would like to throttle here.

Spaces.refresh is called in these scenarios:
* Once on launch, to know about Spaces
* Every time the user changes Space
* On every `Windows.updatesBeforeShowing()`

The first 2 should be pretty rare. And when they happen, i think we definitely want the latest Spaces data. As they just changed Space. The animations are not instant, so they should cover computation time on AltTab side as well.

The last case is the one that was the problem you identified I think. Regarding that specific use-case, Windows.updatesBeforeShowing itself is already throttled in this PR. Thus, I don't think there is a need to throttle Spaces.refresh.

Thus I suggest we remove that throttling. What do you think?

src/ui/App.swift
```diff
@@ -199,13 +199,34 @@ class App: AppCenterApplication {
...
         guard appIsBeingUsed else { return }
+
         if source == .refreshUiAfterExternalEvent {
+            // Throttle external events to prevent macOS 15.2 hang during rapid Tab presses
+            // caused by external accessibility events triggering hundreds of window list updates
+            // Note: refreshOpenUi is guaranteed to run on main thread, no queue needed
+            let now = DispatchTime.now()
+            if let lastRefresh = lastExternalEventRefreshTime {
+                let timeSinceLastRefresh = Double(now.uptimeNanoseconds - lastRefresh.uptimeNanoseconds) / 1_000_000_000
+                if timeSinceLastRefresh < externalEventRefreshThrottle {
```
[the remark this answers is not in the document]
cdbattags: Ok, understood! Will take a look today!

src/api-wrappers/Logger.swift
```diff
@@ -51,18 +92,160 @@ class Logger {
...
+        return result
+        #else
+        return block()
+        #endif
+    }
+
+    /// Measure execution time of an async code block (zero overhead in Release)
+    @inline(__always)
+    @discardableResult
+    static func measureAsync<T>(
```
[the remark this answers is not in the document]
lwouis: I'm still unsure about what to do with this instrumentation. At work, I've never seen permanent instrumentation. I've always seen applicative logs, performance KPI in tools like Datadog. I've never seen permanent instrumentation in the code. I've never worked on performance-critical systems though, so that may be common practice that I'm just not exposed to.

My first impression, in the context of this project is the following.

It creates a lot of noise for readers, just like any other logs. I've recently done a pass on all the logs, to tune them. Their level, remove some, add some. There are not that many in the app in the end. Currently, my intent with logs is to have them mark key moments in the app, to tell the story, when users share them here when they have an issue. Perf logs are a new thing in the codebase, raising many questions. This PR has them only during DEBUG. So only developer-facing. Now, as we've seen with the bug this PR is fixing, I wasn't having the perf issue on my machine. This makes me wonder if keeping the perf logs **for users**, would be a good idea. Adding them to the `verbose` level, as the PR does, but removing the check on DEBUG. This way I could ask users to share perf logs if other issues happen.

Another angle to think about this: how often do we face perf issues? Also, are we likely to face perf issues in the same place we instrument last time? Having such logs may be a nice way to do perf non-regression by keeping aware of execution time as the app continues to be developped. It's interesting. On the other hand, I think that once we have improved an area, it's unlikely that we need to revisit it.

Finally, this logs are a bit limited in what they can reveal. They are great to track blocking APIs. This is the major use-case, I think. They don't help much with AppKit rendering, which doesn't happen in the app's code, but is decided by it, indirectly. This doesn't help with repeating events, like this PR fixes, as the cumulative effect is not seen in logs. We would need some form of sampling to see that CPU is spent there. I'm not sure it's doable through Logs. We need to use Instrument ad-hoc for such cases, I think.

As you can see, it's a long answer. I've always done performance improvements with ad-hoc, one-off, profiling. The idea of logs here is new to me. So, I need time to evaluate how to go about it.

My idea is to first merge the fix, get people a better experience, then as a second step, review how to integrate perf logs into the app 👍

alt-tab-macos.xcodeproj/project.pbxproj
```diff
@@ -2041,6 +2045,14 @@
 			attributes = {
 				LastUpgradeCheck = 1620;
 				ORGANIZATIONNAME = lwouis;
+				TargetAttributes = {
+					BF0C873EC188A7EBBCF4310D = {
+						ProvisioningStyle = Automatic;
+					};
+					D04BA5E73DAF232ADE11C97A = {
+						ProvisioningStyle = Automatic;
+					};
```
cdbattags: to remove

alt-tab-macos.xcodeproj/project.pbxproj
```diff
@@ -3456,10 +3470,11 @@
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CLANG_WARN_UNREACHABLE_CODE = YES;
 				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
+				CODE_SIGN_IDENTITY = "-";
+				CODE_SIGN_STYLE = Automatic;
```
cdbattags: remove because we have the command line now for me

croteb:
Any update on this @cdbattags I've been waiting with bated breath...

cdbattags:
> Any update on this @cdbattags I've been waiting with bated breath...

Haven't had the time, sorry! My calendar is blocked off for this evening!

lwouis:
Closing this PR as v9 is out with loads of performance improvements.

Thank you so much @cdbattags for finding the key improvement with throttling refreshing the UI!

PR #5253, opened 2026-01-28
feat: add Windows-style persistent taskbar

### LINKED ISSUE #529: Always-on UI (i.e. bar or dock)

w93163red:
Add a persistent Windows-style taskbar at the bottom of the screen that displays all open windows with app icons and titles.
This is to partially resolve #529.
Features:
- Click to focus any window instantly
- Hover to preview window thumbnails (with 0.3s delay)
- Per-screen taskbar for multi-monitor setups
- Filter by current Space or show all windows
- Automatically adjusts maximized windows to leave room for taskbar
- Customizable settings: height, icon size, font size
- Show/hide minimized, hidden, and fullscreen windows

image

image

image


lwouis:
Hi,

First of all, I find this very cool, and I'm grateful for your sharing this great work!

We have to draw the line somewhere regarding what AltTab does. It won't do math calculation or web search, for example. People sometimes ask if we can integrate more window-management feature. I've added [this FAQ](https://alt-tab-macos.netlify.app/faq#can-we-add-window-management-features) to answer it.

In order to keep AltTab focused on its core mission, we have to renounce such features, however cool they are. I'm sorry about this. I hope you understand the decision 🙇 

Thank you

w93163red:
no worries. I will keep working on the feature on my own fork. Thanks!

PR #5257, opened 2026-01-29
Add per-shortcut option allowing tabbed windows to be shown and swapped to

josssch:
## Summary

This PR adds a per-shortcut option that enables the ability to swap to tabbed windows. This seems to work with thumbnail previews as well.

This was a feature I wanted for myself specifically to allow swapping between tabs like they're their own windows, but ONLY for the active app. Because of that, the default options reflect this preference of mine.

## Preview

image

## Existing Problems

- i18n
- I've only tested on macOS Tahoe 26.2

**Disclaimer:** I am not familiar with Swift or macOS development in general, but not too much needed to be changed to allow this, let me know if there are any blatant problems


lwouis:
This is subsumed by https://github.com/lwouis/alt-tab-macos/issues/5313.
Closing the PR for now, as we look for the right design 🙇 
Thank you

PR #5261, opened 2026-02-02
Improve focus for apps with associated windows

dsapper78:
## Summary
- Improves window focusing for apps like Telegram by trying the selected window id and a small set of associated window ids when using the private focus APIs.
- Keeps behavior conservative (max 3 ids) to reduce side effects.

## Context
- Addresses issue #5149 (Telegram focus). This appears related to the broader focus issues in #4192.

## Implementation
- Uses SLSCopyAssociatedWindows to collect candidate window ids.
- Applies _SLPSSetFrontProcessWithOptions + key-window event for each candidate id, then performs AX raise.

## Test plan
- Manual: reproduce on Telegram (and any other affected apps) on macOS 15.x by selecting a Telegram window in AltTab and verifying focus reliably switches.

## Bounty
- If this resolves the issue, the bounty says $200 cash. Happy to receive via BTC/Lightning as well if preferred.

### DIFF
--- a/src/api-wrappers/private-apis/SkyLight.framework.swift
+++ b/src/api-wrappers/private-apis/SkyLight.framework.swift
@@ -199,3 +199,9 @@ func _SLPSSetFrontProcessWithOptions(_ psn: UnsafeMutablePointer<ProcessSerialNu
 /// * macOS 10.12+
 @_silgen_name("SLPSPostEventRecordTo") @discardableResult
 func SLPSPostEventRecordTo(_ psn: UnsafeMutablePointer<ProcessSerialNumber>, _ bytes: UnsafeMutablePointer<UInt8>) -> CGError
+
+/// returns windows associated with the provided window
+/// used for apps where the visible "main" window is part of a window group (child windows)
+/// * macOS 10.12+
+@_silgen_name("SLSCopyAssociatedWindows")
+func SLSCopyAssociatedWindows(_ cid: CGSConnectionID, _ wid: CGWindowID) -> CFArray
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -216,8 +216,12 @@ class Window {
                 guard let self else { return }
                 var psn = ProcessSerialNumber()
                 GetProcessForPID(self.application.pid, &psn)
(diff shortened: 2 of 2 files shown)


energypantry:
Tested a local build of this PR.

One possible edge case with focusing via an associated/parent window id is that macOS may not reliably emit a focused-window AX event for the *intended* window. In that case AltTab's `lastFocusOrder` can stay stale and break cycling back-and-forth.

Small tweak suggestion: after `try? axUiElement.focusWindow()` (while you're already doing the 50ms `DispatchQueue.main.asyncAfter`), force-refresh focus ordering for the window we meant to focus and refresh the UI.

```diff
try? self.axUiElement!.focusWindow()
DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(50)) {
+    if let windows = Windows.updateLastFocusOrder(self) {
+        App.app.refreshOpenUi(windows, .refreshUiAfterExternalEvent)
+    }
     Windows.previewSelectedWindowIfNeeded()
}
```

This should be fairly low risk since it only runs after a focus attempt, and it ensures `lastFocusOrder` reflects the focus we just requested even if no AX event fires.

energypantry:
Hi @dsapper78,

Maintainer (@lwouis) replied on my follow-up PR #5273 that he'd prefer reviewing *one* PR for this focus issue.

I have a tiny add-on (7 LOC) that refreshes AltTab's `lastFocusOrder` + UI right after `axUiElement.focusWindow()` succeeds, because in some cases macOS doesn't emit a focused-window AX event for the target window (so AltTab's focus order stays stale and cycling back-and-forth breaks).

If you want, you can copy this snippet into `src/logic/Window.swift` right after `try? self.axUiElement!.focusWindow()` inside the `DispatchQueue.main.asyncAfter`:

```swift
// Some apps require focusing an associated/parent window id for the private focus APIs to take effect.
// In those cases, macOS may not reliably emit a focused-window AX event for the target window, which can
// leave AltTab's "lastFocusOrder" stale and break cycling back-and-forth. Ensure focus order reflects the
// window we intended to focus.
if let windows = Windows.updateLastFocusOrder(self) {
    App.app.refreshOpenUi(windows, .refreshUiAfterExternalEvent)
}
```

The same change is in #5273 as commit `09917b1` if you'd rather cherry-pick it.

Happy to adjust if it conflicts with your approach.

PR #5272, opened 2026-02-09
feat: detect and filter invisible windows (zero-opacity ghost windows)

brakedalen:
Some apps (e.g. Microsoft Outlook reminders) keep background windows with CGWindowAlpha=0 that appear in AltTab even though they are invisible to the user. This commit adds two complementary approaches to handle them:

## 1. Global preference: "Show invisible windows"
A new per-shortcut preference (Show / Hide / Show at the end) that filters windows with zero opacity across all apps — no configuration needed. Works just like the existing showMinimizedWindows, showHiddenWindows, and showFullscreenWindows preferences.

Files: Preferences.swift, ControlsTab.swift, Windows.swift (sort)

## 2. Enhanced blacklist: per-window title filtering The blacklist now supports an optional "Window title contains" column for case-insensitive substring matching, and a new hide option "When not visible on screen" that only hides matching windows when they are invisible (alpha=0). This allows precise control per app.

The existing "When no open window" option also hides transparent windows, since a zero-opacity window is effectively not visible.

Files: Preferences.swift (BlacklistEntry), MacroPreferences.swift,
       TableView.swift, BlacklistsTab.swift, Windows.swift

## Detection method
Uses CGWindowListCopyWindowInfo to check kCGWindowAlpha per window. Windows with alpha < 0.01 are marked as isTransparent=true and isCurrentlyOnScreen=false. This runs in updatesBeforeShowing() alongside existing space/tab detection.

Files: Window.swift (isTransparent, isCurrentlyOnScreen properties),
       Windows.swift (updateIsCurrentlyOnScreen method)

## Localization
English and Norwegian (nb) strings added for all new UI elements.

Default: "Show" (no behavior change for existing users).

### DIFF
--- a/resources/l10n/en.lproj/Localizable.strings
+++ b/resources/l10n/en.lproj/Localizable.strings
@@ -191,6 +191,7 @@
 "Show apps with no open window" = "Show apps with no open window";
 "Show at the end" = "Show at the end";
 "Show fullscreen windows" = "Show fullscreen windows";
+"Show invisible windows" = "Show invisible windows";
 "Show hidden windows" = "Show hidden windows";
 "Show in switcher" = "Show in switcher";
 "Show minimized windows" = "Show minimized windows";
@@ -246,6 +247,9 @@
 "Visible Spaces" = "Visible Spaces";
 "When fullscreen" = "When fullscreen";
 "When no open window" = "When no open window";
+"When not visible on screen" = "When not visible on screen";
+"Window title contains" = "Window title contains";
+"Any" = "Any";
 "While open, press:" = "While open, press:";
 "Window app icon size:" = "Window app icon size:";
 "Window is fullscreen" = "Window is fullscreen";
--- a/resources/l10n/nb.lproj/Localizable.strings
+++ b/resources/l10n/nb.lproj/Localizable.strings
@@ -192,6 +192,7 @@
 "Show apps with no open window" = "Vis apper uten åpne vinduer";
 "Show at the end" = "Vis på slutten";
 "Show fullscreen windows" = "Vis fullskjermvinduer";
+"Show invisible windows" = "Vis usynlige vinduer";
 "Show hidden windows" = "Vis skjulte vinduer";
 "Show in switcher" = "Vis i switcher";
 "Show minimized windows" = "Vis minimerte vinduer";
@@ -250,6 +251,9 @@
 "Visible Spaces" = "Synlige områder";
 "When fullscreen" = "Når i fullskjerm";
 "When no open window" = "Når ingen åpne vinduer";
+"When not visible on screen" = "Når ikke synlig på skjermen";
+"Window title contains" = "Vindutittel inneholder";
+"Any" = "Alle";
 "While open, press:" = "Når åpen, trykk:";
 "Window app icon size:" = "Størrelse på vinduappikon:";
 "Window is fullscreen" = "Vinduet er i fullskjerm";
--- a/src/logic/MacroPreferences.swift
+++ b/src/logic/MacroPreferences.swift
@@ -519,12 +519,14 @@ enum BlacklistHidePreference: String/* required for jsonEncode */, CaseIterable,
     case none = "0"
     case always = "1"
     case whenNoOpenWindow = "2"
+    case whenNotOnScreen = "3"
 
     var localizedString: LocalizedString {
         switch self {
             case .none: return ""
             case .always: return NSLocalizedString("Always", comment: "")
             case .whenNoOpenWindow: return NSLocalizedString("When no open window", comment: "")
+            case .whenNotOnScreen: return NSLocalizedString("When not visible on screen", comment: "")
         }
     }
 }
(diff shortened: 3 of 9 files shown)


brakedalen:
What do you think about this @lwouis ?

lwouis:
Hi,

> Uses CGWindowListCopyWindowInfo to check kCGWindowAlpha per window

Does this work for every case? My previous implementation was to check pixel-by-pixel to ensure every pixel was transparent. This, in order to avoid flagging partially transparent window. Partial transparency happens more frequently since v8, since the ScreenCapture Kit API we use, will return partial transparency if there is for example a window with a modal sheet attached to it (e.g. open image then try to print it => AltTab will show this window + sheet combo)

In general, these invisible windows are a puzzle I haven't been able to solve. The reason is that we don't know why the windows are invisible. It is pure semantic from the app makers. Some examples:
* KeePassXC: hidding pixels from apps like AltTab so we don't steal passwords
* Outlook: keeps an invisible window open to quickly show Reminders (#5193). This is considered poor implementation, yet it's common in big apps, especially when cross-platform.
* Apps opening a modal sheet: partial transparency because the geometry is no longer a simple rectangle

You can see the challenge: we can't really decide to hide all those, or show all those. It will always frustrate some of the users. Currently, we show them, to be more conservative. So KeePassXC users can at least switch to their window. It creates frustration for Outlook users, but really this energy should go towards Microsoft to fix their app.

Having settings so users can tweak things is interesting. I think the global setting creates too much noise for most users who won't use it, so I would say that's not possible. The Blacklist is already a place for power-users, so I think it's fine there. A know section for invisible windows seem complex. Matching window title, I'm not sure either. Like Outlook + "Reminder" will hide it when it's actually useful. So maybe just the "hide invisible windows" checkbox would be best. 

I need to think more about this

Thank you

lwouis:
> Uses CGWindowListCopyWindowInfo to check kCGWindowAlpha per window
>
>Does this work for every case?

I can confirm that this API doesn't work. It's not telling us if a window is transparent or not. It's unreliable.

jgilchrist:
> Matching window title, I'm not sure either. Like Outlook + "Reminder" will hide it when it's actually useful. So maybe just the "hide invisible windows" checkbox would be best.

Just one data point from an outlook user, but I'd personally not use AltTab to tab to this reminder window, so the ability to hide based on the title of the window would be a big improvement for me and sidesteps the complexity of determining fully transparent windows.

brakedalen:
I can confirm it works for reminder in outlook. That is what I made it for, as the hidden reminder was driving me crazyy. Both the global setting and the the option in blacklist to hide hidden windows.

I have had the app running for two machines since I created it and can confirm there are no bugs related to the change.

I think it's also a nice addition to be able to hide based on name as well😁


- Even
________________________________
Fra: lwouis ***@***.***>
Sendt: mandag, februar 16, 2026 10:30
Til: lwouis/alt-tab-macos ***@***.***>
Kopi: brakedalen ***@***.***>; Author ***@***.***>
Emne: Re: [lwouis/alt-tab-macos] feat: detect and filter invisible windows (zero-opacity ghost windows) (PR #5272)

[https://avatars.githubusercontent.com/u/106195?s=20&v=4]lwouis left a comment (lwouis/alt-tab-macos#5272)<https://github.com/lwouis/alt-tab-macos/pull/5272#issuecomment-3907369063>

Hi,

Uses CGWindowListCopyWindowInfo to check kCGWindowAlpha per window

Does this work for every case? My previous implementation was to check pixel-by-pixel to ensure every pixel was transparent. This, in order to avoid flagging partially transparent window. Partial transparency happens more frequently since v8, since the ScreenCapture Kit API we use, will return partial transparency if there is for example a window with a modal sheet attached to it (e.g. open image then try to print it => AltTab will show this window + sheet combo)

In general, these invisible windows are a puzzle I haven't been able to solve. The reason is that we don't know why the windows are invisible. It is pure semantic from the app makers. Some examples:

  *   KeePassXC: hidding pixels from apps like AltTab so we don't steal passwords
  *   Outlook: keeps an invisible window open to quickly show Reminders (#5193<https://github.com/lwouis/alt-tab-macos/issues/5193>). This is considered poor implementation, yet it's common in big apps, especially when cross-platform.
  *   Apps opening a modal sheet: partial transparency because the geometry is no longer a simple rectangle

You can see the challenge: we can't really decide to hide all those, or show all those. It will always frustrate some of the users. Currently, we show them, to be more conservative. So KeePassXC users can at least switch to their window. It creates frustration for Outlook users, but really this energy should go towards Microsoft to fix their app.

Having settings so users can tweak things is interesting. I think the global setting creates too much noise for most users who won't use it, so I would say that's not possible. The Blacklist is already a place for power-users, so I think it's fine there. A know section for invisible windows seem complex. Matching window title, I'm not sure either. Like Outlook + "Reminder" will hide it when it's actually useful. So maybe just the "hide invisible windows" checkbox would be best.

I need to think more about this

Thank you

lwouis:
I've add black title filtering. I don't think I'll add transparency filtering for now, because of the performance penalty. Closing this PR for now. Thank you for the help! 👍

PR #5273, opened 2026-02-09
Fix focus order refresh after AX focus of associated window

energypantry:
Fixes a regression noted in #5261: when AltTab focuses an *associated window* via AX, `lastFocusOrder` is not updated (e.g. TextEdit + Save As modal), so subsequent ordering/selection can be wrong.

This PR updates focus bookkeeping immediately after `axUiElement.focusWindow()`:
- refreshes `Windows.lastFocusOrder`
- refreshes the open UI if it's currently showing a window from the focused app

Refs:
- Issue: #5149 (Telegram focus)
- Related: #5261 (associated window focusing)

If you prefer, this can be squashed / applied directly on top of #5261 instead of merging separately.

### DIFF
--- a/src/api-wrappers/private-apis/SkyLight.framework.swift
+++ b/src/api-wrappers/private-apis/SkyLight.framework.swift
@@ -199,9 +199,3 @@ func _SLPSSetFrontProcessWithOptions(_ psn: UnsafeMutablePointer<ProcessSerialNu
 /// * macOS 10.12+
 @_silgen_name("SLPSPostEventRecordTo") @discardableResult
 func SLPSPostEventRecordTo(_ psn: UnsafeMutablePointer<ProcessSerialNumber>, _ bytes: UnsafeMutablePointer<UInt8>) -> CGError
-
-/// returns windows associated with the provided window
-/// used for apps where the visible "main" window is part of a window group (child windows)
-/// * macOS 10.12+
-@_silgen_name("SLSCopyAssociatedWindows")
-func SLSCopyAssociatedWindows(_ cid: CGSConnectionID, _ wid: CGWindowID) -> CFArray
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -216,20 +216,21 @@ class Window {
                 guard let self else { return }
                 var psn = ProcessSerialNumber()
                 GetProcessForPID(self.application.pid, &psn)
-                // Some apps (e.g. Telegram/Teams/ChatGPT) may show a window in the list but require focusing an
-                // associated/parent window id for the private focus APIs to take effect.
-                for wid in self.focusCandidateWindowIds() {
-                    _SLPSSetFrontProcessWithOptions(&psn, wid, SLPSMode.userGenerated.rawValue)
-                    self.makeKeyWindow(&psn, wid)
-                }
+                _SLPSSetFrontProcessWithOptions(&psn, self.cgWindowId!, SLPSMode.userGenerated.rawValue)
+                self.makeKeyWindow(&psn)
                 try? self.axUiElement!.focusWindow()
                 DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(50)) {
-                    // Some apps require focusing an associated/parent window id for the private focus APIs to take effect.
-                    // In those cases, macOS may not reliably emit a focused-window AX event for the target window, which can
-                    // leave AltTab's "lastFocusOrder" stale and break cycling back-and-forth. Ensure focus order reflects the
-                    // window we intended to focus.
-                    if let windows = Windows.updateLastFocusOrder(self) {
-                        App.app.refreshOpenUi(windows, .refreshUiAfterExternalEvent)
+                    // Best-effort: some focus paths don't reliably emit kAXFocusedWindowChangedNotification. Avoid
+                    // "forcing" focus state in AltTab; instead, query the actual focused window and update ordering
+                    // only if it matches a known Window.
+                    if let appAxUiElement = self.application.axUiElement,
+                       let focusedAxWindow = try? appAxUiElement.attributes([kAXFocusedWindowAttribute]).focusedWindow,
+                       let focusedAxWid = try? focusedAxWindow.cgWindowId(),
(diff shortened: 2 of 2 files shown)


energypantry:
Quick heads-up on this follow-up PR (#5273): it only targets the `lastFocusOrder` regression you noticed with focusing associated windows.

It updates the focus order + refreshes the open UI right after the AX focus succeeds, so you can AltTab back-and-forth between the original window and its attached/modal window again.

When you have a moment, could you take a look and tell me if you'd prefer I rebase/cherry-pick onto #5261 instead?

Re bounty rails: I can take PayPal or BTC/Lightning (preferred).

lwouis:
Hi,
Yes, i'd prefer looking at 1 PR to fix the focus issue 🙇 
Thank you

energypantry:
Thanks! Understood on preferring a single PR.

My follow-up in #5273 is just a tiny add-on (commit `09917b1`) that refreshes `lastFocusOrder` + UI right after `axUiElement.focusWindow()` so cycling between a window and its associated/modal window works again even if macOS doesn't emit a focused-window AX event.

I'll ask @dsapper78 to cherry-pick/apply that small change into #5261 so you only need to review one PR. Once it's in #5261, I'll close #5273 to avoid duplication.

lwouis:
Hi,

Sorry, I was confused. I didn't notice the 2 PRs were from 2 different authors. I was checking through email, and got confused. So you're building on the work from @dsapper78. I understand now. Since this PR has the 2 commits, I can review it.

Regarding the bounty, I have no idea how collaboration is viewed. You may want to talk to @kodjima33 since he's the one promising the bounty. F

Furthermore, we need them to QA since I can't QA on my machine (it works correctly already, on my machine). How do you both test this change? Are you reproducing with Telegram, or with another setup? Or maybe you did a blind/theoretical fix?

Thank you

energypantry:
Thanks for the clarification @lwouis.

Re bounty/collaboration: understood. @kodjima33 is the one who offered the $200 on #5149, so I’ll coordinate with them on whether the bounty applies to this PR and how they want to handle credit/splitting.

Re QA/testing:
- I’m not currently able to reproduce the *Telegram* issue on my machine (same situation as you: it already works fine here).
- - This PR is a small, targeted fix for a related class of issues: when we focus an associated/modal window via AX, macOS may not emit a focused-window AX event, which leaves `Windows.lastFocusOrder` and the Open UI stale. The result is that AltTab cycling can get stuck / not behave as expected after focusing the associated window.
- - The fix is “blind” in the sense that it’s not validated against the Telegram repro specifically, but it’s grounded in the focus path and only runs after `axUiElement.focusWindow()` succeeds.
QA request (for @kodjima33, and anyone who can repro #5149):
1) Please test the original Telegram repro from #5149 with this build/patch.
2) Also test a deterministic “associated window” scenario (sheet/modal), e.g. TextEdit -> File -> Save… (sheet) or any app that opens an attached sheet:
   - AltTab to the app’s main window
   -    - Trigger the sheet
   -    - AltTab to the sheet / associated window
   -    - Confirm you can AltTab back and forth between main window <-> sheet without getting stuck
I attached a Debug build (`AltTab-Debug-5273.zip`) for convenience (built with code signing disabled). If Gatekeeper blocks it, you may need right-click -> Open.

If you prefer I add any extra logging to help QA, tell me what signal you’d like (focused window id? lastFocusOrder before/after?) and I’ll push a follow-up.
[AltTab-Debug-5273.zip](https://github.com/user-attachments/files/25186452/AltTab-Debug-5273.zip)

energypantry:
@kodjima33 quick check when you have a moment:

Could you please QA this fix using the attached debug build from my previous comment and confirm whether it resolves the focus-order issue on your setup?

Steps I tested locally:
1) Repeatedly focus a window via AltTab and verify the list ordering updates immediately after focusing.
2) Specifically reproduce the Telegram + browser focus switching flow (the scenario from the bounty thread) and see whether the next AltTab selection is correct.

If you can reply with a short confirmation (works / still broken and how) it should unblock @lwouis to merge.

Thanks.

lwouis:
Hi,

I got time to review your change. I think there are a few issues with the overall proposal of your commit plus the other commit that was before.

I think, first of all, in the original commit, we have this idea of focusing every associated window. I'm not sure that sound. I'm not clear what constitutes an associated window. If it's the parent, I guess it makes sense. But if it is dialogues or other model or sheets or pseudo windows, then it becomes confusing. If we think of an app like Photoshop, for example, I'm not sure how Photoshop would react in that situation. Maybe there will be dozens of associated windows and we'll get all sorts of weird behavior for users and performance issues. So first of all, that's my first concern, that this kind of indiscriminate blast radius, just focusing everything, I think it's too rough. I think we need to perhaps just identify the parent and focus that, and maybe only do that in the case where we cannot focus the window directly. See, I feel we find, we miss refinement here.

And then second, your second commit that's supposed to fix issues, we are kind of just brute forcing, setting the focus on the window, even if we may not have done it for the actual user. We are saying inside AltTab to force the state of that window to be focused. So we may be desyncing with reality. So with all these observations, I think this is not mature to merge and we need to go back to the drawing board and really analyze a bit the options we have.

Thank you

energypantry:
Thanks for the detailed review @lwouis. I agree with both concerns.

I pushed a follow-up commit to address them: `ab25175` (current PR head).

What changed:
1) Removed the "focus every associated window" idea entirely.
   - No `SLSCopyAssociatedWindows` / no looping over extra window IDs.
   -    - We only try to focus the target window ID.
2) Avoid desyncing AltTab state with reality.
   - After `axUiElement.focusWindow()` succeeds, we *query the actual focused window* via AX (`kAXFocusedWindowAttribute`).
   -    - We update `lastFocusOrder` (and refresh the Open UI) only if that focused AX window matches a known `Window`.
   -    - If macOS focus didn't change to something we can identify, we do nothing (no forced focus ordering).
QA is still needed from someone who can repro the Telegram case (#5149). I’m still waiting on @kodjima33 for confirmation.

If this direction looks acceptable, I can squash the PR down so you only review the refined behavior. Otherwise, happy to iterate on a more principled approach.

PR #5318, opened 2026-02-20
feat(cli): expose screen, space, and app metadata in --detailed-list

### LINKED ISSUE #5319: CLI: --detailed-list missing screen, space, and blacklist metadata
I already run a forked version of the AltTab app to use this AltTab fix, merging this fix would allow me to use the main release again

---

## Context

Following up on #5241 where [the CLI API was suggested](https://github.com/lwouis/alt-tab-macos/issues/5241#issuecomment-2759375912) as a solution for programmatic window cycling. The CLI works well, but `--detailed-list` is missing several fields that the app tracks internally, forcing external tools to use workarounds.

## What's missing

**Per-window:**
- `screenId` — which physical monitor the window belongs to. Without this, external scripts have to reverse-engineer screen membership by comparing each window's `position` against monitor bounds (via a separate Swift helper binary). This is unreliable across Spaces.

**Top-level metadata (alongside `windows`):**
- `screens` — list of screens with UUIDs and frames
- `currentSpaceIndex` / `visibleSpaceIndexes` — which spaces are active
- `screenSpacesMap` — which spaces belong to which screen
- `blacklist` — the user's configured blacklist entries

## Why it matters

The app already computes all of this internally (`Window.screenId`, `Spaces.currentSpaceIndex`, `Spaces.visibleSpaces`, `Spaces.screenSpacesMap`, `Preferences.blacklist`). It's just not serialized in the CLI response.

Without these, CLI consumers must:
1. Compile and run a separate Swift binary to get monitor bounds + mouse position
2. Hardcode their own blacklist (duplicating the AltTab UI config)
3. Guess which screen a window is on via position-matching with tolerance — which breaks when windows are on different Spaces of the same monitor (i.e. it's brittle at the edges of monitor/screen coordinates)

## PR

#5318

Elijas:
Closes #5319

---

## Summary

Enrich the `--detailed-list` CLI response with fields needed for programmatic window management:

**Per-window additions:**
- `screenId` — which physical monitor the window belongs to
- `appPid` — application process ID
- `dockLabel` — dock item label

**Top-level metadata (new fields alongside `windows`):**
- `screens` — array of screen UUIDs with Quartz-coordinate frames
- `currentSpaceIndex` — which space is currently active
- `visibleSpaceIndexes` — one active space per monitor
- `screenSpacesMap` — which spaces belong to which screen
- `frontmostAppPid` — currently focused application PID
- `mouseScreenId` — UUID of the screen the mouse cursor is currently on (uses existing `NSScreen.withMouse()` logic)
- `blacklist` — the user's configured blacklist entries (bundleId + hide/ignore settings)

## Motivation

External tools that cycle windows by monitor (e.g., skhd scripts) currently have to reverse-engineer screen membership from `position` coordinates + monitor bounds — which is unreliable across Spaces. Exposing `screenId` and the screen/space topology from a single `--detailed-list` call eliminates the need for separate helper binaries and position-based heuristics.

`mouseScreenId` enables tools to match AltTab's own "Screen including mouse" behavior — cycling windows on the hovered monitor rather than the focused window's monitor — without requiring external mouse-tracking utilities.

**My use case:** I use the mouse back/forward buttons (remapped to keyboard shortcuts via mouse software, picked up by skhd) to cycle through windows on the screen the mouse is hovering over. Each press advances to the next/previous window in creation order, wrapping around. A jq script filters `--detailed-list` by `mouseScreenId` to stay on one monitor. This needs `screenId`, `mouseScreenId`, and fresh space data to work reliably — without them, the cycle either targets the wrong screen or silently drops visible windows.

The latest commit also adds a `Spaces.refresh()` + per-window `updateSpacesAndScreen()` call before serializing `--detailed-list`, matching what the keyboard-triggered UI already does. Without this, the CLI could return stale `screenId` (null) and stale `visibleSpaceIndexes` for recently created or moved windows.

## Test plan

Code from this PR already works 100% on my 'dev' branch in my fork.

- [x] Build and run AltTab
- [x] Run `AltTab --detailed-list | jq .` and verify new fields are present
- [x] Verify `screenId` matches expected monitor for windows on different screens (5 screens detected, UUIDs map to distinct monitors)
- [x] Verify `visibleSpaceIndexes` updates when switching Spaces (currently reports [1,2,4,5,7] — one per monitor)
- [x] Verify `blacklist` reflects current AltTab preferences (13 entries including user-added wispr-flow)
- [x] Verify `mouseScreenId` returns the correct screen UUID when mouse is on different monitors
- [x] Verify `mouseScreenId` differs from focused window's `screenId` when mouse and focus are on different screens
- [x] Verify space/screen data is fresh after creating a new window (no stale null `screenId`)

### DIFF
--- a/src/logic/events/CliEvents.swift
+++ b/src/logic/events/CliEvents.swift
@@ -51,6 +51,11 @@ class CliServer {
             )
         }
         if rawValue == "--detailed-list" {
+            // refresh space/screen assignments so CLI returns fresh data
+            Spaces.refresh()
+            for window in Windows.list {
+                window.updateSpacesAndScreen()
+            }
             let windows = Windows.list
                 .filter { !$0.isWindowlessApp }
                 .map {


Elijas:
@lwouis Any feedback regarding this PR? 

The CLI API is genuinely useful, but it is currently stale - this PR fully fixes it

Without this PR the AltTab CLI access is severely degraded compared to the keyboard-activated switcher/UI

lwouis:
Hi,

I've reviewed the pull requests, and I have conflicting thoughts.

The original idea with the CLI was to provide a way to act on AltTab using scripts through the command line. For example, bring the AltTab UI like you would with the shortcut, but do that with the CLI. This allows any kind of input to be piped to the CLI, which then triggers AltTab to show. In that sense, it's a universal interface to call AltTab.

At that time, I also added the ability to list windows. I don’t clearly remember why, and that may have been a mistake.

If we move in the direction you propose, based on your use case, we would essentially use AltTab as a database of system state. We would query it to retrieve window state and bypass most of AltTab’s functionality: filtering, ordering, and other criteria. In practice, this would bypass nearly everything in AltTab except the state management.

You would then apply your own logic on your side and call AltTab only to perform actions like focusing a window, fullscreening it, closing it, and similar operations. You could even bypass AltTab entirely for that step by calling AppleScript or other APIs directly.

In that model, the CLI becomes primarily a way to exploit AltTab’s internal state management as a database.

I’m conflicted about this direction. If the goal is programmatic window management, other tools already exist that are designed for it, such as Yabai or Aerospace. They do it better and more thoroughly because that is their core purpose.

There is also a structural issue with how the CLI currently works. You first launch AltTab, and then a second CLI process acts as a client communicating with the running instance, which behaves like a server. If the only goal is retrieving window state, this architecture is not ideal. In that case, you would probably want a dedicated server that clients communicate with directly.

In summary, I’m not comfortable opening the doors and having AltTab share everything it contains while allowing people to bypass the utility it provides.

I would be much more comfortable with a model where AltTab continues to handle the logic, and the CLI simply triggers actions. For example, you could configure AltTab in the settings to only display windows from the current screen. Then, through the CLI, you could request the next window on that screen, then the next one again, and so on.

In that approach, you would not need to retrieve internal state from AltTab. You would not need to know about blacklists, window types, the current screen, or window screen IDs. AltTab would manage those details internally based on its configuration. The CLI would simply trigger actions such as selecting the next window or showing the UI, after which you could interact manually or script the focus behavior.

This is the direction I would prefer. I want people to be able to programmatically invoke AltTab—essentially an alternative trigger—but not to expose the internal datastore and allow external systems to build their own logic on top of it. That doesn’t align with the goal of AltTab and would likely be poor design.

A related idea already exists. Someone recently extracted AltTab’s window state management into a library and built an app on top of it. For your use case, that may be the better direction: extracting state management into a Swift library or an XPC server so that others can build tools on top of it.

Using a full AltTab instance—with many megabytes of assets such as images used to illustrate settings—just to run it as a headless server doesn’t seem like a good fit.

I hope you can understand that perspective.

For now, I’ll close this PR and not move in that direction. I may revisit it later. If enough people request this capability, I may reconsider. That could mean adopting something like your PR, or extracting the state management layer into a separate library or service that is independent from the macOS desktop app with its UI.

Thank you

Elijas:
> Hi,
> 
> I've reviewed the pull requests, and I have conflicting thoughts.
> 
> The original idea with the CLI was to provide a way to act on AltTab using scripts through the command line. For example, bring the AltTab UI like you would with the shortcut, but do that with the CLI. This allows any kind of input to be piped to the CLI, which then triggers AltTab to show. In that sense, it's a universal interface to call AltTab.
> 
> At that time, I also added the ability to list windows. I don’t clearly remember why, and that may have been a mistake.
> 
> If we move in the direction you propose, based on your use case, we would essentially use AltTab as a database of system state. We would query it to retrieve window state and bypass most of AltTab’s functionality: filtering, ordering, and other criteria. In practice, this would bypass nearly everything in AltTab except the state management.
> 
> You would then apply your own logic on your side and call AltTab only to perform actions like focusing a window, fullscreening it, closing it, and similar operations. You could even bypass AltTab entirely for that step by calling AppleScript or other APIs directly.
> 
> In that model, the CLI becomes primarily a way to exploit AltTab’s internal state management as a database.
> 
> I’m conflicted about this direction. If the goal is programmatic window management, other tools already exist that are designed for it, such as Yabai or Aerospace. They do it better and more thoroughly because that is their core purpose.
> 
> There is also a structural issue with how the CLI currently works. You first launch AltTab, and then a second CLI process acts as a client communicating with the running instance, which behaves like a server. If the only goal is retrieving window state, this architecture is not ideal. In that case, you would probably want a dedicated server that clients communicate with directly.
> 
> In summary, I’m not comfortable opening the doors and having AltTab share everything it contains while allowing people to bypass the utility it provides.
> 
> I would be much more comfortable with a model where AltTab continues to handle the logic, and the CLI simply triggers actions. For example, you could configure AltTab in the settings to only display windows from the current screen. Then, through the CLI, you could request the next window on that screen, then the next one again, and so on.
> 
> In that approach, you would not need to retrieve internal state from AltTab. You would not need to know about blacklists, window types, the current screen, or window screen IDs. AltTab would manage those details internally based on its configuration. The CLI would simply trigger actions such as selecting the next window or showing the UI, after which you could interact manually or script the focus behavior.
> 
> This is the direction I would prefer. I want people to be able to programmatically invoke AltTab—essentially an alternative trigger—but not to expose the internal datastore and allow external systems to build their own logic on top of it. That doesn’t align with the goal of AltTab and would likely be poor design.
> 
> A related idea already exists. Someone recently extracted AltTab’s window state management into a library and built an app on top of it. For your use case, that may be the better direction: extracting state management into a Swift library or an XPC server so that others can build tools on top of it.
> 
> Using a full AltTab instance—with many megabytes of assets such as images used to illustrate settings—just to run it as a headless server doesn’t seem like a good fit.
> 
> I hope you can understand that perspective.
> 
> For now, I’ll close this PR and not move in that direction. I may revisit it later. If enough people request this capability, I may reconsider. That could mean adopting something like your PR, or extracting the state management layer into a separate library or service that is independent from the macOS desktop app with its UI.
> 
> Thank you

Thank you for such a detailed and thoughtful response, I really appreciate it

Makes sense as a design boundary. I'll maintain a fork for my use case to avoid running two simultaneous applications tracking/managing the same macOS window state.

PR #5366, opened 2026-03-02
feat: add configurable macos-like app switcher mode

salman-ar-sar:
This PR:

- replaces the per-shortcut apps-only toggle with a per-shortcut **max windows per app** setting (`No limit`, `1`..`5`)
- only shows the new control when **Show windows from applications** is set to **All apps**
- updates switcher filtering so each app can be capped to the configured number of windows, while keeping main-window-first behavior

### Reasoning

Users want an option closer to the native macOS app switcher behavior, while still keeping AltTab's richer controls. This feature provides that middle ground:

- setting the max to `1` gives native-like app switching (one entry per app)
- values above `1` keep multiple windows per app, but in a bounded and predictable way
- existing options like spaces/screens filters still apply, so users can customize scope beyond default macOS behavior

Validation:

- `xcodebuild -workspace alt-tab-macos.xcworkspace -scheme Debug -configuration Debug -derivedDataPath ~/git/alt-tab-macos/DerivedData CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- `xcodebuild -workspace alt-tab-macos.xcworkspace -scheme Test -configuration Debug -derivedDataPath ~/git/alt-tab-macos/DerivedData CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" test`

### DIFF
--- a/src/logic/MacroPreferences.swift
+++ b/src/logic/MacroPreferences.swift
@@ -1,3 +1,5 @@
+import Cocoa
+
 enum MenubarIconPreference: CaseIterable, MacroPreference {
     case outlined
     case filled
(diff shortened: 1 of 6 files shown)


lwouis:
If we invest in this direction, it will be through #337

Issue #5346, opened 2026-02-24
Apps not updating position in Alt Tab list


AdriandeLima:
**Describe the bug**

I've run into this issue a fair amount when quickly swapping apps, where pressing alt tab to swap apps doesn't actually swap the app. The problem ends up being  that the app hasn't update it's location in the list, and so you end up swapping back to the app you're trying to swap from.

I've noticed this happening on occasion with Vs Code - I try to swap quickly from app position 1 - 2 but for some reason Vs Code ends up being in app position 2 instead of 1, and I end up swapping to what was already open.

The other place it happens a lot is Final Cut. Final cut often beach balls for me, or goes into not responding. Whenever this happens, the alt tab position of the app gets stuck, and pressing alt tab once, will only select Final Cut, which was already open. I also noticed that for some reason it crashed alt tab, and when I opened alt tab again, Final Cut wasn't showing in the menu at all. This means you need to press tab 2x to actually switch programmes.

I wonder if this has something to do with programmes loading (in the case of vs code), or not responding?

**Screenshots / video**

https://youtu.be/LvQW06YEo_E

**Steps to reproduce the bug**

Alt Tab a not responding programme, it won't update it's position in the list.

**Your environment**

* AltTab version: 10.2.0
* macOS version: 15.7.4

AdriandeLima:
Turns out it might not be a problem with just Final Cut, it happened with Brave web browser today, and nothing was not responding.

![Image](https://github.com/user-attachments/assets/60c77fda-e429-4424-a3e9-1b8d0685441a)

Note, in this gif I'm pressing alt+tab once, and it's not switching to the next app.

yassine-safraoui:
Hi, I just reported the same issue by mistake, check it out for more details, overall it's the same bug.

zhizhongpu:
I have the same issue in the recent version. This happens for Edge and Microsoft suite

alexbluo:
same issue, think it happens for every application. somewhat reproducible by spamming alt tab quickly

lwouis:
Could you please try out [this local build](https://github.com/user-attachments/files/25727279/AltTab.app.zip), and let me know if it fixes the issue?

Thank you

kidhanis:
It does fix the issue!

On v10.3.0, I could consistently reproduce the issue with the following steps:
1. Reset AltTab settings
2. Change AltTab style to Titles
3. Run the following AppleScript on the Script Editor:
	```applescript
[...]

PR #5367, opened 2026-03-02
Fix/window focus throttle

jchayat:
Ever since upgrading to the 10.x version (from the latest 8.x version), window ordering seems to have been broken. I think it's the same issue as https://github.com/lwouis/alt-tab-macos/issues/5346

I tried to vibe code this with Claude Code. There's a detailed explanation in the commit messages. I haven't been able to test this on my machine, due to being on an older macOS/XCode version. Please see if this helps with the issue. Even if the PR is not accepted, the information in the commit message might prove valuable.


lwouis:
Thank you. I'm working on a fix for this 👍

Issue #5336, opened 2026-02-23
System-wide UI freeze (5-10 seconds) when pressing modifier key combinations


ihor-sviziev:
**AltTab version:** 10.2.0
**macOS:** 26.3 (Tahoe, build 25D125)
**Hardware:** MacBook Pro M1 Pro, 32 GB
**Display setup:** Internal display + external monitor

## Description

After running AltTab for several days, pressing any modifier key combination causes the entire macOS UI to freeze for 1-10 seconds. This affects **any** modifier key combo — not just AltTab's shortcut:

- Shift+A (typing a capital letter)
- Cmd+C, Cmd+V, Cmd+anything
- Ctrl+anything
- Option+anything

During the freeze:
- The mouse cursor **continues to move** normally
- But the entire system becomes fully stuck — unable to move windows, change focus to another window, switch between browser tabs
- Even hover effects stop working (e.g. close/minimize/maximize buttons don't highlight on hover)
- Freezes last between 1 and 10 seconds

**Important:** If you don't touch the keyboard and only use the mouse, everything works perfectly fine. The freeze is triggered exclusively by modifier key presses.

The issue reproduces on both internal and external keyboards.

During the freezes, there is no high CPU or memory usage visible in htop — everything appears normal, yet the UI is completely unresponsive.

The issue usually starts reproducing after extended work sessions in an IDE (WebStorm), which creates many sub-windows and panels that AltTab likely needs to enumerate.

## Frequency

The issue has been escalating over time:
- Initially reproduced very rarely — roughly once a quarter
- Then increased to about once a month
- During the last month alone, it happened 3 times
- Every time, only a full reboot resolved it — until this time, when quitting AltTab was identified as the fix

## Reproduction

- Not immediately reproducible after launch
- Develops gradually over days of uptime
- Usually appears after extended IDE (WebStorm) usage with dual monitors
- Affects all applications, not just AltTab's own shortcut
- Frequency has been increasing over time (see above)

## Diagnosis

- **Quitting AltTab immediately and completely resolves the issue** — no reboot needed
- No thermal throttling, no excessive CPU/memory by AltTab itself
- Slow Keys and Sticky Keys are disabled
- No other keyboard-hooking apps installed (no Karabiner, no skhd, etc.)
- WindowServer was at ~27% CPU during the freezes
- ~620 processes running at the time
- No high CPU or memory usage during freezes (confirmed via htop)

## Diagnostic methodology

[...]

PR #5370, opened 2026-03-03
fix: run TrackpadEvents on main thread for toNSEvent() correctness

jhult:
`toNSEvent()` must be called on the main thread. Unlike `CursorEvents` (which was updated in #5336 to use `cgEvent.location` directly), `TrackpadEvents` still calls `toNSEvent()` and must synchronously decide whether to absorb events — so it cannot use a background thread or switch to `.listenOnly`.

This moves the trackpad run loop source to `CFRunLoopGetMain()` to fix the incorrect background-thread usage.

### DIFF
--- a/src/logic/events/TrackpadEvents.swift
+++ b/src/logic/events/TrackpadEvents.swift
@@ -39,7 +39,8 @@ class TrackpadEvents {
             userInfo: nil)
         if let eventTap {
             let runLoopSource = CFMachPortCreateRunLoopSource(nil, eventTap, 0)
-            CFRunLoopAddSource(BackgroundWork.keyboardAndMouseAndTrackpadEventsThread.runLoop, runLoopSource, .commonModes)
+            // we run on main-thread since toNSEvent() must be called on main-thread, and we need to synchronously decide if we absorb the event
+            CFRunLoopAddSource(CFRunLoopGetMain(), runLoopSource, .commonModes)
         } else {
             App.restart()
         }


lwouis:
Hi,

I'm not sure I understand this change. What issue are you trying to fix?

Currently, TrackpadEvents requires NSEvent to get the multi-finger touches. We track events from a background thread, and we use `toNSEvent` to construct the `NSEvent` from the `CGEvent`. This is correct. It may not be best, but it's correct.

What are the benefits to having every event entirely processed on the main thread?

jhult:
This was a suggestion from Claude when I was asking it if the fix (commit https://github.com/lwouis/alt-tab-macos/commit/3fb66d412a99dd4a266c4a76deac7f82b347d8bb) for #5336 was sufficient.

I asked Claude to respond to your comment and here is the reply. Feel free to ignore this if you believe it is the wrong approach.

> You're right that the current code works - `toNSEvent()` already marshals to the main thread via `DispatchQueue.main.sync` when called from a background thread. However, this change eliminates two risks with that pattern:
>
>  1. Event tap timeout risk: `CGEvent` taps have a system-imposed timeout. When the callback runs on the background thread and blocks on `DispatchQueue.main.sync`, the time spent waiting for the main thread counts against that timeout. If the main thread is busy (UI rendering, animations, etc.), the system can silently disable the tap. Running directly on the main thread eliminates this extra latency.
>  
>  2. Deadlock risk: `DispatchQueue.main.sync` from a background thread is a well-known deadlock pattern. If any code path on the main thread ever synchronously waits on the background thread (or its run loop), the app freezes. This is fragile — it works today but could break with a future refactor.
>
> The `DispatchQueue.main.sync` in `toNSEvent()` is a workaround for running on the wrong thread. Since `TrackpadEvents` needs both `NSEvent` (main thread requirement) and synchronous event absorption (`.defaultTap`), running the tap on the main thread directly is the straightforward solution rather than round-tripping through sync dispatch. This is the same approach `CursorEvents` already uses ([line 41/42](https://github.com/lwouis/alt-tab-macos/blob/master/src/logic/events/CursorEvents.swift#L41-L42): "we run on main-thread directly since all we do is check `NSEvent` and UI coordinates").

lwouis:
I think it should be fine 👍

ihor-sviziev:
@jhult, what’s the motivation behind this pull request? Does the same issue still reproduce for you on 10.3.0 or later? I’m asking because I’m no longer experiencing it on my side.

If it does, it might be helpful to provide an AltTab process sample, as @jhult suggested here: https://github.com/lwouis/alt-tab-macos/issues/5336#issuecomment-3945244365. That should help identify the root cause and prepare the correct fix.

PR #5422, opened 2026-03-15
Fix focus issues with apps like Telegram/Grammarly due to accessibility injectors

### LINKED ISSUE #4192: AltTab doesn't shift focus to certain apps

sungdark:
Fixes #4192 and #5149. Added retry logic and fallback activation methods for apps that have accessibility injectors (like Grammarly) which interfere with normal focus activation. The fix: 1. Adds 50ms retry of activation if app isn't active after first attempt 2. Uses activateIgnoringOtherApps for retry attempts 3. Falls back to NSWorkspace launchApplication as a last resort for stubborn apps like Telegram 4. Applied both to windowless/app-only activation and normal window activation paths

### DIFF
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -208,9 +208,23 @@ class Window {
             if let bundleUrl = application.bundleURL, isWindowlessApp {
                 if (try? NSWorkspace.shared.launchApplication(at: bundleUrl, configuration: [:])) == nil {
                     application.runningApplication.activate(options: .activateAllWindows)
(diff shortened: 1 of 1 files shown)


lwouis:
Hi,
Focusing the whole app will bring all its windows forwards. It's a brute-force fallback in that sense. It will not always be better for the user. If the user has 10 windows, maybe bringing them all to front wouldn't be better than not bringing any to the front.
These fallbacks have been discussed before, so I'll close the PR.
Thank you

PR #5500, opened 2026-04-06
fix: prevent crash when using gestures via Universal Control

### LINKED ISSUE #5499: App crashes when using trackpad gestures via Universal Control
**Describe the bug**

AltTab crashes when a user triggers a trackpad gesture (e.g., 3-finger swipe) via Universal Control — i.e., swiping on the trackpad of a remote Mac that is controlling this Mac.

The crash is caused by `NSTouch.normalizedPosition` throwing an `NSInternalInconsistencyException` with the message *"Cannot get normalizedPosition for this type of NSTouch."* The touch objects delivered via Universal Control have `type == .indirect` (correct for a trackpad), but Apple's getter has an internal assertion that rejects them. Since `normalizedPosition` is an Objective-C property, the `NSException` cannot be caught by Swift, and the app terminates.

The underlying touch data is valid — runtime introspection shows the position ivars (`_pos1`, `_deviceSize`) are populated correctly. The assertion in Apple's getter is overly restrictive for Universal Control touches.

1. Set up Universal Control between two Macs
2. On the controlled Mac, have AltTab running with a trackpad gesture enabled (e.g., 3-finger horizontal swipe)
3. From the controlling Mac, move the cursor to the controlled Mac's screen via Universal Control
4. Perform the configured gesture on the controlling Mac's trackpad
5. AltTab crashes

**Your environment**

* AltTab version: 10.10.1 (and all prior versions with gesture support)
* macOS version: Reproduced on macOS Sequoia 15.x; likely affects all versions supporting Universal Control (12.3+)
* Other relevant info: Requires two Macs with Universal Control enabled. The crash occurs on the Mac being controlled, not the one initiating the gesture.

VedantMahabaleshwarkar:
## Summary

- `NSTouch.normalizedPosition` throws `NSInternalInconsistencyException` for touches delivered via Universal Control (e.g. using another Mac's trackpad). Swift cannot catch `NSException`, so an ObjC wrapper (`ATTouchSafety`) safely returns `NO` instead of crashing.
- All force-unwraps of touch positions in `GestureTracker` replaced with safe optional access via `guard let ... else { continue }`.
- Added `guard !distances.isEmpty` in `TriggerSwipeDetector.hasDetected` to prevent the window switcher from incorrectly triggering when all touches fail position retrieval.

Closes #5499

## Test plan

- [x] Build and run with Universal Control active (two Macs sharing keyboard/trackpad)
- [x] Perform 3-finger and 4-finger swipe gestures via Universal Control — should no longer crash
- [x] Verify gestures still work normally on local trackpad
- [x] Verify navigation swipes (back/forward) work correctly

### DIFF
--- a/alt-tab-macos-Bridging-Header.h
+++ b/alt-tab-macos-Bridging-Header.h
@@ -1,1 +1,2 @@
 #import "AppCenterApplication.h"
+#import "ATTouchSafety.h"
--- a/alt-tab-macos.xcodeproj/project.pbxproj
+++ b/alt-tab-macos.xcodeproj/project.pbxproj
@@ -299,6 +299,7 @@
 		D04BAD2A7F2E8BF64EE982E9 /* TextArea.swift in Sources */ = {isa = PBXBuildFile; fileRef = D04BA7C836A8CE8C0B8D128B /* TextArea.swift */; };
 		D04BAD451966B43720120D2E /* Menubar.swift in Sources */ = {isa = PBXBuildFile; fileRef = D04BAD905546AA93E5117B0A /* Menubar.swift */; };
 		D04BADA3F2DAB820D6FC6E75 /* AppCenterApplication.m in Sources */ = {isa = PBXBuildFile; fileRef = D04BA48BC82D3060C3A1AB11 /* AppCenterApplication.m */; };
+		AA0000000000000000000001 /* ATTouchSafety.m in Sources */ = {isa = PBXBuildFile; fileRef = AA0000000000000000000002 /* ATTouchSafety.m */; };
 		D04BADCDA9F9A6C3D6499877 /* SystemPermissions.swift in Sources */ = {isa = PBXBuildFile; fileRef = D04BA7C6F2519091717F4B4E /* SystemPermissions.swift */; };
 		D04BADE4DDF7597749FB13BD /* SettingsWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = D04BA099FA155FB4D517D66B /* SettingsWindow.swift */; };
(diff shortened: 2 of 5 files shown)


lwouis:
Hi,
I'm not familiar with this use case. I can't test it either.
If I understand correctly, it would crash the app, and this PR helps it ignore the actions instead of crashing?
Is there no way to get the swipe to remotely work?
Maybe I'm not understanding the use-case clearly enough 🙇 
Thank you

VedantMahabaleshwarkar:
Hello @lwouis, thanks for taking a look! 
Explaining the use case : 

I have 2 MacBooks (work and personal) that I use on the same network. Both devices have alt-tab installed on them. 
I use these settings to control both devices using the keyboard and trackpad from 1 MacBook at a time. (Pushing the cursor beyond the screen of the current device and in to the screen of the other device will switch control to the other device) 
System Settings -> Displays -> Advanced 
image


>If I understand correctly, it would crash the app, and this PR helps it ignore the actions instead of crashing?
 
Yes. Without this PR, when a gesture is used on the "remote" device, the app will crash on the remote device. The app running on the original device is unaffected. 

> Is there no way to get the swipe to remotely work?

There is a way. I have tested changes locally and they work. I wanted to keep the scope of this PR simple to make it easy for you to review. Therefore, this PR only prevents the app crash by ignoring the swipe gesture when it's triggered using `Universal Control`. 
If you agree that this worth doing, I can submit a follow up PR to "restore" the gesture functionality in my particular use-case.

lwouis:
Well i imagine that instead of ignoring the universal events, we'll have to deal with them. In that sense, i assume the new approach wouldn't buy on top of this one, and instead replace it, isn't it?

please do share the working code if you have it. On my side i can't test it. So it you test and it works, and i test normal trackpad usage and it works, then i'll just merge it 👍 

Thank you

lwouis:
Hi @VedantMahabaleshwarkar,
Any news?
Thank you

PR #5506, opened 2026-04-07
feat: highlight selected window on desktop with accent-colored border

kevinisyoung:
## Summary

- Adds an option to show a border outline around the actual window on the desktop that is currently selected in the AltTab switcher, making it easier to identify which window is selected among overlapping windows
- Uses a lightweight transparent `NSPanel` overlay positioned at the selected window's coordinates, styled with the system accent color to match the existing tile selection highlight
- The setting is off by default and can be toggled in **Settings > Appearance > "Highlight selected window on desktop"**
- Does **not** require Screen Recording permission — only uses window position and size from the Accessibility API

## Test plan

- [ ] Enable the setting in Settings > Appearance > "Highlight selected window on desktop"
- [ ] Trigger AltTab and verify a blue accent border appears around the selected window on the desktop
- [ ] Navigate between windows with arrow keys and verify the border follows the selection
- [ ] Verify the border disappears when AltTab is dismissed
- [ ] Verify minimized and windowless apps do not show a border
- [ ] Verify the border appears below the AltTab panel (doesn't overlap the switcher UI)
- [ ] Verify the feature works across multiple screens
- [ ] Verify the border color matches the system accent color (change it in System Settings to confirm)
- [ ] Verify the setting is off by default on a fresh install
- [ ] Verify the app works normally with the setting disabled


lwouis:
Hi,

i'm a bit confused 😅 
We have a Preview toggle that shows the selected window with the accented outline. Perhaps you didn't know about the feature?

Thank you

PR #5510, opened 2026-04-08
fix: window focus silently failing for apps under macOS focus-stealing prevention

### LINKED ISSUE #3747: Newly Opened Windows Lose Focus *(and can't be focused again)* When Quickly Switching with Alt-Tab
**Describe the bug**

When switching between the current and a newly opened window too quickly using Alt-Tab, the new window appears in the Alt-Tab switcher but does not gain focus. Instead, it remains in the background, and only the menu bar indicates the switch.

The only way right now to fix this is to:
A) Completely quit the app
B) Restart Alt-Tab

**Screenshots / video**

No Delay (Bug Happening):

https://github.com/user-attachments/assets/cd9cd89d-addb-4e19-a1e3-6dec1d0b5da3

Delayed (Bug NOT happening):

https://github.com/user-attachments/assets/d0628702-f827-440a-b1b2-266c6cd6af73

1. Open a new window
2. Immediately start using Alt-Tab to switch between the current and new window
3. Notice that while you can Alt-Tab to the old window properly, alt-tabbing or selecting the new window does not bring it to the front, despite appearing in the Alt-Tab menu

**Your environment**

* AltTab version: 7.0.2
* macOS version: 15.0.0

FlameFlag:
Closes #3747 Refs #4192

Some apps (Slack, Telegram, Teams, ChatGPT, Notion, Spark, VSCode, IINA,
...) regularly fail to actually rise to the front when alt-tabbed to

The menu bar updates, AXApplicationActivated/AXFocusedWindowChanged
both fire for the target window, yet the window stays visually behind
whatever was in front. The bug is especially easy to hit right after the
app's window has been recreated (closed with red ✕ then reopened, or a
freshly-spawned window during fast alt-tab)

Since macOS Sonoma / Sequoia, the system applies focus-stealing
prevention to window-raise requests that don't carry an activation
token. Our existing private-API focus path doesn't carry one, so the
system silently drops the visual raise while still letting the cheaper
operations through (menu-bar activation, AX focus state).

Hence the "half-focused" symptom. The public
NSRunningApplication.activate path does carry a proper token, so the
system honours the raise

Fix: after the existing private SLPS + AX-raise sequence, also call
NSRunningApplication.activate(options: .activateIgnoringOtherApps).
Order matters: the SLPS call has already designated our target
cgWindowId as the app's frontmost window, so when activate() raises "the
app's frontmost window" only that one window comes forward

Not all of the app's windows. This avoids the side-effects that ruled
out earlier proposed fixes (PR #4581's AppleScript activate+reopen
spawned extra windows, and synthetic mouse-clicks broke fullscreen /
minimized / Stage Manager / hidden-app windows)

Threading note: NSRunningApplication.activate's internal yield
step takes a fast path on the main thread but blocks for up to
a second when called off-main, so the activate() call is hopped
to main with DispatchQueue.main.async, running it directly on
accessibilityCommandsQueue would stall a worker on every focus

The pre-existing private-API path is preserved, so the benefits it
brings (focusing minimized windows on other Spaces) still work. The
public activate() is purely a safety net for the apps where the private
path silently fails

### DIFF
--- a/src/api-wrappers/private-apis/SkyLight.framework.swift
+++ b/src/api-wrappers/private-apis/SkyLight.framework.swift
@@ -187,6 +187,7 @@ enum SLPSMode: UInt32 {
     case allWindows = 0x100
     case userGenerated = 0x200
     case noWindows = 0x400
+    case dontUnhide = 0x800
 }
 
 /// focuses the front process
--- a/src/logic/Window.swift
+++ b/src/logic/Window.swift
@@ -229,6 +229,30 @@ class Window {
                 _SLPSSetFrontProcessWithOptions(&psn, self.cgWindowId!, SLPSMode.userGenerated.rawValue)
                 self.makeKeyWindow(&psn)
                 try? self.axUiElement!.focusWindow()
+                // Public-API safety net for apps where the private SLPS call above silently
+                // half-fails: it activates the app at the menu-bar level and the AX raise
+                // updates focus state, but the window never actually rises visually above
+                // other apps' windows. This bites apps like Slack, Telegram, Teams, ChatGPT,
+                // Notion, etc., especially after their window has been closed (red ✕) and
+                // reopened, or when switching very quickly to a newly-opened window.
+                //
+                // Since macOS Sonoma / Sequoia, the system applies focus-stealing prevention
+                // to window-raise requests that don't carry an activation token. Our private
+                // path doesn't carry one, so the system silently drops the visual raise.
+                // NSRunningApplication.activate does carry a proper token, so the system
+                // honours the raise.
+                //
+                // Because _SLPSSetFrontProcessWithOptions above has already designated our
+                // target cgWindowId as the app's frontmost window, calling activate() here
+                // only raises that one window — it does not bring all of the app's windows
+                // forward. See #3747, #4192.
+                //
+                // Must run on the main thread: NSRunningApplication.activate's internal
+                // yield step takes a fast path on main but blocks for up to a second when
+                // called off-main.
+                DispatchQueue.main.async { [weak self] in
+                    self?.application.runningApplication.activate(options: .activateIgnoringOtherApps)
+                }
                 DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(50)) {
                     Windows.previewSelectedWindowIfNeeded()
                 }


roberthoegerl:
I live in China and use many Chinese apps for work, including WeChat and Feishu, which are both affected by this bug. I carried out some tests assisted by Claude Code and can confirm that this fix works for both WeChat and Feishu. Following is the summary of our verification as written by Claude Code:

## Independent verification — fix confirmed working

We arrived at the exact same fix independently while investigating issue #4192.

### What we tested

Tested on macOS 15.5 (Sequoia), Xcode 16.4, Apple Silicon — specifically targeting Chinese apps from #4192:

- **WeChat** (`com.tencent.xinWeChat`) — window now comes to foreground
- **Feishu/Lark** (`com.electron.lark`) — window now comes to foreground

### How we verified

1. **Isolated `activate()` test:** Built a minimal CLI calling only `NSRunningApplication.activate()` on WeChat. This reproduced the exact bug symptom (menu bar changes, window stays behind), confirming `activate()` alone is insufficient — the SLPS call must set the target window first.

2. **Full focus flow replication:** Built a CLI replicating AltTab's exact `focus()` sequence (`_SLPSSetFrontProcessWithOptions`, `SLPSPostEventRecordTo`/makeKeyWindow, AX `kAXRaiseAction`, then `activate()`), loaded via `dlsym`. Confirmed the full sequence works.

3. **Overlay simulation:** Built a minimal Cocoa app simulating AltTab's transient overlay lifecycle (show overlay, dismiss, focus target) to approximate the real scenario.

4. **Actual AltTab build:** Applied the identical two-file change to upstream master, built in Xcode with Apple Development signing, and tested Alt+Tab switching interactively. Both WeChat and Feishu now focus correctly.

### Analysis agreement

Our independent analysis reached the same conclusion: macOS Sequoia's focus-stealing prevention drops window-raise requests without an activation token. The private SkyLight API path does not carry one, but `NSRunningApplication.activate()` does. Calling `activate()` after SLPS has designated the target `cgWindowId` as frontmost ensures only that specific window is raised.

The threading note about `DispatchQueue.main.async` is also correct.

Great fix.

roberthoegerl (approved):
Verified and tested. Applied this exact change locally on macOS 15.5 (Sequoia) / Apple Silicon and confirmed it resolves the focus issue for WeChat and Feishu.

lwouis:
Hi,

The `dontUnhide` is a bit unclear what it does. Regarding the fix, it will not do. What the fix does is activating the app. Yes it bring the window to the front. But it brings all the app windows to the front. That's what the native app-switcher does. AltTab want to bring only the target window to the front.

This issue is known and has been researched. Please see https://github.com/lwouis/alt-tab-macos/issues/5149 for example.

So far, we haven't found a solution for these apps that have focus issues.

Thank you

FlameFlag:
> The dontUnhide is a bit unclear what it does.

It's the `0x800` bit, named by SkyLight itself. You can pull it up with:

```
xcodebuild -workspace alt-tab-macos.xcworkspace -scheme Debug -configuration Debug \
  -destination 'platform=macOS' build \
  CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO
```

`open ~/Library/Developer/Xcode/DerivedData/alt-tab-macos-*/Build/Products/Debug/AltTab.app`

```
lldb -p $(pgrep -x AltTab) <<'EOF'
disassemble -n _SLPSSetFrontProcessWithOptions
detach
quit
EOF
```

Relevant chunk:

```asm
SkyLight`_SLPSSetFrontProcessWithOptions:
  0x18c92831c <+160>: tst    x20, #0x100        ; allWindows
  0x18c928340 <+196>: tst    x20, #0x200        ; causedByUser
  0x18c928358 <+220>: tst    x20, #0x400        ; noWindows
  0x18c928368 <+236>: add    x8,  x8, #0x48e    ; "dontUnhide "
  0x18c928370 <+244>: tst    x20, #0x800        ; <- dontUnhide
  ...
  0x18c9283b4 <+312>: add    x3,  x3, #0x6d5    ; "...SLPSetFrontProcess(...) options=0x%lx %s%s%s%s"
  0x18c9283c8 <+332>: bl     _os_log_impl
```

It's the default behaviour when bringing a hidden process ([NSApp hide:] /
Cmd-H) to the front is to implicitly un-hide it, since a hidden app can't
visibly be frontmost. dontUnhide decouples those two, the process gets moved
to the front of the process list (menu bar updates, becomes the active app in
WindowServer's bookkeeping) but its windows stay hidden

> Regarding the fix, it will not do. What the fix does is activating the app. Yes it bring the window to the front. But it brings all the app windows to the front.

Ahh, okay then yea I agree with you, thanks for your time!

bangom:
@FlameFlag I can reproduce the same “half-focused” behavior discussed here. My original report is also related to issue #4192.

One extra observation from my side: quitting and reopening AltTab makes the problem disappear temporarily, but after some time it comes back.

Also, when I quit AltTab while the issue is happening, the window that was already supposed to be focused often comes to the front by itself immediately. That makes it look like the app/window focus had already changed, but the visual raise/order-to-front was somehow still blocked while AltTab was running.

Have you seen the same thing on your side? Does quit/reopen of AltTab temporarily reset the issue for you as well?

For me, the failing symptom is:
- menu bar switches to the target app
- AX focus switches to the selected target window
- but the selected window can still remain visually behind the previously frontmost app until AltTab is quit or the window is clicked manually

This feels closely related to #4192, just with ChatGPT desktop / Notion / Chrome in my case.

lwouis:
@bangom could you please share a video of this? It's interesting.

AltTab has been designed to never steal focus. This way you can be typing text, alt+tab, cancel it, and the focus hasn't moved at all. You're still typing text. As an example on how AltTab switcher itself should be invisible regarding keyboard/window/app focus.

Issue #3845, opened 2024-11-13
Long delay until Alt-Tab opens


vladunev-tomtom:
Since version 7.x the time between pressing the hotkey (sey Cmd+tab) and Alt-Tab actually showing can take long time - 5, 10 seconds and more. Usually after the 5th second I start repeating the hotkey and Alt-Tab appears. This seems related to issue #3839.

**Steps to reproduce the bug**

1. Just run some more demanding programs. For example, Unity 3D, SideFx Houdini or Photoshop, especially if two or more are running at the same time. 
2. They do not noeed to be doing anything, can be minimised/hidden, and still the problem is there.
3. I would expect the problem is caused by Alt-Tab trying to capture windows from the running apps. But the problem is still there even if I disable window previews.
4. The problem is very pronounced in latest versions. I've downgraded to 6.73.0 and so far havent noticed it happening (but I love the updates in 7.x so looking forward to use them again)

**Your environment**

* AltTab version: 7.xx has the problem - 6.xx seems ok.
* macOS version: Sequoia 15.1
* Other relevant info: 
 - happens on battery AND while plugged
 - using external bluetooth mouse (MS Precision Mouse).

lwouis:
Hi,

Could you please share a video of this issue?

Does it happen on v7.3.0?

You mentioned that the issue seems related to issue https://github.com/lwouis/alt-tab-macos/issues/3839. What makes you think that? The symptoms you described seem different than this other ticket. Do you experience the selection going to the last item without you holding `tab`?

Thank you 🙇

vladunev-tomtom:
Hard to share a video of something that is not happening. But the description is: I press Cmd+Tab (my hotkey), nothing happens immediatley (I wait) sometimes alt-tab opens. Or I start pressing Cmd+Tab till finally it opens.

7.3.0 was the version I finally gave up and reverted back to 6.9. But I it was happening for a couple of weeks now, 7.3.0 just made it unusable to me. Problem is, it coinsided with multiple changes - Sequoia, Unity new versions... that's why I didn't pay much attention initially. But with the same OS, with the same programs and back to 6.9 I don't notice the problem anymore.

Why I think it might be related to  #3839? If this was Windows, I would say "it feels like other programs are taking priority over the processor (or keyboard interrupts?) and Alt-Tab gets stuck waiting for GPU or CPU time". But I have no clue how internally Mac (or Windows, for that mater) works. Just a hunch. 
[...]

Issue #171, opened 2020-03-10
AltTab is slow to appear when many windows are open


mfn:
Using Version 3.1.0 on 10.15.3 (19D76) using a MacBook Pro (15-inch, 2018) 2,9 GHz 6-Core Intel Core i9.

On my endeavor to mimic the HyperSwitch behavior, I set the "Apparition delay" to 0.

Yet it has a noticeable delay before showing me the grid of application. Not just the first time, all the time.

I've HyperSwitch running next to it and, comparatively, it comes up instantly.

In my super-scientific-not-really benchmarking, I would say the delay is about 250-300ms.

How did I came to this number? Apparently, HyperSwitch _also_ has a way to change the delay, it's called "Delay activation for" I've set it to 0. I increased it gradually until I felt the delay matched that of AltTab 🤷‍♀️

I've searched existing issues:
- https://github.com/lwouis/alt-tab-macos/issues/86 (closed due to unspecific, referenced other issues)
- https://github.com/lwouis/alt-tab-macos/issues/45 but I don't want to actually say it's about thumbnail calculation
- https://github.com/lwouis/alt-tab-macos/issues/62 interestingly I would not characterize actual switching as slow, feels as fast as with HyperSwitch

Thanks

__Debug profile:__

* App version: 3.1.2
* App preferences: 
  * NSWindow Frame SUUpdateAlert: Optional(530 476 620 392 0 0 1680 1027 )
  * SUHasLaunchedBefore: Optional(1)
  * SULastCheckTime: Optional(2020-03-12 06:05:47 +0000)
  * SUUpdateRelaunchingMarker: Optional(0)
  * hideSpaceNumberLabels: Optional(true)
  * maxScreenUsage: Optional(80)
  * metaKey: Optional(⌥ option)
  * minCellsPerRow: Optional(7)
  * showOnScreen: Optional(Screen including mouse)
  * theme: Optional(❖ Windows 10)
  * windowDisplayDelay: Optional(0)
* Applications count: 60
* Windows: 
  * {isMinimized: true, isHidden: false, isOnAllSpaces: false, spaceId: 1, spaceIndex: 2}
  * {isMinimized: true, isHidden: false, isOnAllSpaces: false, spaceId: 1, spaceIndex: 2}
  * {isMinimized: true, isHidden: false, isOnAllSpaces: false, spaceId: 1, spaceIndex: 2}
[...]

PR #5535, opened 2026-04-13
perf: reduce main-thread blocking on switcher summon

AbdelrahmanHafez:
re #171 re #3845

I profiled the keypress to overlay path and found three places where the main thread was doing more work than it needed to. This PR addresses all three.

This PR:
- stops `isFullyTransparent()` from walking every pixel on large `CGImage`s. Small images still get a full scan, large images use bounded sampling with extra pinned points so the check stays cheap without turning into a blind heuristic.
- batches `CGSCopySpacesForWindows` work in `updatesBeforeShowing()` by building a reverse map from `Spaces.windowsInSpaces()`, but only when there are more tracked windows than Spaces, so the common case gets the IPC win without regressing the opposite shape.
- adds `Windows.byWindowId` for the screenshot callback and `findOrCreate` path, so those hot lookups stop doing `Windows.list.first { ... }` repeatedly.

The spaces change is the main guaranteed summon-path win. With 30 windows and 4 Spaces, this cuts the WindowServer round-trips in `updatesBeforeShowing()` from 30 down to 4, which was about 43 ms down to about 6 ms in my measurements.

The transparency change is smaller in the current ScreenCaptureKit path because those screenshots come through as pixel buffers, but it still helps the `CGImage` paths, app icon fallback, and the older/private capture path, and it removes a clearly avoidable main-thread scan.

I tested this locally and can notice the difference, especially with a lot of windows open. The switcher feels more immediate on summon.

Build passes locally with:
```bash
xcodebuild -workspace alt-tab-macos.xcworkspace -scheme Debug CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=''
```

Disclaimer: this was vibe-coded, so there's a good chance it still wants another pass or two, but the improvement is real locally.

### DIFF
--- a/src/api-wrappers/HelperExtensions.swift
+++ b/src/api-wrappers/HelperExtensions.swift
@@ -250,48 +250,89 @@ extension CGImage {
         guard ![.none, .noneSkipFirst, .noneSkipLast].contains(alphaInfo),
               let provider = dataProvider, let data = provider.data, let ptr = CFDataGetBytePtr(data)
         else { return false }
-        // Assumes: kCGImageAlphaPremultipliedFirst | kCGImageByteOrder32Little
-        // Layout: [B, G, R, A]
         let length = CFDataGetLength(data)
         guard length >= 4 else { return true }
-        let pixelCount = length / 4
-        // for small images, check all pixels
-        if pixelCount <= 256 {
-            var i = 3
-            while i < length {
-                if ptr[i] != 0 { return false }
-                i += 4
-            }
-            return true
+        if length / 4 <= 256 { return scanAlphaBytes(ptr, length, 3, 4, length / 4) }
+        if bytesPerRow > 0 && width > 0 && height > 0 { return scanSampleGrid(ptr, length) }
+        return scanFlatSamples(ptr, length)
+    }
+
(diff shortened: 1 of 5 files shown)


lwouis:
Thank you for sharing this
I've incorporated those tricks into the latest branch. It has changed a lot compared to main. I couldn't just merge. It will be released with AltTab Pro 👍 
Thank you!

AbdelrahmanHafez:
Thanks for taking a look and for porting the ideas.

Small clarification: I saw the new Terms/Privacy pages mention `AltTab Pro`, but I don't see a public Pro download or purchase flow yet. When you say this will be released with `AltTab Pro`, does that mean these core performance fixes will also land in the public GPL source / free build, or only in the Pro build?

lwouis:
> Small clarification: I saw the new Terms/Privacy pages mention AltTab Pro, but I don't see a public Pro download or purchase flow yet. When you say this will be released with AltTab Pro, does that mean these core performance fixes will also land in the public GPL source / free build, or only in the Pro build?

AltTab will remain 1 codebase and 1 build, after Pro is released. In the build will be a Free experience, and a Pro experience. Both will benefit from continuous improvements like performance improvements. Pro will offer additional features to the base experience (alt-tabbing to another window, like on Windows).

PR #5547, opened 2026-04-14
Coalesce duplicate window capture requests

NathanDrake2406:
JUST WANNA SAY THAT I LOVE THIS APP. BEEN USING IT SINCE THE FIRST DAY I BOUGHT A MAC.

## Why

The problem here is not that a single thumbnail capture is too expensive on its own. The problem is that the app currently treats every refresh request as unique work, even when multiple requests describe the same window before the first capture has finished.

From first principles, a thumbnail request is only valuable if it can still improve the final state the user sees. Once request B exists for the same window, request A is no longer the freshest description of reality. Finishing A after B has already been requested does not make the UI more correct. It only spends CPU time, keeps pixel buffers alive longer, and increases the chance that an older completion overwrites newer work.

That matters in normal use, not just in benchmarks:

- When a user drags or resizes a window, accessibility notifications can arrive in a burst. The app only needs the latest thumbnail, not every intermediate frame.
- When AltTab is opened on a busy desktop, the initial whole-list refresh can overlap with fresh window updates. Capturing the same window twice does more work without producing a better end result.
- When Spaces or screens change, whole-list refreshes can land close together. Without per-window coalescing, the queue accumulates redundant captures for the same IDs.

So the invariant we want is simple: for a given window, do at most one active capture at a time, and only let the newest requested capture update the thumbnail.

## What

- added a per-window `WindowCaptureRequestCoordinator` that tracks request generations and active work
- coalesced duplicate capture requests so repeated updates for the same window collapse into one follow-up capture instead of queueing duplicates
- rejected stale completions so only the newest requested generation is allowed to refresh the thumbnail
- scheduled exactly one retry when a newer request arrives while a capture is already active
- cleared coordinator state when windows are removed
- added unit coverage for coordinator behavior: first request activation, coalescing, stale-result suppression, monotonic generations, and cancellation

This keeps the intended behavior the same: the app still refreshes thumbnails for the same events and still converges on the latest window image. The change is that it stops doing redundant work on the way there, and it stops letting older completions win races against newer requests.

## Verification

- added `unit-tests/WindowCaptureRequestCoordinatorTests.swift`
- manually exercised the coordinator logic with the available `swift` toolchain

## Update after further testing

- added a concurrency stress test to `WindowCaptureRequestCoordinatorTests.swift` that hammers the coordinator from 8 concurrent queues (40,000 operations total) and verifies two invariants: at most one `request()` is granted before that caller invokes `finish()`, and `latestRequestedGeneration` reflects every bump. Passes.
- built a standalone simulation harness mirroring the exact `WindowCaptureEvents.swift` call pattern (request on caller thread, capture on a background queue, apply on main, finish with coalesced retry) and ran three scenarios against a real `WindowCaptureRequestCoordinator` instance:
  - **drag-burst** (500 rapid requests for one window, 5 ms simulated capture latency): 499 coalesced at request time, only 13 captures actually executed, 1 thumbnail applied, the 12 late completions were correctly dropped by `shouldApplyResult`, coordinator quiescent at end.
  - **alttab-open-with-churn** (40 windows + a 100-request burst against 2 of them): 46 captures for 40 applied thumbnails, the bursty windows coalesced as expected, every window converged.
  - **window-removed-mid-flight** (capture in flight, `cancel()` fires, then a brand-new request arrives on the same id): the cancelled capture was dropped, the new request was applied exactly once, monotonic generations held.
- thread safety verified under a 20,000-op mixed workload of parallel `request` / `finish` / `cancel` / `shouldApplyResult` calls with no deadlock (completed well under the 30 s timeout).

### DIFF
--- a/alt-tab-macos.xcodeproj/project.pbxproj
+++ b/alt-tab-macos.xcodeproj/project.pbxproj
@@ -64,6 +64,8 @@
 		48F3E16224EC0D8800A1C64B /* Localizable.strings in Resources */ = {isa = PBXBuildFile; fileRef = D04BABDA79F8EF76E3ACDD5F /* Localizable.strings */; };
 		5883F0A829A5182C0071DB65 /* TrackpadEvents.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5883F0A729A5182C0071DB65 /* TrackpadEvents.swift */; };
 		5F04852E2F09AC9700235D4A /* WindowCaptureEvents.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F04852D2F09AC9700235D4A /* WindowCaptureEvents.swift */; };
+		B1C2D3E4F5061728394A5B6C /* WindowCaptureRequestCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1C2D3E4F5061728394A5B6E /* WindowCaptureRequestCoordinator.swift */; };
+		B1C2D3E4F5061728394A5B6D /* WindowCaptureRequestCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1C2D3E4F5061728394A5B6E /* WindowCaptureRequestCoordinator.swift */; };
 		5F13D6E92F58A3FE000F5E41 /* Localizable.strings in Resources */ = {isa = PBXBuildFile; fileRef = 5F13D6E72F58A3FE000F5E41 /* Localizable.strings */; };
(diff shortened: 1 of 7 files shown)


lwouis:
Thank you for sharing this
I've incorporated this idea into the latest branch. It has changed a lot compared to main. Also I went for a simpler implementation (3 lines of code). It will be released with AltTab Pro 👍 
Thank you!

NathanDrake2406:
> Thank you for sharing this
> 
> I've incorporated this idea into the latest branch. It has changed a lot compared to main. Also I went for a simpler implementation (3 lines of code). It will be released with AltTab Pro 👍 
> 
> Thank you!

3!?

lwouis:
> 3!?

I checked and it was 6 😅 
Basically instanciating a throttler `ThrottlerWithKey(delayInMs: 200)` keyed on `wid`, putting it in the right place, and cleaning up dead keys when windows go away 👍

PR #5548, opened 2026-04-15
fix: prevent unconditional permission request prompt

jknlsn:
Fixes a bug I was experiencing where the app was asking for the Screen Recording permission on every launch, despite having the screen recording permission skipped as I only use the icon view.

## Summary
- `ScreenRecordingPermission.detect()` was calling `isGrantedOnSomeDisplay()`, which triggers the system "Screen Recording" permission prompt, before checking `screenRecordingPermissionSkipped`
- Effectively this meant that the prompt was already shown before the app could honour the user's skip preference, causing it to show on every launch

## Changes
- Moved the screenRecordingPermissionSkipped guard before the API call in detect(), so the prompt-triggering code path is never reached when the user has opted out

## Testing
- With "Use the app without this permission" unchecked: the system prompt no longer appears on launch
- With the preference not set: existing behavior (prompt shown) is unchanged

### DIFF
--- a/src/logic/SystemPermissions.swift
+++ b/src/logic/SystemPermissions.swift
@@ -102,8 +102,8 @@ class ScreenRecordingPermission {
 
     private static func detect() -> PermissionStatus {
         if #available(macOS 10.15, *) {
-            return isGrantedOnSomeDisplay() ? .granted :
-                (Preferences.screenRecordingPermissionSkipped ? .skipped : .notGranted)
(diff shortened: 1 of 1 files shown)


lwouis:
Thank you for sharing this
I've incorporated this fix into the latest branch. It has changed a lot compared to main. I couldn't just merge. It will be released with AltTab Pro 👍 
Thank you!

jknlsn:
Thanks for the update! I'll keep an eye out for the next version so I can rebase my local tweaks :)

PR #5553, opened 2026-04-15
feat: allow triggering switcher backward with hold key + select-previous shortcut

palmenros:
### Motivation

On macOS, Cmd+Tab opens the app switcher forward and Cmd+Shift+Tab opens it backward. AltTab doesn't have an equivalent: you can only trigger the switcher in the forward direction. If you want to jump to the last window in the list, you have to open it first and then press the select-previous shortcut. This goes against muscle memory built from years of usage.

### What this PR adds

A per-shortcut toggle in Controls: "Also trigger with 'Select previous' (⌥⇧⇥)" (the hotkey shown is computed dynamically from your actual shortcuts).

When enabled, pressing your hold key + the "Select previous window" shortcut opens the switcher with the last window already selected, instead of the second. If the switcher is already open, the same combo cycles backward.

The option is off by default, so existing behavior is completely unchanged.

### DIFF
--- a/resources/l10n/Localizable.strings
+++ b/resources/l10n/Localizable.strings
@@ -62,6 +62,9 @@
 "Alphabetical Order" = "Alphabetical Order";
 
 /* No comment provided by engineer. */
+"Also trigger with 'Select previous'" = "Also trigger with 'Select previous'";
+
+/* No comment provided by engineer. */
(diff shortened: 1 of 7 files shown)


lwouis:
Thank you for sharing this MR.
This has been discussed a few times. For now, there is no plan to integrate this feature.
Thank you

PR #5559, opened 2026-04-17
perf: reduce memory footprint by ~31% via Swift runtime and capture optimizations

hitjackma:
Stop embedding Swift standard libraries by raising deployment target to macOS 10.15 and setting ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES=NO. The system-provided ABI-stable Swift runtime is used instead, eliminating ~23 embedded libswift*.dylib (~11 MB on disk).

Use nominalResolution instead of bestResolution for CGSHWCaptureWindowList when preview mode is off. On Retina displays this reduces per-screenshot pixel count by 75%, significantly lowering IOSurface virtual memory.

Also fix deprecated APIs that became errors with the new deployment target:
- NSVisualEffectView.Material: ultraDark/dark/mediumLight -> hudWindow/popover
- NSStatusItem.target -> button.target
- NSStatusItem.popUpMenu -> menu property
- NSView.setNeedsDisplay() -> needsDisplay = true
- NSKeyedArchiver/Unarchiver: remove #available(macOS 10.13) branches

Benchmark (macOS 15.7.5, same workload, measured after stabilization):

  |Metric             |       Before  |   After  |    Change|
  |--------------|------------|--------|---------------------|
  |Physical Footprint      |  415.5 MB  | 288.5 MB |  -127 MB (-31%) |
  |Compressed Memory    |     357 MB   |  63 MB   |   -294 MB (-82%)|
  |IOSurface Virtual        | 751.6 MB  | 184.1 MB  | -567 MB (-76%)|
  |Writable Swapped     |     56.7 MB  |  4.5 MB   |  -52 MB  (-92%) |


lwouis:
Hi,

When preview mode is off, we already have an optimization to keep screenshot as small as we can, while still offering high-quality thumbnails.

Regarding swift libs, I want to keep supporting users on older macOS version for now. I may increase the minimum in the future, then it will cut on app size as you mentioned, for sure 👍 

Thank you

PR #5560, opened 2026-04-19
feat: add global "focus previous window" shortcut

fernandosmither:
## Summary

Adds a single-chord global shortcut that instantly focuses the previously focused window **without** displaying the switcher UI — GNOME-style quick-flick / Windows-style tap-and-release behavior.

Under the hood the action mirrors the already-existing CLI path `AltTab --focusUsingLastFocusOrder=1` (see `src/logic/events/CliEvents.swift`). This PR just exposes the same behavior as a first-class configurable keybind so users don't need Raycast / Keyboard Maestro / skhd to glue it together.

Default binding is **empty** — users opt in by recording their own chord (e.g. `⌘+Esc`) to avoid shipping a surprise default that may clash with other apps.

## Where it lives

`Preferences → Controls → Additional controls…` — a new "Global shortcuts" section at the top of the sheet with a single recorder row: "Focus previous window".

Why the sheet and not the main tab: the sheet already hosts a multi-table layout (Additional controls, Miscellaneous), so adding a third table is ~7 lines; the main tab has a heavier sidebar+editor layout. Happy to move it if you'd prefer.

## Design choices

- **Scope `.global`, trigger phase `.down`** — fires from anywhere, immediately on press.
- **No UI shown** — the new action deliberately skips `focusTarget()` / `hideUi()`; it's a pure background focus-switch.
- **Mission Control guard** — copied from `focusSelectedWindow` to avoid focusing during Mission Control animations.
- **Honors `cursorFollowFocus == .always`** — consistent with the existing focus path.
- **Picks window with `lastFocusOrder == 1`** — same predicate the CLI flag uses.

## Changes

- `src/logic/Preferences.swift` — register `focusPreviousWindowShortcut` (default empty) + static accessor; add to `staticShortcutKeys`.
- `src/ui/App.swift` — new `focusPreviousWindow()` static method.
- `src/ui/settings-window/tabs/controls/ControlsTab.swift` — wire the action; extract `scopeForControlId()` helper so a new `globalStaticShortcutPreferences` list can opt shortcuts into `.global` scope alongside the existing `nextWindowShortcut*` prefix check.
- `src/ui/settings-window/tabs/controls/AdditionalControlsSheet.swift` — new "Global shortcuts" table with the recorder row.
- `resources/l10n/Localizable.strings` — two new English strings ("Focus previous window", "Global shortcuts"), alphabetically inserted. Other locales are expected to flow through POEditor.

## Test plan

- [ ] Build and launch the app.
- [ ] Open `Preferences → Controls → Additional controls…`. Verify a new "Global shortcuts" section shows at the top with a "Focus previous window" recorder.
- [ ] Record `⌘+Esc` (or any chord). Close Preferences.
- [ ] Open three apps; focus A → B → C. Press the shortcut → focus snaps to B. Press again → snaps to C. (Classic two-window toggle.)
- [ ] Verify the switcher UI never appears during the flick.
- [ ] Clear the recorder; pressing the shortcut becomes a no-op.
- [ ] While the switcher is open (e.g. via `⌥+⇥`), pressing the shortcut should not disrupt the session.
- [ ] Sanity: `AltTab --focusUsingLastFocusOrder=1` CLI still works unchanged.
- [ ] Edge case: only one window visible → no-op.
- [ ] Edge case: previous window on another Space → `window.focus()` handles space switching.

### DIFF
--- a/resources/l10n/Localizable.strings
+++ b/resources/l10n/Localizable.strings
@@ -212,6 +212,9 @@
 "Failed to import settings" = "Failed to import settings";
 
 /* No comment provided by engineer. */
+"Focus previous window" = "Focus previous window";
+
+/* No comment provided by engineer. */
 "Focus selected window" = "Focus selected window";
 
 /* No comment provided by engineer. */
@@ -222,6 +225,9 @@
 
 /* No comment provided by engineer. */
 "Gesture" = "Gesture";
+
+/* No comment provided by engineer. */
+"Global shortcuts" = "Global shortcuts";
 
 /* Menubar callout button */
 "Grant permission" = "Grant permission";
--- a/src/logic/Preferences.swift
+++ b/src/logic/Preferences.swift
@@ -8,6 +8,7 @@ class Preferences {
             "shortcutCount": "2",
             "nextWindowGesture": GesturePreference.disabled.indexAsString,
             "focusWindowShortcut": defaultShortcut(returnKeyEquivalent()),
+            "focusPreviousWindowShortcut": defaultShortcut(""),
             "previousWindowShortcut": defaultShortcut("⇧"),
             "cancelShortcut": defaultShortcut("⎋"),
             "lockSearchShortcut": defaultShortcut("Space"),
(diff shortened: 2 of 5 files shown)


lwouis:
Hi,

For now, there is no plan to introduce a global "Select Previous Window" shortcut. It was discussed a few times. For now, workflows involving using this seem a bit niche, and the cost of adding the complexities that go with a new global shortcuts don't seem worth it for the vast majority of users.

We may revisit the topic over time, as things evolve.

Thank you

fernandosmither:
fwiw @lwouis if it was "discussed a few times" it does not seem to be as niche as it may seem. As per the complexities, for an optional trigger (that's already how GNOME works out of the box) that requires +37 -5 diff, I'm not sure that adds too much complexity.

it's your project and I respect your vision, though. thank you for building it in the first place.

Issue #5492, opened 2026-04-04
Window ordering is not working as expected on 10.10


drauziooppenheimer:
**Describe the bug**

This is a regression from 10.9, it seems the window ordering is not being displayed correctly again, the switcher would constantly show the current window in the second position and command + tab has no effect unless pressed twice. 

* AltTab version: 10.10
* macOS version: 26.4

lwouis:
Hi,
Is it always like that or only sometimes with certain apps?
Thank you

drauziooppenheimer:
I've only been using Safari and Terminal so far, and is happening frequently with both. I have claude-code running on the terminal, so I wonder if it's triggering some events that are causing alt-tab to think that it was selected and is the focused window now, when that's not the case.

yutotakano:
This happens with VS Code and Microsoft Edge for me - I swap between these two often and it maybe happens once every 3 times (and when it happens, it gets stuck in that incorrect ordering unless I swap between the two again).

I swap from VS Code to Microsoft Edge, but VS Code still remains the first entry in alt tab, and Edge second. So a single invocation of alt-tab still keeps focus on Edge (no matter how many times I invoke it, until I manually change to VS Code and back). It almost feels like the code to update the recency order sometimes just doesn't run for whatever reason.

josephdpurcell:
AltTab version: 10.10.1
macOS version: 15.7

I had a similar issue but it has gone away... here is the story...

I installed 10.10.1 and the behavior I noticed was if I hit CMD+SHIFT+` it did not go to previous window as expected -- it did nothing.

I rolled back to 10.9.0 and it worked as I expected https://github.com/lwouis/alt-tab-macos/releases/tag/v10.9.0

Behavior with 10.9.0:
1. Tap CMD+` goes to previous window
2. Press and hold CMD and tap ` opens AltTab window switcher UI: tap ` to go next and SHIFT to go prev
3. Tap CMD+SHIFT+` goes to previous window
4. Press and hold CMD+SHIFT and tap ` opens AltTab window switcher UI: tap ` to go next and you can't go prev

Well, I forgot *exactly* the behavior of 10.10.1 so I deleted the AltTab app in the Applications folder, unzipped the 10.10.1, copied it in, and now I'm getting the behavior I expect.

[...]

Issue #5289, opened 2026-02-13
Space changed issue


sbmpost:
**Describe the bug**

Open "Notes" on workspace 1
Open "TextEdit" on workspace 2

1: Quickly Alt-Tab back and forth between "Notes" and "TextEdit" resulting in the workspace changing.
2: Notice how sometimes the app being switched to isn't actually being switched to.

**Your environment**

* AltTab version: latest (v9)
* macOS version: Sequoia 15.6
* Other relevant info: Single display

lwouis:
Thank you for opening this issue

I was able to reproduce the issue. It happens sometimes, not 100%. I'm also able to force the issue to happen by running this command, then opening AltTab and waiting:

```sh
sleep 3 && osascript -e 'tell application "Finder" to set w to (make new Finder window)' -e 'delay 1' -e 'tell application "Finder" to activate' -e 'tell application "Finder" to set index of w to 1'
```

Could you please let me know if [this build](https://github.com/user-attachments/files/25326756/AltTab.app.zip) fixes it for you? Note that it also contains all the recent work (see https://github.com/lwouis/alt-tab-macos/issues/5255#issuecomment-3904464703)

Thank you

sbmpost:
@lwouis 

Thanks for the quick response.

I tried with the build you provided (dated 2 Januari 2026). It doesn't fix the problem unfortunately. It is indeed only occurring if you alt-tab quickly (without even really looking at the thumbnails). Sometimes I need to quickly alternate between just two apps situated on different workspaces. This is when it happens.

lwouis:
Could you please share a video? I can't reproduce on my side. Maybe you go so fast that you "switch back" before the workspace animation is over, thus technically, the window is not yet focused, so technically we can't switch back?

sbmpost:
@lwouis 

Could not upload it in github, however this link should be valid for the next 24 hours:
https://jumpshare.com/s/nDmRmjZuZIeE46vIFBIg

When both apps are in the same workspace, things work as espected. So it seems related to the workspace changing.

lwouis:
I fear that it's a limitation of how macOS does the Space transition. The animation introduces a delay in how we receive events, I think.

The only way around the issue, I think, would be to add a blocking call to ask for the frontmost window, before showing the switcher. We currently removed this to optimize responsiveness.

I think the current tradeoff of a few hiccups, is worth the overall stelar responsiveness. People expect alt+tab to be very snappy.

Thank you

[...]

Issue #4772, opened 2025-09-17
Quickly switching between 2 recently opened apps no longer works


DarthGigi:
**Describe the bug**

This was working fine until one of the recent updates. I often need to switch between my 2 recent windows by pressing cmd + tab quickly for just a few ms. I've noticed that one of the recent versions broke this behavior and is messing up my muscle memory 😅 .

Maybe an option to automatically select (not focus) the previous active window in the switched could fix this?

**Screenshots / video**

https://github.com/user-attachments/assets/d02bd6df-81f6-4f3c-b68f-871bd3793a3c

**Steps to reproduce the bug**

**Your environment**

* AltTab version: Version 7.28.0
* macOS version: Tahoe 26.0 (25A354)
* Other relevant info: e.g. using external keyboard, multiple monitors, etc

wildlyinaccurate:
I have found the same bug in 7.28.0. The tab order is not correct, even though the "Order windows by" config is set to "Recently Focused First".

jim-alexander:
Same here, I downgraded and seems fine: [v7.27.0](https://github.com/lwouis/alt-tab-macos/releases/tag/v7.27.0)

olehmisar:
I have the same issue. It only happens when the cursor.com app open.

azhongzl:
I have the same issue. Another issue is the order is wrong sometimes. Downgraded to 7.27.

sergiuwaxmann:
+1 I have the same issue after updating AltTab to v7.28.0 (macOS Tahoe 26.0).

I solved the issue by downgrading AltTab to [v7.27.0](https://github.com/lwouis/alt-tab-macos/releases/tag/v7.27.0) (macOS Tahoe 26.0). Thank you @jim-alexander!

Looks like there are a few issues opened that mention the same problem:
https://github.com/lwouis/alt-tab-macos/issues/4775
https://github.com/lwouis/alt-tab-macos/issues/4776
https://github.com/lwouis/alt-tab-macos/issues/4777
https://github.com/lwouis/alt-tab-macos/issues/4778
https://github.com/lwouis/alt-tab-macos/issues/4783
https://github.com/lwouis/alt-tab-macos/issues/4787
https://github.com/lwouis/alt-tab-macos/issues/4788

Issue #5565, opened 2026-04-20
AltTab selects ChatGPT/Linear but does not bring the selected window to the front


RonenMars:
# AltTab selects ChatGPT/Linear but does not bring the selected window to the front

## Describe the bug

AltTab detects and selects the target app/window correctly, but after confirming the selection, the visible frontmost app does not change.

In my repro, AltTab highlights the ChatGPT app/window, logs `focusTarget()` for the correct PID/window, and macOS reports `AXApplicationActivated` for ChatGPT, but Ghostty remains the visible focused app on screen.

I expected ChatGPT or Linear to become the visible frontmost app/window after I selected them in AltTab. Instead, AltTab dismisses its UI, but Ghostty stays frontmost and the selected target window does not appear.

This does **not** look like a permissions issue or an app-discovery issue, because:
- AltTab lists the apps correctly
- AltTab targets the correct PID/window/title
- Accessibility and Screen Recording permissions are granted

Most relevant log lines:

```text
06:21:46.168 INFO App.swift:140 focusTarget() [main] (pid:755 com.openai.chat) (wid:183 title:ChatGPT)
06:21:46.297 INFO AccessibilityEvents.swift:45 handleEventApp() [main] AXApplicationActivated app:(pid:755 com.openai.chat)
```

Additional repro with ChatGPT:

```text
06:20:56.042 INFO App.swift:140 focusTarget() [main] (pid:755 com.openai.chat) (wid:183 title:ChatGPT)
06:20:56.117 INFO AccessibilityEvents.swift:45 handleEventApp() [main] AXApplicationActivated app:(pid:755 com.openai.chat)
```

Observed with Linear as well:

```text
06:20:36.708 INFO AccessibilityEvents.swift:45 handleEventApp() [main] AXApplicationActivated app:(pid:767 com.linear)
```

Permissions status:

```text
accessibility:granted screenRecording:granted
```

## Video

https://github.com/user-attachments/assets/81720ff7-a963-492a-8226-36460baf101c

_Funny side note:_ sometimes (not consistently) the AltTab Settings window itself doesn’t come to the front when selected in AltTab 😅

## Screenshots

<img width="920" height="951" alt="Image" src="https://github.com/user-attachments/assets/bec20cf3-1c06-4327-bca6-38153ecfc0d4" />
<img width="546" height="658" alt="Image" src="https://github.com/user-attachments/assets/2bc62542-80a5-4b68-9143-87b5c7fc88fe" />
[...]

PR #5567, opened 2026-04-21
fix: rapid switches leave stale lastFocusOrder, selecting wrong window

ghefraim:
## Problem

When rapidly alt-tabbing between two apps (e.g. iTerm ↔ Brave), AltTab's internal "recently-focused" order goes stale. Symptom: user alt-tabs to Brave, Brave becomes frontmost in macOS, but AltTab still thinks iTerm was most-recently focused. Next alt-tab then selects position 1 (Brave again) instead of returning to iTerm — the muscle-memory flip stops working.

This looks related to the threads around #4772 / #5289, but focuses on a different root cause in the focus-tracking pipeline.

## Root cause

Two independent places leave `lastFocusOrder` stale:

1. `AccessibilityEvents.focusedWindowChanged` is gated on `window.application.runningApplication.isActive`. The guard exists to prevent Photoshop-style "focus event without actually being active" cases, but `NSRunningApplication.isActive` is racy — during rapid switches the AX callback can run while macOS hasn't yet flipped `isActive` to `true` for the newly focused app. The legitimate event is dropped and `lastFocusOrder` never updates.

2. Even when the AX event does eventually arrive, there is a latency window between `window.focus()` and the async AX callback. A second alt-tab triggered inside that window sees the old order.

## Fix

- Replace the `isActive` guard with `Applications.frontmostPid == window.application.pid`. `frontmostPid` is set synchronously by AltTab itself in `applicationActivated`, so it's a reliable "this app is now the frontmost one" signal and still rejects Photoshop's spurious events.
- Call `Windows.updateLastFocusOrder` eagerly after `window.focus()` in `focusSelectedWindow`, so the alt-tab path does not depend on the async AX event round-trip at all.

## Test

Repro steps:
1. Open iTerm and Brave (or any two apps).
2. Alt-tab between them rapidly several times.

Before: occasionally the next alt-tab selects the app you just came from instead of the previous one.
After: order stays consistent; alt-tab always returns to the previously focused app.

Verified locally on macOS 26.x against master (v10.12.0).

### DIFF
--- a/src/logic/events/AccessibilityEvents.swift
+++ b/src/logic/events/AccessibilityEvents.swift
@@ -131,7 +131,11 @@ class AccessibilityEvents {
     private static func focusedWindowChanged(_ window: Window) {
         // photoshop will focus a window *after* you focus another app
         // we check that a focused window happens within an active app
-        guard window.application.runningApplication.isActive else { return }
+        // NSRunningApplication.isActive is racy: during rapid switches it can be stale/false when
+        // this callback runs, making us drop legitimate focus events and leaving lastFocusOrder stale.
+        // Applications.frontmostPid is set synchronously when AltTab processes the activation event,
+        // so it's a reliable signal that this window's app is the currently frontmost app.
+        guard Applications.frontmostPid == window.application.pid else { return }
         // if the window is shown by alt-tab, we mark it as focused for this app
         // this avoids issues with dialogs, quicklook, etc (see scenarios from #1044 and #2003)
         window.application.focusedWindow = window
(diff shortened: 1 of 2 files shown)


jwatt:
Thank you! I've been encountering this issue very frequently. (I'm using 10.12.0 on an older, slower MBP.)

I've no idea about the validity of the changes in this patch, but once I managed to get the master branch building I tried with/without the patch. The patch does indeed appear to fix the issue for me.

jwatt:
Maybe fixes #5492, #5565, #5580?

lwouis:
Thank you

pacesmarek:
@lwouis Hi!

So FIX will be included in next version/update? I have currently 10.12.0.
I need this FIX or I'll eat my keyboard :D

**When can we expect the new version?** 

Thanks!

lwouis:
@pacesmarek hopefully this week, with the release of AltTab Pro 👍

pacesmarek:
> @pacesmarek hopefully this week, with the release of AltTab Pro 👍

After udate. Works great! Thanks @lwouis

PR #5598, opened 2026-05-04
perf: prioritize screenshot capture for viewport-visible windows

defer2xn:
## Problem

When the switcher panel appears, `refreshThumbnailsAsync` submits screenshot requests for **all** windows at once. With 20+ windows open, this creates a burst of ScreenCaptureKit/CGSHWCaptureWindowList calls that compete for system resources, delaying the thumbnails users actually need to see.

## Solution

Split thumbnail capture into two phases in `buildUiAndShowPanel()`:

1. **Immediately** capture windows whose tiles are within the scroll viewport
2. **After 150ms**, capture the remaining off-screen windows (guarded by `appIsBeingUsed`)

This keeps the screenshot queue peak bounded by the number of visible tiles (~6-10) rather than total window count, so the thumbnails users see first arrive faster.

## Details

- Uses `documentVisibleRect` (document coordinate space) for correct viewport detection, even after scroll offset changes
- Windows with `.zero` frame (filtered by `shouldDisplay`) are skipped entirely — no wasted capture requests
- Deferred batch is cancelled if the user dismisses the panel before 150ms
- Both capture paths (ScreenCaptureKit on macOS 14+ and private API fallback) are covered since the split happens before `refreshThumbnailsAsync`

## Testing

- Verified with 30-window setup: first-screen thumbnails appear noticeably faster
- No behavior change for users with few windows (all fit in viewport → single batch)
- Zero-window and single-window edge cases handled via existing guards in `refreshThumbnailsAsync`

### DIFF
--- a/src/ui/App.swift
+++ b/src/ui/App.swift
@@ -344,7 +344,30 @@ class App: AppCenterApplication {
             TilesView.enableSearchEditing()
         }
         KeyRepeatTimer.startRepeatingKeyNextWindow()
-        Windows.refreshThumbnailsAsync(Windows.list, .refreshOnlyThumbnailsAfterShowUi)
+        refreshThumbnailsViewportFirst()
+    }
+
+    private static func refreshThumbnailsViewportFirst() {
+        let visibleBounds = TilesView.scrollView.documentVisibleRect
+        var visibleWindows = [Window]()
+        var deferredWindows = [Window]()
+        for (index, window) in Windows.list.enumerated() {
+            guard index < TilesView.recycledViews.count else { break }
+            let tileFrame = TilesView.recycledViews[index].frame
+            if tileFrame == .zero {
+                continue
+            } else if visibleBounds.intersects(tileFrame) {
+                visibleWindows.append(window)
+            } else {
+                deferredWindows.append(window)
+            }
+        }
+        Windows.refreshThumbnailsAsync(visibleWindows, .refreshOnlyThumbnailsAfterShowUi)
+        guard !deferredWindows.isEmpty else { return }
+        DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(150)) {
+            guard appIsBeingUsed else { return }
+            Windows.refreshThumbnailsAsync(deferredWindows, .refreshOnlyThumbnailsAfterShowUi)
+        }
     }
 
     static func checkIfShortcutsShouldBeDisabled(_ activeWindow: Window?, _ activeApp: Application?) {
--- a/src/ui/App.swift
+++ b/src/ui/App.swift
@@ -364,9 +364,12 @@ class App: AppCenterApplication {
         }
         Windows.refreshThumbnailsAsync(visibleWindows, .refreshOnlyThumbnailsAfterShowUi)
         guard !deferredWindows.isEmpty else { return }
-        DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(150)) {
-            guard appIsBeingUsed else { return }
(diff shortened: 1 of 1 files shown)


lwouis:
Hi,
Thank you for sharing this MR
I think the windows are already screenshot from top to bottom. I mean that we make the screenshot API calls in that order. Thus, there should already be prioritisation from macOS to provide screenshots in this order.
Have you tested the MR, and found substantial improvement? If yes, any numbers to share?
Thank you

defer2xn:
Hi @lwouis,

Thanks for the quick review!

You're right that the screenshot API calls are already ordered top-to-bottom. However, what I observed is that when all requests are submitted in a single burst (e.g. 20-30 windows), macOS doesn't guarantee they complete in submission order — ScreenCaptureKit processes them somewhat concurrently, and the system resource contention means even the first few screenshots take longer to come back than they would in a smaller batch.

The key difference here isn't about ordering, but about **batch size**. By only submitting ~6-10 visible-viewport requests initially, the system has fewer concurrent capture tasks competing for resources, so those first results return faster.

In my testing with 30 windows open, I measured the time from panel appearance to the first visible thumbnails being rendered:
- **Before**: ~250-350ms for first-screen thumbnails to appear
- **After**: ~120-180ms for the same thumbnails

For users with fewer windows (all fitting in the viewport), there's no behavior change since everything goes in the first batch anyway.

That said, I acknowledge this improvement is most noticeable on setups with many windows. If you'd like, I can add more detailed profiling numbers or simplify the approach further.

lwouis:
Thank you for your help. This makes sense 👍 

It would be nice to keep the split into 2 batches: visible vs out-of-sight thumbnails, while not hardcoding the 150ms delay. The issue with timing-based can go both ways:
* Machine is powerful: we delay the second batch which could be done faster
* Machine is slow: we start batch 2 while batch 1 is still processing, reducing the gains
It would be nice to launch batch 2 once we observe batch 1 is done. However, if we wait for 100% done, we miss out on concurrency. Maybe some heuristic here would help like when less than 8 screenshots remain, we start batch 2, since we cap the concurrency at 8 in the screenshot queue?

Thank you

defer2xn:
Hi @lwouis,

Great point about the timing-based approach. I've replaced the 150ms delay with a queue-based trigger.

Instead of `DispatchQueue.main.asyncAfter`, I now add a lightweight trigger operation to `screenshotsQueue` right after batch 1's operations. Since the queue's `maxConcurrentOperationCount` is 8, this trigger naturally fires when the queue has a free slot — i.e., when batch 1 has fewer than 8 operations remaining. At that point it dispatches batch 2 to main thread for submission.

This means:
- **Fast machine**: batch 1 operations complete quickly → trigger fires sooner → no wasted idle time
- **Slow machine**: trigger waits until the queue actually has capacity → batch 1 isn't starved
- **Few windows** (all visible): no deferred batch, same as before

For the private API path (macOS 15 and older, where each capture is a blocking operation), this directly maps to "< 8 screenshots remaining" since each operation = one capture. For the ScreenCaptureKit path (async `captureSampleBuffer`), the queue operations don't map 1:1 to in-flight captures, so the trigger fires a bit earlier — but SCK manages its own concurrency internally, so the impact is minimal there.

The change is just a few lines in `refreshThumbnailsViewportFirst()` — swapping the timer for a queue operation.

lwouis:
Hi,

I've found a more robust approach, I think. Flagging the first screenshots with `.high` priority, versus the others with `.normal` priority. This seems like it covers every use case that we describe, and seems quite robust.

I'll release this improvement in the next release 👍 

Thank you

PR #5599, opened 2026-05-04
perf: defer content rendering for off-screen tiles during layout

defer2xn:
## Problem

`layoutTileViews()` calls `updateRecycledCellWithNewContent()` for every window regardless of scroll visibility. Each call involves NSAttributedString construction, CALayer.contents assignment, accessibility label updates, and font metric calculations — roughly ~400µs per tile on the main thread. With 30 windows, that's ~12ms of layout work, most of it for tiles the user can't even see.

## Solution

Split tile updates into two paths based on viewport visibility:

- **Visible tiles**: full `updateRecycledCellWithNewContent()` as before
- **Off-screen tiles**: lightweight `updateFrameSizeOnly()` (~7µs) for frame calculation, with content deferred via `deferContentUpdate()`

Deferred tiles get their content populated when they actually need to be shown:

- **Scroll end**: `ScrollView.updateVisibleTiles()` flushes newly-visible deferred tiles
- **Keyboard navigation**: `updatePendingContent()` is called on the focused tile before `scrollToVisible`, so the selected tile always has correct content
- **Next layout pass**: `refreshUi()` re-evaluates all tiles, clearing stale deferred state

## Additional fixes

- Scoped scroll notification observers from `object: nil` to `object: self` — previously any NSScrollView in the process (e.g. Settings window) could trigger the handler

## Performance impact (30 windows, 8 visible)

| Metric | Before | After |
|---|---|---|
| Full content updates per layout | 30 | 8 |
| Layout main thread time | ~12ms | ~3.7ms |
| Improvement | | **~69%** |

## Testing

- Edge cases verified: 0 windows, 1 window, windows > recycledViews.count
- `visibleRect.isEmpty` (first layout before frame is set) falls back to full update — no blank tiles
- Search mode triggers full `refreshUi()`, correctly re-evaluating all tiles
- `deferContentUpdate` sets mouseUp/mouseMoved callbacks immediately to prevent force-unwrap crash on user interaction before content is populated

### DIFF
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -342,6 +342,7 @@ class Windows {
         guard let index else { return }
         TilesView.highlight(index)
         let focusedView = TilesView.recycledViews[index]
+        focusedView.updatePendingContent()
         TilesView.scrollView.contentView.scrollToVisible(focusedView.frame)
         voiceOverWindow(index)
     }
--- a/src/ui/main-window/TileView.swift
+++ b/src/ui/main-window/TileView.swift
@@ -55,13 +55,52 @@ class TileView: FlippedView {
         return CGRect(origin: .zero, size: frame.size)
     }
 
+    var needsContentUpdate = false
+    private var pendingWindow: Window?
+    private var pendingIndex: Int = 0
+
+    func updateFrameSizeOnly(_ element: Window, _ newHeight: CGFloat) {
+        window_ = element
+        if !thumbnail.isHidden {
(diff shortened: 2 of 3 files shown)


lwouis:
Hi,
Thank you for sharing this MR
Moving to a virtual list approach complexifies the app. Also it has its own downside, like maybe suddently scrolling is not smooth now.
I would consider it if we were struggling, however, the number you show are quite small. 12ms to render 30 windows is quite reasonable.
I'll close the MR for now
Thank you

PR #5600, opened 2026-05-05
perf: viewport-first rendering and deferred tile updates

defer2xn:
## Summary
- **Viewport-first thumbnail refresh**: prioritize screenshot capture for visible tiles, defer off-screen ones by 150ms
- **Lazy content update during layout**: only render tile content (thumbnail/icon/label/search highlight) for visible tiles; off-screen tiles get frame-only sizing with deferred content update on scroll or keyboard navigation
- **Skip content rendering in dry-run layout**: `dryRunLayoutTileViews` (auto-size mode) now uses lightweight `updateFrameSizeOnly` instead of full `updateRecycledCellWithNewContent`, avoiding up to 3 rounds of wasted rendering
- **Flush deferred tiles after keyboard scroll**: `scrollToVisible` is programmatic and doesn't fire `didEndLiveScrollNotification`, so explicitly async-flush deferred tiles that entered the viewport
- **Dictionary-based SCWindow cache lookup**: replace O(n·m) linear scan with O(n+m) dictionary lookup in screenshot capture pipeline

## Test plan
- [ ] Auto-size mode: trigger Alt-Tab with 15+ windows, verify panel shows correct size and all tiles render correctly
- [ ] Fixed size mode: verify no regression (optimizations are no-op in this path)
- [ ] Keyboard navigation with scrolling: open 10+ windows (thumbnails style), press ↓ to scroll down — tiles scrolling into view should render correctly without blank/stale content
- [ ] Search + navigation: activate search, type query, navigate with arrow keys — search highlights should display correctly on all tiles
- [ ] All appearance styles (thumbnails/titles/appIcons): verify layout and content rendering
- [ ] Screenshot capture: with 20+ windows, all thumbnails should load progressively

### DIFF
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -344,6 +344,9 @@ class Windows {
         let focusedView = TilesView.recycledViews[index]
         focusedView.updatePendingContent()
         TilesView.scrollView.contentView.scrollToVisible(focusedView.frame)
+        DispatchQueue.main.async {
+            TilesView.scrollView.updateVisibleTiles()
+        }
         voiceOverWindow(index)
     }
 
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -342,6 +342,7 @@ class Windows {
         guard let index else { return }
         TilesView.highlight(index)
         let focusedView = TilesView.recycledViews[index]
+        focusedView.updatePendingContent()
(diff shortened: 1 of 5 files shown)


lwouis:
The thumbnails defered loading are discussed in the other MR. The other optimizations in this MR, making the list a virtual list, have downsides, so I'd rather keep the current design 👍

PR #5602, opened 2026-05-05
Claude's fix for #5492

### LINKED ISSUE #5492: Window ordering is not working as expected on 10.10

singalen:
This seems to have fixed #5492.

Tested it for a while, the issue seems to stop repro.

### DIFF
--- a/src/logic/Windows.swift
+++ b/src/logic/Windows.swift
@@ -463,10 +463,14 @@ class Windows {
     }
 
     static func updateLastFocusOrder(_ focusedWindow: Window) -> [Window]? {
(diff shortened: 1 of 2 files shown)


singalen:
...and then I noticed #5567. It looks better, please merge it!

lwouis:
Thank you

PR #5702, opened 2026-06-02
add greek translation

dimitrisscript:
Hello! I didn't get a response [over here](https://github.com/lwouis/alt-tab-macos/discussions/5533#discussioncomment-17059864) but I still went ahead and opened a PR that adds the Greek language to the app.

@lwouis fyi :)


lwouis:
Hi,

With v11 and AltTab Pro, I've decided to have a rule for language support. In the past, I would just add any language contributed by anyone. That led to 2 issues:
* People would come and go, and often leave me with the burden of localizing new strings/changes. Making adding content difficult
* Languages would pill up, increasing disk space taken by the app, which I'm very careful to keep down, to respect the users resources.

Thus I've decided to support the top 20 languages by user count. Greek is not in the top 20, unfortunately.

Thank you

PR #5711, opened 2026-06-04
Fix: trackpad gesture tap silently dying after 1-2 days of uptime

MR-MANDELBROT:
## Problem

After 1-2 days of continuous uptime **without sleep/wake**, the 4-finger swipe gesture silently stops working. Keyboard shortcut (Alt+Tab) continues to work. An app restart fixes it. Reproduced on macOS 26.1.

## Root cause

The CGEventTap's underlying `CFMachPort` can be **invalidated by macOS** after extended uptime. In this state:

- `tapIsEnabled()` may still return `true` → `reEnableTapIfNeeded()` exits early without doing anything
- `tapDisabledByTimeout` callback never fires → the existing handler on line 59 never runs
- The tap is silently dead with no recovery path

This is distinct from the sleep/wake case (already handled). The mach port itself becomes invalid, not just the tap being disabled.

## Fix (3 changes)

**1. `CFMachPortIsValid` check in `reEnableTapIfNeeded()`**  
If the port is invalid, recreate the tap entirely instead of just re-enabling it.

**2. `CFMachPortSetInvalidationCallBack` in `observe_()`**  
Fires immediately when macOS drops the mach port — triggers full tap recreation without waiting for the next periodic check.

**3. 5-minute periodic watchdog in `SleepWakeEvents`**  
Safety net that catches any case not covered by the two mechanisms above (e.g. macOS 26 silently stopping event delivery without actually invalidating the port).

## Testing

Hard to reproduce in a short test cycle (requires 1-2 days uptime), but the logic is straightforward:
- `CFMachPortIsValid` returns `false` → `observe_()` is called → fresh tap created
- `CFMachPortSetInvalidationCallBack` fires → `observe_()` scheduled on main thread
- Timer fires every 5 min → `reEnableTapIfNeeded()` called on all taps

### DIFF
--- a/src/events/SleepWakeEvents.swift
+++ b/src/events/SleepWakeEvents.swift
@@ -3,6 +3,9 @@ import Cocoa
 class SleepWakeEvents {
     static func observe() {
         NSWorkspace.shared.notificationCenter.addObserver(self, selector: #selector(handleWake), name: NSWorkspace.didWakeNotification, object: nil)
+        // Periodic watchdog: catches cases where the tap silently stops working without a
+        // sleep/wake event (e.g. CFMachPort invalidated after 1-2 days of continuous uptime).
+        Timer.scheduledTimer(withTimeInterval: 5 * 60, repeats: true) { _ in reEnableAllTaps() }
     }
 
     @objc private static func handleWake(_ notification: Notification) {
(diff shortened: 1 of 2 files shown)


lwouis:
Hi,

Could you please help me with making a case for the root cause?

> The CGEventTap's underlying CFMachPort can be invalidated by macOS after extended uptime. In this state:

This is not documented by Apple, to my knowledge. Could you please share evidence that this can happen?

Thank you

PR #5748, opened 2026-06-12
Fix thumbnails when Stage Manager is enabled

bnjdpn:
## Summary
- Fall back to the existing SkyLight private capture path when Stage Manager is enabled.
- Keep ScreenCaptureKit for macOS 14+ when Stage Manager is off, while preserving the existing macOS 15 private API fallback.
- Add unit tests for capture backend selection and Stage Manager preference detection.

## Test plan
- `xcodebuild test -project alt-tab-macos.xcodeproj -scheme Test -configuration Release -derivedDataPath DerivedData -only-testing:unit-tests/WindowCaptureMethodResolverTests CODE_SIGNING_ALLOWED=NO CODE_SIGN_IDENTITY=-`
- `xcodebuild -project alt-tab-macos.xcodeproj -scheme Debug -configuration Debug -derivedDataPath DerivedData CODE_SIGNING_ALLOWED=NO CODE_SIGN_IDENTITY=- build`

## Notes
- I could launch the local debug app with Stage Manager enabled, but macOS required a fresh Screen Recording permission grant for the debug bundle before thumbnails could be visually inspected.


lwouis:
Hi,

Thank you for sharing this PR

I tested it, and it doesn't work for me. I'm on macOS 26.5. When I enable Stage Manager, the thumbnails break.
I think I remember that private or public APIs got "fixed" at some point, like macOS 26 release perhaps. I think they "optimized" the rendering so they stopped storing the full-size window bitmaps when the window goes small into Stage Manager. Thus there is no way neither for us nor for macOS itself to get the full-size images.

Perhaps it works for you because you're not on macOS 26?

Thank you

PR #5782, opened 2026-06-18
Two-level window switcher

decodism:
Add the necessary settings to create a two-level window switcher: selecting the app and then the window

- Restore state when switching shortcut
- Use previous selected app as active app when switching shortcut
- Select last focused window when switching shortcut
- Show titles for app icons
- After keys are released: Open
- Hide more windows icon

This is more of a proof of concept. A better user experience might be to display both apps and windows, either in two panels or in a grid.

Example of settings for selecting the app and then the window:

```
<dict>
	<key>appearanceStyleOverride</key>
	<string>1</string>
    <key>shortcutStyleOverride</key>
	<string>3</string>
    <key>showAppsOrWindows</key>
	<string>0</string>
	<key>restoreState</key>
	<string>true</string>
    <key>usePreviousSelectedApp2</key>
	<string>true</string>
	<key>selectLastFocusedWindow2</key>
	<string>true</string>
</dict>
```

Example of settings for selecting the main window and then another window:

```
<dict>
    <key>showAppsOrWindows</key>
	<string>0</string>
	<key>restoreState</key>
	<string>true</string>
	<key>usePreviousSelectedApp2</key>
	<string>true</string>
</dict>
```

### DIFF
--- a/resources/l10n/en.lproj/Localizable.strings
+++ b/resources/l10n/en.lproj/Localizable.strings
@@ -236,6 +236,7 @@
 "Show separate window for each tab" = "Show separate window for each tab";
 "Show single window for all tabs" = "Show single window for all tabs";
 "Show titles" = "Show titles";
+"Show titles for app icons" = "Show titles for app icons";
 "Show windows from applications" = "Show windows from applications";
 "Show windows from screens" = "Show windows from screens";
 "Show windows from Spaces" = "Show windows from Spaces";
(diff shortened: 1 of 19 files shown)


lwouis:
Hi @decodism,

I think you're tackling #337 here?

> This is more of a proof of concept

What is your goal with this PR? Exploring, opening the conversation?

Thank hyou

decodism:
Hello,

I'm not very interested in screen/space selection. My goal was to see if it was possible to easily implement app selection. It's mostly the case, but it's still a bit hacky and not very user-friendly to configure. I don't plan to develop this further, but maybe it can inspire someone. The implementations I’ve considered, roughly from least to most ambitious:

- New option for app grouping setting: show all windows per app
- Settings to change the behavior of shortcuts when switching between them (this PR). Maybe more work is needed on the configuration
- New ways to quickly apply temporary filters
- A new appearance style that would display applications on one axis and their windows on another, for example

PR #5787, opened 2026-06-20
fix: ignore Dock when inspecting window

decodism:
Dock overlays a full-screen window that prevents window inspection.


lwouis:
Thank you for sharing! 👍 
I've cherry picked the commit. It will be in the next release

PR #5821, opened 2026-07-02
feat: add Bengali (bn) localization

adnan29979:
I am a native Bengali speaker and I translated the UI into Bengali (বাংলা). I wired it up the same way as the other languages: a new bn.lproj folder under resources/l10n, a bengali case in LanguagePreference, and the matching entries in the Xcode project.

I built and ran it locally, switched the app to Bengali to make sure nothing overflows or gets cut off. I left product names like Pro and Lifetime, and the macOS "Spaces" feature, untranslated to match the other locales, and kept the %@ / %d format specifiers intact.


lwouis:
Hi,
Please see https://github.com/lwouis/alt-tab-macos/pull/5702#issuecomment-4605484341
Thank you

Issue #5786, opened 2026-06-19
[Bug] "Capture windows in background" destabilizes WindowServer → unexpected full session logout (macOS 26 / Apple Silicon, multi-display)


pavel-spacil:
**Describe the bug**

With **"capture window contents in background"** enabled, AltTab continuously screen-captures windows (~once per second) using the legacy display-stream API. On macOS 26.5.1 (Apple Silicon, multiple external displays) this appears to progressively destabilize `WindowServer` until it **exits**, which triggers a **full macOS session logout** — all apps are killed and the machine drops to the login window. It happened ~1–2×/day on 4 separate days.

The macOS unified log shows AltTab's capture path firing at ~1 Hz right up to the crash:

```
WindowServer  Creating sharing context <UUID>
WindowServer  Screenshots via streams are inefficient and should be avoided. Please use the SLSScreenshot SPI.
```

(~36 such contexts + 18 of these warnings in the ~2 min before one crash). This builds into:

```
WindowServer  _CGXPackagesSetWindowConstraints: Invalid window   (flooding, 900+ in 2 min)
WindowServer  Cursor disabled: failed set_cursor_surface
WindowServer  pid NNN failed to act on a ping it dequeued before timing out.
```

culminating in:

```
loginwindow  -[LoginApp windowServerExited] | ERROR | Window Server exited, closing down the session immediately
loginwindow  -[SessionLogoutManager startDirectLogout:reason:] | logoutType: Logout, directLogoutReason:WindowServerExited
launchd      Successfully spawned WindowServer[...] because ipc (mach)
```

There is **no WindowServer `.ips` crash report** — it exits rather than crashing.

**Mitigation found:** disabling **"capture window contents in background"** (`captureWindowsInBackground = false`) immediately stops the capture churn (0 stream events afterward). No logouts since.

**Steps to reproduce the bug** (intermittent — not reliably on-demand)

1. Enable "capture window contents in background".
2. Run on macOS 26 / Apple Silicon with multiple external displays and many open windows.
3. Use normally for an extended period; WindowServer CPU climbs and eventually it exits, forcing a session logout.

**Your environment**

* AltTab version: **11.2.0**
* macOS version: **26.5.1 (25F80)**
* Other relevant info: Apple **M4 Pro**; **3 displays** (built-in Liquid Retina XDR + 2× external 1920×1200 via dock), all at native resolution; "capture windows in background" was **on**.

[...]

PR #5845, opened 2026-07-09
fix: use safer screenshot API for regular thumbnails

Ynng:
## The problem

AltTab takes screenshots of open windows to display its thumbnails. On macOS 26 it currently uses `captureSampleBuffer` for those screenshots.

Apple implements each `captureSampleBuffer` call by creating and destroying a short-lived capture stream. On some Macs, repeatedly doing this causes WindowServer's memory use to grow until WindowServer is killed. When that happens, the screen goes black and the user is logged out. This is the problem reported in #5786 and earlier memory reports such as #4255.

I investigated one affected Mac where WindowServer reached 269,176 MB of compressed memory before macOS killed it with `no paging space`. The diagnostic showed the same display-stream and window-capture functions discussed in #5786, with capture work attributed to AltTab.

## Why this was not fixed by simply changing APIs

macOS 26 has another API named `captureScreenshot`. It does not create a short-lived capture stream, so it avoids this WindowServer churn.

However, it has two important limitations:

- it fails to capture fullscreen windows
- it returns a copied image, which is noticeably slower for AltTab's full-resolution Preview feature

The maintainer already tried changing to this API in #5786, but could not ship it because fullscreen thumbnails stopped working.

## What this PR changes

This PR uses both APIs, depending on the window:

1. For a normal window with Preview disabled, use `captureScreenshot`. This is the common case and avoids creating a capture stream.
2. For a fullscreen window, keep using `captureSampleBuffer` because `captureScreenshot` cannot capture it.
3. When Preview is enabled, keep using `captureSampleBuffer` because Preview needs the faster full-resolution path.
4. If `captureScreenshot` unexpectedly fails for another window, retry once with `captureSampleBuffer` so the thumbnail is not left blank.

For users with the default Preview setting disabled, ordinary thumbnail updates and background captures will therefore stop creating a new capture stream for every screenshot.

## What this does not solve

This is a mitigation, not a complete fix for Apple's WindowServer bug. Fullscreen windows and users with Preview enabled still use `captureSampleBuffer`, so those captures can still exercise the problematic macOS code. The goal is to remove the large majority of stream creation without breaking fullscreen windows or making Preview substantially slower.

## Validation

- the app built successfully with Xcode 26.0.1 on a macOS GitHub runner
- the resulting app was installed, signed locally, and launched on macOS 26
- the new `captureScreenshot` call type-checks against the macOS 26.2 SDK
- the changed Swift source parses successfully and the diff passes `git diff --check`

The PR is still a draft because the following behavior should be checked manually before merging:

- normal, minimized, and other-Space window thumbnails
- fullscreen window thumbnails
- Preview mode
- background thumbnail updates

### DIFF
--- a/src/events/WindowCaptureEvents.swift
+++ b/src/events/WindowCaptureEvents.swift
@@ -12,6 +12,8 @@ class WindowCaptureScreenshots {
         let window: Window
         let size: CGSize
         let scaleFactor: CGFloat
+        let isFullscreen: Bool
+        let usesPreview: Bool
     }
 
     static func oneTimeScreenshots(_ windowsToScreenshot: [Window], _ source: RefreshCausedBy, prioritizedIds: Set<CGWindowID>? = nil) {
@@ -22,6 +24,11 @@ class WindowCaptureScreenshots {
         // Trade-off: size is fixed at call time, so a window resized between snapshot and capture will be captured
         // at the old size. Acceptable because the next refresh will re-snapshot.
         var requests = [CGWindowID: CaptureRequest]()
+        // Preview renders the selected window at full resolution. Keep the zero-copy sample-buffer path for that
+        // case; ordinary thumbnails can use macOS 26's stable one-shot screenshot API instead.
+        let usesPreview = (0...Preferences.maxShortcutCount).contains {
+            Preferences.effectivePreviewSelectedWindow($0)
+        }
         for window in windowsToScreenshot {
             guard let wid = window.cgWindowId, let size = window.size else { continue }
             let scaleFactor: CGFloat
@@ -30,7 +37,8 @@ class WindowCaptureScreenshots {
             } else {
                 scaleFactor = NSScreen.preferred.backingScaleFactor
             }
-            requests[wid] = CaptureRequest(window: window, size: size, scaleFactor: scaleFactor)
+            requests[wid] = CaptureRequest(window: window, size: size, scaleFactor: scaleFactor,
+                isFullscreen: window.isFullscreen, usesPreview: usesPreview)
         }
         guard !requests.isEmpty else { return }
         let prioritized = prioritizedIds ?? []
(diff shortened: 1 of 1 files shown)


mschmidtkorth:
This matches what I traced on AltTab 11.3.1 / macOS 26.5: besides the WindowServer memory growth, captureSampleBuffer's per-call stream also churns the screen-capture privacy attribution, which pegs systemstatusd on CPU.

The replayd log shows each thumbnail capture flipping attribution on/off ~8×/secs:

22:17:22.842  SCMediaAttributionManager updateScreenCaptureDidStart  hasStarted=1
22:17:22.957  SCMediaAttributionManager updateScreenCaptureDidStart  hasStarted=0   (~115ms later)
22:17:23.104  SCMediaAttributionManager updateScreenCaptureDidStart  hasStarted=1
22:17:23.194  SCMediaAttributionManager updateScreenCaptureDidStart  hasStarted=0
...  +[SCScreenShotSession captureSLContentStreamScreenshot]  "individual window"  isFullDisplayShare=0

Each hasStarted 1→0 is an XPC attribution update to systemstatusd.. It also made screenshot apps or MacOS native screenshots freeze my UI. Quitting AltTab drops replayd and systemstatusd to 0.0% instantly.

lwouis:
I'll ship very similar work in the next release. I've re-run some analysis on the issue, adding all the perspective, and benchmarking again on my machine to ground in reality. Here's a recap as an archive for posterity:

## Summary of the investigation (archive)

I benchmarked both APIs on macOS 26.5.1 (Apple Silicon), forcing each path in a modified build of #5845. Payload: 29 windows covering all states (2 fullscreen on background Spaces, 3 minimized, 3 partially offscreen, tabbed, all sizes), 10 switcher cycles per run, ~355 captures/run, 2 runs per config.

### Insights

1. `captureSampleBuffer` makes WindowServer create sharing contexts on every capture (the leak vector behind the logouts). `captureScreenshot` produces exactly zero, in every run.
2. `captureScreenshot` is ~10-15% *faster* at thumbnail sizes. Even forced to full-res (≤5.2 MP, single display) it stayed slightly faster; the known copy penalty only shows on >5 MP multi-display setups. Keeping Preview on `captureSampleBuffer` is the conservative call, not a measured necessity here.
3. `captureScreenshot` fails only for fullscreen windows on a non-frontmost Space, deterministically, always `SCStreamError -3811`. Minimized, other-Space, offscreen, tabbed, frontmost-fullscreen: all fine. Routing by `isFullscreen` covers the whole failure surface, so no API fallback is needed.
4. Switching APIs does not help systemstatusd: both flip replayd's capture attribution ~4×/capture (identical counts), systemstatusd CPU within 2%. That overhead is per capture, not per stream; only capturing less would reduce it.
5. WindowServer burst CPU is identical for both; the damage is the leak, not CPU.

### Raw data

Per ~22s run, mean of 2 (latencies include AltTab's 8-way capture queue):

| | sampleBuffer thumb | screenshot thumb | sampleBuffer full-res | screenshot full-res |
|---|---|---|---|---|
| latency mean / p90 (ms) | 706 / 1123 | 644 / 1035 | 709 / 1119 | 664 / 1083 |
| WS sharing contexts / warnings | ~355 / ~355 | 0 / 0 | ~354 / ~354 | 0 / 0 |
| replayd attribution flips | ~1403 | ~1409 | ~1415 | ~1420 |
| replayd / systemstatusd CPU (s) | 1.97 / 4.83 | 1.83 / 4.74 | 2.01 / 4.87 | 1.84 / 4.77 |
| failed captures | 0 | 25 (only the 2 fullscreen windows, always -3811) | 0 | 25 (same) |

WindowServer CPU and all RSS deltas: within ambient noise (a 20-cycle run can't reproduce the leak; the context count is the proxy).

### Outcome

macOS 26+: `captureScreenshot` everywhere, except fullscreen windows and full-res Preview which stay on `captureSampleBuffer`. No fallback; unexpected failures are logged. Details in `src/events/WindowCaptureEventsSpecs.md`.

Thanks @Ynng for the PR, @pavel-spacil for the early benchmarks, @mschmidtkorth for the attribution trace 👍

I'll close this PR. It's contents in spirit will be in the next release 👍

PR #5948, opened 2026-08-13
fix: compact thumbnail rows around previews

bnjdpn:
## Summary

- size thumbnail tiles from the preview content that is actually rendered, while keeping the existing maximum height
- use the tallest tile in each row so mixed-aspect-ratio previews remain aligned
- cover the new sizing behavior with unit tests and update the appearance specification

## Root cause

For very wide windows, the thumbnail becomes height-limited by its aspect ratio. The tile and row layout still reserved the theoretical maximum thumbnail height, which left a visible empty area below the preview.

## Impact

Thumbnail rows now compact around ultrawide previews without stretching or cropping the image. The existing maximum height remains in place, and the titles and app-icons appearance styles are unchanged.

## Visual comparison

Both screenshots were captured on the same display with the same four windows and the same selected tile. The before image uses the PR's base commit; the after image uses this change.

### Before

Thumbnail layout before the fix

### After

Thumbnail layout after the fix

In this mixed-aspect-ratio row, the panel height decreases from 276 px to 252 px while all tiles remain aligned.

## Validation

- `AppearanceTests`: 6/6 passing
- unsigned Debug build: succeeded
- full unit-test suite: attempted; it reaches the updated appearance tests successfully, then crashes in the unrelated existing `CustomRecorderControlTests.testCombinedModifiersMatchEqualToItself` during shortcut initialization


lwouis:
Hi,

Thank you for sharing this work.

As it is, I'm afraid I can't merge it. It makes the switcher more compact, which seems great. On the other hand, it means rows now have varying heights. That downside is too bit, in my judgment, for the compaction benefit.

Furthermore, when I look at the screenshots, the thumbnails content seems hard to distinguish. It seems that maybe the real topic to address is the ultrawide screen. All issues come from the fact that the ultrawide screen has such an extreme ratio.

Perhaps we could detect ultrawide screens, and adjust the whole layout for them? I see issues with this though. For example, someone buying an ultrawide screen so they can always have 2 windows side-by-side confortably. For such users, the window ratios may actually be tall.

I'll think if I can improve the layout somehow. What I'd like to avoid though, is interface instability. That is: depending on the contents, things move around.

Thank you

PR #5964, opened 2026-08-19
fix: stop re-enabling macOS keyboard shortcuts that you disabled

aviseth:
Whether a native symbolic hotkey (⌘⇥, ⌘⇧⇥, ⌘\`) is enabled is global state the user sets in System Settings, and as the comment on `CGSSetSymbolicHotKeyEnabled` says, it sticks around after we quit.

We treat every hotkey we don't need as one to switch on. `resolve` returns `enable` as the complement of `disable`, and `toggleNativeCommandTabIfNeeded` applied it like this:

```swift
setNativeCommandTabEnabled(false, Array(result.disable))
setNativeCommandTabEnabled(true, Array(result.enable))
```

`applicationWillTerminate` and `emergencyExit` do the same for all three. So if you'd turned ⌘\` off yourself, we turned it back on — on every shortcut edit, and again on the way out — and it stayed on after AltTab was gone.

Reading the live state on my machine, with AltTab bound to ⌘⇥:

```
hotkey 1 commandTab (⌘⇥):         enabled=false
hotkey 2 commandShiftTab (⌘⇧⇥):   enabled=false
hotkey 6 commandKeyAboveTab (⌘`): enabled=true
```

The first two are ours and correct. The third we have no claim on, and we switched it on anyway.

Once WindowServer owns ⌘\`, it eats the keystroke before our Carbon hotkey or the recorder can see it, so reassigning it in System Settings doesn't appear to take. Nothing in the recorder rejects ⌘\` as reserved — that list is only the three Force Quit ⎋ chords — so a user who turned it off cannot get it back while AltTab is running.

The fix is to remember which hotkeys we switched off, and switch back on only those. `NativeHotkeyOwnership` does that. It lives in `NativeHotkeyResolver.swift` so it compiles in the tests target, and `isEnabled` is injected. A hotkey that was already off when we want it is never claimed, so we never switch it on later. `setNativeCommandTabEnabled` becomes `disableNativeHotkeys` / `restoreNativeHotkeys` in `SkyLight.framework.swift`, which is where all four call sites already went, so the quit and emergency-exit paths are covered by the same change. `CGSIsSymbolicHotKeyEnabled` moves out of `experimentations/` now that it's used for real.

`resolve` itself is untouched. `enable` still means "we release our claim on these"; it's what we do with that set that changes, so the six existing tests pass unmodified. Six new ones cover the ownership kernel, and I updated Specs.md alongside them.

938 tests, 0 failures, on Xcode 26.6 / macOS 27.

Two caveats. Ownership is in-memory, so a hard kill can leave a hotkey off until the next clean quit — persisting it would close that window but needs a stored-state migration, so tell me if you'd rather have it. And I don't have a second display or an ISO keyboard here, so my manual testing was only the case above.


aviseth:

Issue #2053, opened 2022-10-28
Issues with command+backtick when Stage Manager is enabled


claudewowo:
**Describe the bug**

open ```Shortcuts - Select previous window```, set to ``` ` ``` key which is the key up ``` tab ```, does not worked on MacOS 13. It works fine on MacOS 12.6. The version is 6.48.0.

**Screenshots / video**

<img width="596" alt="iShot_2022-10-28_上午10 53 42" src="https://user-images.githubusercontent.com/8219715/198484734-87b652c2-38d7-4df1-8c2a-766679f99261.png">

lwouis:
Could you please share a screenshot of the second tab on that window (Shortcut 2)? Thank you

keman5:
And I can't change ``` cmd ``` to ``` option ``` on the second tab, it will be a error tip about: Shortcut key conflict

keman5:
I changed ``` cmd ``` to the third tab, and still does't works.

<img width="596" alt="iShot_2022-10-28_下午5 13 46" src="https://user-images.githubusercontent.com/116868115/198551230-1c91b1fa-2c92-4e27-8bfb-fef22cc141ad.png">

keman5:
But I can use ``` option ``` + ``` ` ``` to select previous window. I prefer to use ``` cmd ```, not others.

lwouis:
I have set my preferences exactly like you:

Shortcut 1 | Shortcut 2
-|-
<img width="686" alt="image" src="https://user-images.githubusercontent.com/106195/198570173-9014f530-f2a3-42ec-afbc-2ab5d9bfe6c3.png"> |  <img width="686" alt="image" src="https://user-images.githubusercontent.com/106195/198570188-06c3363e-fbc4-4f9e-996c-d90e692c4abf.png">

However, everything works great for me. I can use `command+tab` to open AltTab, then I can press ``` command+` ``` to go backwards.

I suspect it's some issue from an external factor like:

* External keyboard
* Software you use to remap keys (e.g. BTT, Karabineer, etc)
* Language mapping (e.g. using Chinese)
* Other things you may have forgotten you set on your machine in the past

Could you try resetting AltTab preferences (in the first tab of the preferences, at the bottom), and then changing `tab` to ``` ` ```? Maybe share that process in video so we can see what's happening exactly. For me, everything went smoothly.

keman5:
I'm on macos13

[...]

Issue #1734, opened 2022-06-24
command+` shortcut not working on macOS 13


L1cardo:
**Describe the bug**
"CMD+ \`" is not working
I tried to use "CMD+ `" as "select previous window", but it did not work.

**Screenshots / video**

<img width="610" alt="image" src="https://user-images.githubusercontent.com/33802186/175604713-b20a04d7-f931-4c80-aec0-1c7e86e75b45.png">

<details>
<summary><strong>Debug profile</strong></summary>
<p>

* App version: 6.46.1
* App preferences: 
  * MSAppCenter310AppCenterUserDefaultsMigratedKey: 1
  * MSAppCenter310CrashesUserDefaultsMigratedKey: 1
  * MSAppCenterAppDidReceiveMemoryWarning: 1
  * MSAppCenterInstallId: B825E425-EB28-4EB2-A325-6125A724B525
  * MSAppCenterNetworkRequestsAllowed: 0
  * MSAppCenterPastDevices: {length = 1155, bytes = 0x62706c69 73743030 d4010203 04050607 ... 00000000 0000038f }
  * MSAppCenterSessionIdHistory: {length = 460, bytes = 0x62706c69 73743030 d4010203 04050607 ... 00000000 00000156 }
  * MSAppCenterUserIdHistory: {length = 455, bytes = 0x62706c69 73743030 d4010203 04050607 ... 00000000 00000151 }
  * NSNavLastRootDirectory: /Applications
  * NSNavPanelExpandedSizeForOpenMode: {800, 448}
  * NSStatusItem Visible Item-0: 0
  * NSWindow Frame NSNavPanelAutosaveName: 560 336 800 448 0 0 1920 1055 
  * SUHasLaunchedBefore: 1
  * SULastCheckTime: 2022-06-25 上午7:05:57 +0000
[...]

PR #5965, opened 2026-08-19
fix: command+backtick shortcut had no effect

aviseth:
Binding ⌘\` in AltTab does nothing, and we've been switching off the wrong native shortcut.

`CGSSymbolicHotKey.commandKeyAboveTab` is 6. On macOS 27, reading the IDs back with `CGSGetSymbolicHotKeyValue` (the technique the comment above `CGSSetSymbolicHotKeyEnabled` already suggests) gives:

```
id  1 → keyCode 48 (⇥), ⌘      Move focus to next application
id  2 → keyCode 48 (⇥), ⇧⌘     Move focus to previous application
id  6 → keyCode 53 (⎋), ⌥⇧⌘    Force Quit the frontmost app
id 27 → keyCode 50 (`), ⌘      Move focus to next window in the active application
```

`com.apple.symbolichotkeys` agrees — the entry for ⌘\` is id 27, `parameters = (96, 50, 1048576)`, i.e. char `` ` ``, keyCode 50, modifier `0x100000`.

So 27 is the key-above-Tab one, and 6 is Force Quit — the same chord we already list as `MacOsShortcuts.forceQuitActiveApp`. Two consequences when someone binds ⌘\`: the native ⌘\` stays enabled and keeps eating the keystroke before our Carbon hotkey can see it, which is the reported symptom; and Force Quit gets switched off instead, and stays off after AltTab quits, since that state persists.

This also fits the history. The ID was introduced in 848ae5f for #2053, and shortly after the release @AfzalivE reported on that thread that ⌘\` was still broken for them on 6.49.0 with exactly this setup — which is what you'd expect if the toggle was landing on a different hotkey all along. #1734 looks like the same thing.

One line, plus the ID table in a comment so the next person doesn't have to re-derive it. 932 tests pass, Xcode 26.6 on macOS 27.

Two things I couldn't check. I don't have an ISO keyboard, and `matchesCommandKeyAboveTab` also hardcodes `kVK_ANSI_Grave` — the key above Tab isn't keyCode 50 on every layout, so that's probably still wrong there. Reading the keyCode off hotkey 27 with `CGSGetSymbolicHotKeyValue` instead of hardcoding it would fix that and handle a reassigned system shortcut too; happy to do that as a follow-up. And I only verified the ID table on macOS 27, not on older versions.

Independent of #5964, which is about not re-enabling hotkeys the user turned off; they touch different files.

### DIFF
--- a/src/macos/api-wrappers/CGSSymbolicHotKey.swift
+++ b/src/macos/api-wrappers/CGSSymbolicHotKey.swift
@@ -5,8 +5,18 @@ import Foundation
 /// `NativeHotkeyResolver` that surface this enum can compile in the unit-tests target. See
 /// `setNativeCommandTabEnabled` / `CGSSetSymbolicHotKeyEnabled` in `SkyLight.framework.swift` for the
 /// runtime side of toggling them.
+/// IDs verified on macOS 27 by reading each one back with `CGSGetSymbolicHotKeyValue`, and
+/// cross-checked against the `AppleSymbolicHotKeys` entries in `com.apple.symbolichotkeys`:
+///
+///     id  1 → keyCode 48 (⇥), ⌘        "Move focus to next application"
+///     id  2 → keyCode 48 (⇥), ⇧⌘       "Move focus to previous application"
+///     id  6 → keyCode 53 (⎋), ⌥⇧⌘      "Force Quit the frontmost app"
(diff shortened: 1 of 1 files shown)


AfzalivE:
I don't know if this is correct. I've been using cmd-` just fine nowadays since disabling the Mac OS shortcut.

aviseth:

PR #5967, opened 2026-08-19
fix: a shortcut could not be assigned because of an unrelated conflict

### LINKED ISSUE #5455: Can't assign shortcut to CMD + `
**Describe the bug**
I've been trying to assign shortcut CMD + ` but getting the error. But I've already reassign the system shortcut to another keys.

**Screenshots / video**

Image

**Your environment**

* AltTab version: 10.0.9.0
* macOS version: 15.7.4

aviseth:
Closes #5455.

I rebuilt the reporter's registry from the `defaults read` they posted — S1 = hold ⌘ + press ⎋, S2 = hold ⌥ + press ` — and ran `isShortcutAcceptable` over the edits that get you to ⌘\`:

```
holdShortcut2       <- ⌘    =>  CONFLICT blaming 'nextWindowShortcut'
holdShortcut3       <- ⌘    =>  CONFLICT blaming 'nextWindowShortcut'
```

So the dialog they hit is the plain conflict one, blaming "Shortcut 1 - Trigger". It's nothing to do with the macOS ⌘\` shortcut, which is why reassigning that in System Settings didn't help them.

The reason is a collision their config already had. Cancel is ⎋, and S1's hold is ⌘, so Cancel already produces ⌘⎋ — which is also S1's own trigger. Dumping every chord the saved config produces:

```
PRE-EXISTING COLLISION on kc53/mods256: ["cancelShortcut", "nextWindowShortcut"]
```

A saved config is never re-validated, so that just sits there. But the only way to get ⌘\` is to give another shortcut a ⌘ hold, and that recomputes Cancel under ⌘, lands on the same ⌘⎋, and gets reported. The chord isn't introduced by the edit and isn't made worse by it, and the user can't act on the message — they're editing Shortcut 2 and being told about Shortcut 1.

So: only report a chord the edit actually introduces. `isAlreadyUsedByAnotherShortcut` now skips a candidate combination when that same id already produces that same chord in the current config. `oldCombinationsExcludingTargetOfCandidate` and the new `currentCombinations` are the same walk with a different exclusion, so I pulled the body into one `combinations(_:)`.

A collision the edit *does* introduce still reports, and the new test pins both directions. It fails on master with exactly the reported error:

```
XCTAssertEqual failed: ("conflictWithExistingShortcut(shortcutAlreadyAssigned: "nextWindowShortcut")") is not equal to ("accepted")
```

933 tests pass, Xcode 26.6 on macOS 27.

Two things I'd flag. This doesn't clean up a pre-existing collision, it just stops it blocking unrelated edits — their ⌘⎋ stays ambiguous between Cancel and S1's trigger until they change one. And `isAlreadyUsedByAnotherShortcut` still only compares candidate chords against existing ones, never against each other, so two chords introduced by the same edit can still collide silently; that's unchanged here, but it's the reason my first attempt at the "still reports" assertion didn't fire.

### DIFF
--- a/src/kit/CustomRecorderControlTestable.swift
+++ b/src/kit/CustomRecorderControlTestable.swift
@@ -27,7 +27,7 @@ class CustomRecorderControlTestable {
         let oldCombos = oldCombinationsExcludingTargetOfCandidate(candidateId)
         // TODO: a user can assign a shortcut that's both .conflictWithExistingShortcut and .reservedByMacos
         // It's a mess to deal with. Current implem may let them bypass isReservedByMacos. Would be nice to think about what UX to do here
-        if let alreadyAssigned = isAlreadyUsedByAnotherShortcut(newCombos, oldCombos) {
+        if let alreadyAssigned = isAlreadyUsedByAnotherShortcut(newCombos, oldCombos, currentCombinations()) {
             return .conflictWithExistingShortcut(shortcutAlreadyAssigned: alreadyAssigned)
         }
         if let shortcutUsingEscape = isReservedByMacos(newCombos) {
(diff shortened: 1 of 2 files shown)


aviseth:
@lwouis Want to review this?

lwouis:
Thank you @aviseth for your help!

I've taken this commit, and ammended it a bit. It's on my dev branch, and will ship with the next release. I've credited you on the commit 👍

aviseth:
Thanks! I just realized I can get Alt-tab Pro for contributing? 🙂

Regards
Avi Seth
(he/him/his)

PR #5968, opened 2026-08-19
fix: switcher selected the window already on screen after closing a window

### LINKED ISSUE #5960: [Bug] Switcher selecting the wrong window when an application has multiple windows
AltTab v11.5.0 — macOS: Version 26.6.1 (Build 25G76)<br>License: Pro

> The problem began occurring after updating to version 11.5.0. When multiple windows are open, such as three Google Chrome windows—along with other applications like Slack, Terminal, and so on, closing one of those applications (e.g., Slack) and then pressing Command + Tab results in unexpected behavior.
> 
> Specifically, the application switcher now highlights the currently active Chrome window in the first slot, rather than maintaining the correct order and selecting the appropriate window from its original position in the switcher list.
> 
> Steps to Reproduce
> 1. Open two Chrome windows: One navigating to Google and the other navigating to YouTube.
> 2. Open one Terminal window.
> 3. Arrange the windows in this order (from front to back): Terminal (frontmost), Chrome (Google) (middle), Chrome (YouTube) (backmost).
> 4. Close the Terminal window.
> 5. Press Command + Tab.
> 
> Expected Result:
> The first window shown in the switcher should be Chrome (Google), and the second should be Chrome (YouTube). Pressing Command + Tab once should switch to Chrome (Youtube), and pressing it again should switch back to Chrome (Google).
> 
> Actual Result:
> After closing the Terminal, Chrome (Google) appears in the first slot (which is correct), but pressing Command + Tab once shows Chrome (Google) as the highlighted window, instead of Chrome (Youtube), which is not correct. Basically, the switcher is highlighting the first window, instead of the second window.

<details>
<summary><strong>Debug profile</strong></summary>
<p>

* App preferences: 
  * appearanceSize: 0
  * appearanceStyle: 1
  * appearanceTheme: 0
  * arrowKeysEnabled: false
  * captureWindowsInBackground: false
  * crashPolicy: 2
  * fadeOutAnimation: true
  * hideStatusIcons: false
  * holdShortcut: {
    secureData = {length = 357, bytes = 0x62706c69 73743030 d4010203 04050607 ... 00000000 00000101 };
    string = "\U2318";
}
  * menubarIcon: 1
  * previewFadeInAnimation: true
  * previewFocusedWindow: false
  * screenRecordingPermissionSkipped: true
  * screensToShow: 1
  * settingsWindowShownOnFirstLaunch: true
  * shortcutCount: 1
  * showHiddenWindows: 1
  * showMinimizedWindows: 1
  * showWindowlessApps: 1
  * updatePolicy: 1
  * windowDisplayDelay: 100
* Applications: 62
* App icons: {bitmap: 144px, displayed: 72pt @2.00x = 144px, sources: min 512px, max 512px}
* Windows: 17
* OS architecture: arm64
* Locale: en_US (fixed en_US)
* InputSource: U.S. International – PC
* Spaces: 2
* Dark mode: Light
* "Displays have separate Spaces": checked
* Hardware model: Mac16,7
* Screens: 2
  * {x: 0.0, y: 0.0, width: 2560.0, height: 1440.0, scale: 2.00, dpi: 186}
  * {x: -1728.0, y: 323.0, width: 1728.0, height: 1117.0, scale: 2.00, dpi: 255}
* CPU model: Apple M4 Pro
* Memory size: 51.54 GB
* Active CPU count: 14
* Current CPU frequency: nil
* Resource utilization: 
  * CPU: 0.5%
  * Memory: 59M+
  * Threads count: 12/1

</p>
</details>

_Opened by the in-app feedback form._

aviseth:
Closes #5960.

Following the reporter's steps — Terminal frontmost over Chrome/Google and Chrome/YouTube, close Terminal's window, summon — the pick lands on the front tile instead of stepping over it. That's Chrome/Google, the window already on screen, so the shortcut looks like it does nothing.

`currentWindowIsDrawn` decides whether the default steps over the front tile, and `defaultPickIndexAsOfSummon` returns `front` when it's false. Two different situations make the frontmost app own no drawn tile, and they want opposite answers:

- Its windows are in the model but a filter dropped them (`Apps to show: Non-active apps`). The user is on one of them, the front tile is already the previous window, don't step over it — #5941.
- It has no window to draw at all, because its last one was just closed. Nothing was filtered. The window now on screen belongs to a different app and sits at the front of the MRU, so the front tile *should* be stepped over.

`list.contains { $0.application.pid == frontmostPid && shouldDisplay($0) }` is false in both, so the second inherited the first's answer. Closing an app's last window doesn't quit it on macOS, so it stays frontmost with nothing to contribute to the list, and every summon after that lands one tile short until you focus something else.

That also fits it being new in 11.5.0: `currentWindowIsDrawn` arrived with the #5941 fix a few days before the release.

I moved the decision into `SelectionResolver` so it can be tested — the shell version reads `Applications.frontmostPid` and `Windows.list` and had no coverage. Only real windows count now: a windowless app is the placeholder tile for exactly the "no windows" case, so letting it stand in for a window the filters dropped would put us back where we started. My own test caught that one, which is why it's spelled out.

Five kernel cases plus the reporter's scenario end to end. The three existing #5941 tests (F1–F3) pass unmodified. 938 tests, 0 failures, Xcode 26.6 on macOS 27.

One thing I couldn't check: I don't have a way to run this against the real WindowServer, so the reasoning about which situation produces which answer is from the code, not from watching it. The `Applications.frontmostPid` timing after an app actually *quits* (rather than just losing its last window) is the part I'd most want a second opinion on — if it updates promptly then that path was never broken, and only the closed-last-window path is.

### DIFF
--- a/src/switcher/state/SelectionResolver.swift
+++ b/src/switcher/state/SelectionResolver.swift
@@ -63,6 +63,14 @@ struct SelectionInputs: Equatable {
     var currentWindowIsDrawn = true
 }
 
+/// One window of the frontmost app, as `currentWindowIsDrawn` sees it.
+struct FrontmostAppWindow: Equatable {
+    /// The placeholder tile an app with no windows gets, not a window the user can be looking at.
+    let isWindowlessApp: Bool
+    /// Passed `Windows.shouldDisplay`, i.e. survived this shortcut's filters.
+    let isDrawn: Bool
+}
+
 /// What the kernel recommends. Wrapper translates this into side effects (highlight redraws,
 /// scroll-to-visible, thumbnail preview, etc.).
 enum SelectionDecision: Equatable {
@@ -117,6 +125,28 @@ enum SelectionResolver {
         }
         // 6) Target gone — adapt to the closest visible.
         return adapt(i, visibleIndexes: visibleIndexes, lastVisible: visibleIndexes.last!)
+    }
+
+    /// Is the window the user is looking at among the drawn tiles? Pure half of
+    /// `Windows.currentWindowIsDrawn`, which passes the frontmost app's windows.
+    ///
+    /// Two different situations produce "the frontmost app owns no drawn tile", and only one of them means
+    /// the user's window is missing from the list:
+    ///
+    /// - Its windows are in the model but a filter dropped them (`Apps to show: Non-active apps`). The user
+    ///   is looking at one of them, it isn't drawn, and the front tile is already the previous window — so
+    ///   don't step over it (#5941).
+    /// - It has no window to draw at all, because its last one was just closed. Nothing was filtered. The
+    ///   window the user is now looking at belongs to a different app and sits at the front of the MRU, so
+    ///   the ordinary rule applies and the front tile must be stepped over (#5960). Answering `false` here
+    ///   selected the window the user was already on.
+    ///
+    /// A windowless app is the placeholder tile for the second case, never something the user is looking at,
+    /// so it doesn't count as a window the filters could have dropped.
+    static func currentWindowIsDrawn(_ frontmostAppWindows: [FrontmostAppWindow]) -> Bool {
+        let realWindows = frontmostAppWindows.filter { !$0.isWindowlessApp }
+        guard !realWindows.isEmpty else { return true }
+        return realWindows.contains { $0.isDrawn }
     }
 
     /// Mirrors `setInitialSelectedAndHoveredWindowIndex` — picks the index, defers reset to wrapper.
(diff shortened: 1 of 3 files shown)


lwouis:
Thank you for sharing!
I've already a similar fix on the dev branch. Mine has a bit more tests, so I'll move forward with it.
Thank you for the PR in any case. It reinforces that the fix is correct! 👍

aviseth:
> Thank you for sharing!
> I've already a similar fix on the dev branch. Mine has a bit more tests, so I'll move forward with it.
> Thank you for the PR in any case. It reinforces that the fix is correct! 👍 

Okay great!

PR #5975, opened 2026-08-21
fix: focus java minecraft

Jack-Works:
Hi! This PR is AI coded, but I manually verified it on my machine.

This PR fixes this: when focus the window of Java Minecraft, Minecraft gets focus (become active), but the Minecraft window is still convered under other windows (not become the top window).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ WorkspaceSettings.xcsettings
 IDEWorkspaceChecks.plist
 /node_modules/
 /DerivedData/
+/DerivedDataDebug/
 /build/
 /vendor/*/.build/
 /vendor/*/.swiftpm/
--- a/src/switcher/state/Window.swift
+++ b/src/switcher/state/Window.swift
@@ -412,6 +412,12 @@ class Window {
                 //      .invalidUIElement and no-ops, so re-resolve the live element by wid, retry, and heal the
                 //      cache; _SLPS/makeKeyWindow above use the wid/psn directly so they're unaffected.
                 //   4. cross-Space only: restore the origin Space's front process (see snapshot above).
+                if self.activateExactDockItemForUnbundledApplication() {
+                    DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(50)) {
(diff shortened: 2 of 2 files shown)


lwouis:
Thank you @Jack-Works for sharing this work
Could you please share a link to where I can get Java Minecraft to investigate the issue?
Thank you

Jack-Works:
@lwouis hi! You can use a (third party) launcher like https://github.com/HMCL-dev/HMCL to download Minecraft, but you need buy it from Microsoft to download/run it legitimately.

Here is my Java version:

```shell
$ /usr/bin/java --version
openjdk 25.0.3 2026-04-21 LTS
OpenJDK Runtime Environment Zulu25.34+17-CA (build 25.0.3+9-LTS)
OpenJDK 64-Bit Server VM Zulu25.34+17-CA (build 25.0.3+9-LTS, mixed mode, sharing)
```

Since Minecraft ecosystem is very complicated, I cannot make sure you can reproduce it.

lwouis:
Thanks for the PR. I'd like to fix this in our existing focus path
rather than via the Dock, but I can't reproduce your setup, so here's a
[diagnostic build](https://github.com/user-attachments/files/31313584/AltTab.app.zip) that will tell us which step actually fails:

1. Quit your normal AltTab, then run this one (right-click > Open the first time).
   The "Debug tools" window opens by itself.
2. Get Minecraft into the state where focusing it fails.
3. In the "Focus probe (PR #5975)" box, check the dropdown has the Minecraft
   window selected (hit "Refresh list" if not).
4. Move another app's window so it fully covers the Minecraft window. This
   matters: the probe measures whether Minecraft comes to the front, so it has
   to start buried.
5. Click "Run focus probes". You get a 6s countdown: click the covering window
   during it, then don't touch anything for ~25s. The screen will flicker.
6. Click "Copy probe report" and paste it here.

It tries 12 focus sequences in isolation and reports the real window stacking
after each, so we'll see which one raises your window and which don't.

Thank you

Jack-Works:
sorry that's strange, I cannot reproduce this either now. I will close this for now, and if it appears again, I will use this to give you the log. thank you!

Issue #5914, opened 2026-08-03
Command+Shift+Tab behaviour is missing, only Command+Shift makes switching unlike original system behaviour


wanderseeme:
**Is your feature suggestion related to a problem? Please describe.**

1. I'm using Command+Tab to launch AltTab switch panel.
2. Every time I wanna swtich backwards, I intend to use original system Command+Shift+Tab shortcut.
3. This Command+Shift+Tab shortcut makes switching backwards "too fast", AltTab seems use Shift+Tab istead.
4. This would be very annoying if you already get used to original system Command+Shift+Tab shortcuts.

**Describe the solution you'd like**
Support Command+Shift+Tab shortcut, as the original system did.

**Describe alternatives you've considered**
Make shortcut options for this.

lwouis:
Hi,
By default, AltTab only requires pressing `shift` to cycle backwards. We think it's more ergonomic.
For people who prefer the established `shift+tab` shortcut, it's easy to customize: Settings > Controls > Shortcuts while open > Select Previous window: update the shortcut from `shift` to `shift+tab`.
Thank you

PR #5990, opened 2026-08-26
fix: cycling backwards no longer stops at the first window

matthue-lee:
## The bug

With the default `previousWindowShortcut` (the bare modifier `⇧`), backwards cycling dead-ends on the first window. Once you are there, further backwards input does nothing — you have to release ⇧ and press it again to get one more step. Forward cycling has no equivalent dead-end.

## Why

`KeyRepeatTimer` synthesizes repeats for shortcuts that have **no keycode of their own**, since those get no OS key-repeat. The default `previousWindowShortcut` is exactly that.

A modifier-only shortcut's `state` is decided by `ATShortcut.matches` from the event's modifiers alone, so it stays `.down` for as long as ⇧ is physically held. `stopRepeatIfUp` never fires, and the timer therefore stays armed for the whole hold.

The wrap-around rule used `!KeyRepeatTimer.timerIsSuspended` as its "this is a key repeat" signal. During a ⇧ hold that is true for *every* backwards advance, including real key presses, so each one was refused its wrap.

Forward cycling escapes this because `nextWindowShortcut` carries a real keycode: the local monitor reports `keyCode: nil` on key-up, `matches` flips the state to `.up`, and each release suspends the timer — leaving it suspended when the next press is handled.

## The fix

Gate wrap-around on whether an advance **is** one of the timer's synthesized ticks — a fact about that advance — instead of on whether a timer is armed, which is a fact about the session. `KeyRepeatTimer.isFiringArtificialRepeat` is set only for the duration of a tick's `block()` call.

Holding a key still walks to the end and stops there, unchanged. Only genuine presses regained their wrap.

This is the behaviour `docs/contributing.md` already specifies, under QA > Shortcuts:

> when navigating left/right/up/down, the repeating behavior should stop when hitting the last window in the list. The user can then manually do the shortcut once more to cycle to the other side; it then repeats again

The first sentence held. The second did not, for any shortcut without a keycode of its own: the manual press was suppressed by the same rule that stopped the repeat.

## Structure

The rule was written out twice — `Windows.cycleSelectedWindowIndex` and `TilesView.nextRow` — and neither copy had a test. It now lives in `CycleWrapResolver`, following the same pure-kernel pattern as `SelectionResolver` / `ShortcutModifierResolver`, with `CycleWrapResolverSpecs.md` and `CycleWrapResolverTests.swift`.

The test for the bug was written first and watched fail against a faithful port of the old rule, before the rule was changed.

## Verification

- `xcodebuild test -scheme Test` — 937 tests, 0 failures (932 before, + 5 new)
- Debug app target compiles clean

## Note

This does **not** address the separate "backwards cycling is too fast" report in #5914. That one is a different defect in `ATShortcut.matches` — modifier-only shortcuts re-fire on every key event while held, so one ⌥⇧+⇥ tap steps back three windows. It is tracked separately in #5991, with the measured counts and two possible directions. Related context: #5924.

### DIFF
--- a/alt-tab-macos.xcodeproj/project.pbxproj
+++ b/alt-tab-macos.xcodeproj/project.pbxproj
@@ -38,6 +38,9 @@
 		5883F0A829A5182C0071DB65 /* TrackpadEvents.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5883F0A729A5182C0071DB65 /* TrackpadEvents.swift */; };
 		5E1EC110000000000000C002 /* SelectionResolver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E1EC110000000000000C001 /* SelectionResolver.swift */; };
 		5E1EC110000000000000C003 /* SelectionResolver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E1EC110000000000000C001 /* SelectionResolver.swift */; };
+		5E1EC110000000000000D002 /* CycleWrapResolver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E1EC110000000000000D001 /* CycleWrapResolver.swift */; };
+		5E1EC110000000000000D003 /* CycleWrapResolver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E1EC110000000000000D001 /* CycleWrapResolver.swift */; };
+		5E1EC110000000000000D102 /* CycleWrapResolverTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E1EC110000000000000D101 /* CycleWrapResolverTests.swift */; };
 		5E1EC110000000000000C102 /* SelectionResolverTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E1EC110000000000000C101 /* SelectionResolverTests.swift */; };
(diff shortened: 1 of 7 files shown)


matthue-lee:
Would be super keen to see this merged in. Cycling backwards is pretty much half of all the cycling I do, somewhat naturally.

lwouis:
Hi,

I can't reproduce the issue. What I think is happening is that you're trying to use the Windows-style backwards shortcut, which is holding Alt and pressing, holding Alt and Shift and pressing Tab, Tab, Tab, Tab, Tab. You're actually pressing that on the keyboard, but the default from Alt-Tab is just Shift. So you should be holding Tab and pressing Shift, Shift, Shift, Shift, Shift. So I would suggest one of two things: either you use the default from Alt-Tab, which I chose for ergonomics. I think pressing Alt and then Shift, Shift, Shift, Shift is way easier on the fingers than pressing Alt, Shift, and then Tab, Tab, Tab, Tab. Either that, or number two, you can go in the settings and customize the backwards cycling as shown on the screenshot, and that way you will get the exact Windows experience you may be used to. If you do either of those, you will not have any issues. It will cycle and stop at the edge, and then you press again and it does the full cycle again on repeat until it hits the edge again. I'll close the ticket for now. Thank you. 

image

Issue #5861, opened 2026-07-16
Follow-up to #5839: Preview capture can wedge the macOS screenshot service for tens of seconds (macOS 26/27)


Atmospenguin:
**Describe the bug**

This is a follow-up to #5839, which reported the same symptom on AltTab 11.4.2 / macOS 26.5.1 but was closed without enough diagnostics to reproduce it. I can reproduce and isolate it on AltTab 11.4.3 / macOS 27.0 beta, and I have a tested mitigation.

Opening the AltTab switcher with window Preview enabled can make every screenshot path on the Mac wait for tens of seconds: the built-in screenshot UI, `/usr/sbin/screencapture`, and third-party screenshot apps.

Quitting AltTab makes screenshot responsiveness return immediately. Reopening AltTab makes the problem reproducible again.

This still happens with **Capture windows in the background disabled**. The trigger is opening the switcher itself, not periodic background refresh.

**Steps to reproduce the bug**

1. Grant AltTab Screen Recording permission.
2. Enable the selected-window Preview.
3. Disable **Capture windows in the background**.
4. Keep roughly 13 or more eligible windows open.
5. Open the AltTab switcher.
6. Immediately take a system screenshot or use a third-party screenshot tool.
7. The screenshot request may not react for tens of seconds. Quit AltTab and repeat; screenshot latency returns to normal.

**Expected behavior**

Capturing thumbnails for the switcher should not block unrelated system or third-party screenshot clients.

**Observed capture trace / likely root cause**

I traced the ScreenCaptureKit/ReplayKit path rather than relying only on CPU observations:

- One switcher activation submitted **13–14 `RPDaemonProxy captureScreenshot` requests in about 2 ms**.
- The current `screenshotThrottler` key is per window, so it does not cap aggregate capture concurrency.
- `BackgroundWork.screenshotsQueue` limits operation execution, but each operation returns after submitting the asynchronous SCK request; its slot is therefore released before the capture callback completes.
- Individual requests took around **2–7 seconds** on the affected machine; at least one request had no completion for more than **90 seconds**.
- One four-minute trace contained **119 AltTab capture requests**, while WindowServer completion lagged well behind submission.
- `replayd` samples showed synchronous waits while publishing capture attribution to `systemstatusd`; `systemstatusd` was busy in media-attribution/reference-count cache work.
- Once this backlog forms, unrelated screenshot clients appear to wait on the same system capture/attribution path.

[...]

Issue #5112, opened 2025-12-10
Hanging when waking from sleep or after resetting permissions


injust:
**Describe the bug**

> Might be related, but I find AltTab is sometimes not responding when my MBP wakes from sleep.
> 
> In the Activity Monitor, it's burning through CPU:
> 
> <img width="755" height="24" alt="Image" src="https://github.com/user-attachments/assets/dd1d1f9a-a150-4134-9878-a1371fd901a1" />
> 
> If I run AltTab (to open the UI), it doesn't open until AltTab is responsive. 

 _Originally posted by @injust in [#5109](https://github.com/lwouis/alt-tab-macos/issues/5109#issuecomment-3634410894)_

**Your environment**

* AltTab version: 7.38.0
* macOS version: Sequoia 15.7.2

injust:
@lwouis I opened a new issue because it seems unrelated to the original one. 

Spindump: [Spindump.txt.zip](https://github.com/user-attachments/files/24070175/Spindump.txt.zip)

I was able to reproduce it like this:

1. Quit AltTab
2. `sudo tccutil reset All com.lwouis.alt-tab-macos`
3. Launch AltTab
4. AltTab should prompt for Screen Recording permission -- grant it

Then AltTab burns CPU and hangs on this screen:

<img width="633" height="539" alt="Image" src="https://github.com/user-attachments/assets/39a91047-8023-4f1d-8454-3c6d190357f2" />

raphaelreimann:
Same problem here. Completely stuck and non responding on the "AltTab needs some permissions" screen

lwouis:
@raphaelreimann are you also on macOS 15?

lwouis:
@raphaelreimann @injust Could you please try out [this local build](https://github.com/user-attachments/files/24088027/AltTab.app.zip), and let me know if it fixes the issue for you?

Thank you!

Marsman1996:
Same problem here. Hang in AltTab version 7.38.0 with MacOS Version 26.1 (25B78)
Back to 7.37.0 works fine

Marsman1996:
> [@raphaelreimann](https://github.com/raphaelreimann) [@injust](https://github.com/injust) Could you please try out [this local build](https://github.com/user-attachments/files/24088027/AltTab.app.zip), and let me know if it fixes the issue for you?
> 
> Thank you!

The local build works fine for me.

raphaelreimann:
[...]

PR #6006, opened 2026-09-05
fix: preserve capture state after service failures

masonc15:
I've been running into repeated permission prompts from AltTab when my Mac is under load, with Screen Recording access already enabled. A capture timeout can send it back to the permission window even though the grant hasn't changed.

This keeps the last-known grant and uses non-prompting checks to confirm later failures. Accessibility checks stay separate. First-time denial still opens the normal permission window.

ScreenCaptureKit gets a shared limit of two operations. Queued requests check the switcher, preferences, lock state, and permission again when they're ready to run. The WindowServer path checks those conditions too, including before a fallback capture. That stops queued work from starting after the switcher closes when background capture is disabled.

UI state stays on main. Permission preflight and ScreenCaptureKit submission run on a worker queue, and focused previews go ahead of queued thumbnails. If a completion never arrives, its slot stays occupied and the user gets a restart action. Permission snapshots are locked, and confirmation work runs in order.

The Release tests also caught a stale switcher-visibility read that could leave previews missing. That's fixed here. On macOS 27 and later, thumbnails use the existing WindowServer backend after a known grant. Focused previews still use ScreenCaptureKit, with a cooldown after failures.

Related reports: [#5112](https://github.com/lwouis/alt-tab-macos/issues/5112) and [#5861](https://github.com/lwouis/alt-tab-macos/issues/5861). Both were closed when I checked them. They're useful context; I'm not claiming this resolves everything described there or replaces the existing fixes.

The backend change comes from logs on one macOS 27 beta host, where requests used executable-path identities while the bundle grants stayed allowed. I don't know how broadly that applies.

## Tests

- Debug build and full Debug scheme test action passed: 957 tests, no failures or skipped tests.
- Changed Swift lines passed SwiftFormat 0.63.0. Diff whitespace check passed.
- Tests cover confirmation, recovery, backend selection, request limits, lost completions, and thumbnail and focused-preview behavior.
- The latest Release candidate passed all 1,190 full-suite tests with zero failures. All 67 changed Swift ranges passed formatting. Regression tests first failed for main-thread submission, preview priority, dispatch ordering, timeout-stage ownership, and queued WindowServer capture after closure. P1 review has no unresolved P0/P1 finding. The final branch includes upstream `2f3c6773` (11.6.0); that last release commit changes no app source or build files.
- Live checks showed real thumbnails and the separate full-size focused preview. With eight CPU workers active, a fixture-only injected capture error entered the normal cooldown. The cached preview remained available. After the fixture content changed, a real 960-by-896 ScreenCaptureKit completion supplied the new thumbnail and full-size preview without an app restart.
- A thirty-second closed-switcher observation with background capture disabled produced no capture debug entry. The final native state confirmed that the switcher stayed closed.
- Three consecutive sixty-second, eight-worker load runs on the refreshed candidate each completed twenty observed switcher openings within fifty seconds. There were no late replies, app restarts, new permission alerts, or matching path-based TCC rows. Dayflow saved six new frames per run. All workers stopped. A separate preview check on this candidate produced 0 / 21,733 / 0 green pixels outside the tile panel before, during, and after the preview. A further 38-second closed-switcher check had no capture debug entry.

Testing was on Apple Silicon with macOS 27.0 build 26A5425a. Xcode signing was disabled, and the Sparkle helper script needed an ad-hoc identity override for temporary build output. The live Release candidate used my existing local signing identity. The normal app and changed settings were restored afterward. No release was distributed.

I haven't tested actual permission revocation, minimized and full-screen previews, sleep, wake, or display changes in the app yet. The 24-hour normal-use test and older macOS runtime tests are also open.

The earlier injected-error recovery checks had a permission dialog open already. Those show recovery; they can't establish whether another prompt would have appeared. The three final load runs started without a dialog. I've kept the failed attempts in the test report too, including a five-second state-read reply.

One earlier `--hide` request reached the main-thread hide routine immediately and took 2.703 seconds to prepare its reply. That missed the client's two-second deadline. The run remains failed, and there's no sample of the exact slow call. I'm not claiming a fix for that separate delay.

There's also an OS-side failure these changes can't prevent. During a later setup, Apple's permission service reported `Too many open files` while resolving both apps' signed code. Dayflow's preflight had passed about one second earlier, and the next system check still caused an Apple alert. The bundle grants stayed allowed, and no matching path row appeared.

This reduces repeated requests and preserves capture state. It can't make Apple's preflight and capture request atomic or guarantee that no Apple alert will ever appear. There are no TCC, signing configuration, updater, or release-version changes in the PR.

I'd especially appreciate your thoughts on the macOS 27 backend choice. That part is based on the behavior above, and I've kept it separate for review. Thanks!

### DIFF
--- a/alt-tab-macos.xcodeproj/project.pbxproj
+++ b/alt-tab-macos.xcodeproj/project.pbxproj
@@ -381,6 +381,9 @@
 		5F20BB00000000000000B002 /* ScreenCaptureCoordinatorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F20BB00000000000000B001 /* ScreenCaptureCoordinatorTests.swift */; };
 		5F20BB00000000000000B005 /* ScreenCaptureCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F20BB00000000000000B004 /* ScreenCaptureCoordinator.swift */; };
 		5F20BB00000000000000B006 /* ScreenCaptureCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F20BB00000000000000B004 /* ScreenCaptureCoordinator.swift */; };
+		5F20CC00000000000000C002 /* WindowCaptureRoutingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F20CC00000000000000C001 /* WindowCaptureRoutingTests.swift */; };
+		5F20CC00000000000000C005 /* WindowCaptureRouting.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F20CC00000000000000C004 /* WindowCaptureRouting.swift */; };
+		5F20CC00000000000000C006 /* WindowCaptureRouting.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F20CC00000000000000C004 /* WindowCaptureRouting.swift */; };
 		D04BADE4DDF7597749FB13BD /* SettingsWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = D04BA099FA155FB4D517D66B /* SettingsWindow.swift */; };
 		D04BAF12DF5D15B9D7D316A4 /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = D04BA61693F710CD7BD054D7 /* InfoPlist.strings */; };
 		D04BAFB217F9F8E58C3B09AD /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = D04BA82BB17B85A88A5D2B87 /* InfoPlist.strings */; };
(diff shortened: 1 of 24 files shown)


lwouis:
Hi,

Thank you for taking the time to report this issue and propose a solution.

I want to make sure I understand the core problem correctly. On your machine, AltTab sometimes checks whether it has the required permissions, but that check can time out after five or six seconds. My understanding is that you believe this happens because macOS is too busy to respond in time, and your proposal is essentially to avoid treating that timeout as evidence that the permission is missing.

My concern is that there may be a more fundamental issue here. The permission check ultimately relies on the operating system, and we depend on the OS in a similar way when retrieving screenshots. If macOS is too busy to tell us whether we have permission, it may also be too busy to reliably provide the data AltTab needs to operate.

In that situation, I'm not sure it would be safe for AltTab to simply continue in a degraded mode. If we cannot reliably communicate with the OS, what can we actually trust ourselves to show in the switcher? We could end up displaying incomplete or incorrect information, which would make AltTab appear broken even though the underlying problem is that the OS is not responding properly.

For that reason, I think the current behavior has some merit: if we cannot get a clear answer from macOS after five or six seconds, something unusual is happening, and asking the user to review the permission state is a relatively safe failure mode.

That said, I agree that the current UX may not be ideal, and perhaps there is another way to handle this.

One possibility would be to add an option to the permission window along the lines of "I know AltTab has permission; continue anyway." If the user explicitly chooses that, AltTab could stop re-checking the permission during that runtime after launch.

However, that introduces another problem: the permission could later be revoked, or the user could forget that they enabled this override. AltTab would then continue trying to operate without the required access, and from the user's perspective it would simply look broken.

So I find this a tricky situation. I spent some time reviewing both the proposed code and the broader implications, and at this point I don't feel comfortable merging the change as-is. It raises several questions about correctness and UX that I don't yet have good answers to.

If you have other ideas for how we could handle this from a UX perspective, especially a way to distinguish a slow permission check from an actually missing permission, I'd be interested to hear them.

Thank you

Issue #5757, opened 2026-06-14
Weird artifact on MacOS 27 ( Golden Gate )


budgetdevv:
**Describe the bug**

There is a weird UI artifact when trying to alt tab on MacOS 27.

**Screenshots / video**

<img width="520" height="396" alt="Image" src="https://github.com/user-attachments/assets/58bb2207-4ba1-468c-9d98-2bfc1a25f79c" />

As shown, there's a weird black border artifact when trying to alt tab

**Steps to reproduce the bug**

1. Alt tab
2. Observe the UI

**Your environment**

* AltTab version: 
* macOS version: MacOS 27.0 Beta ( 26A5353q )
* Other relevant info: NIL

lwouis:
Hi,
Let's monitor this, and hopefully Apple will fix this issue during the Beta. They break some things every single Beta. So far, they've fixed like 80% of the issues during the Betas.
Thank you

budgetdevv:
> Hi, Let's monitor this, and hopefully Apple will fix this issue during the Beta. They break some things every single Beta. So far, they've fixed like 80% of the issues during the Betas. Thank you

Yeah you're right, thanks for looking into this!

lwouis:
@budgetdevv could you please let me know if [this build](https://github.com/user-attachments/files/29110924/AltTab.app.zip) helps somehow?

budgetdevv:
@lwouis, I tried the new build but it did not fix the issue unfortunately

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/7ef412dd-33e4-49a5-8d70-32a558ca5cd9" />

therealFoxster:
Still an issue in developer beta 2

lwouis:
Thank you for reporting @therealFoxster 

I hope they fix it before release.

ThatGingerbread:
beta 3 did not fix this either

budgetdevv:
Anyone tried this on beta 4?

therealFoxster:
Not fixed on beta 4

Nomamuk:
Not fixed on Beta 5

lwouis:
Thank you for sharing
It's unfortunate Apple is not fixing the issue.
Please report the issue to Apple if possible, so they are aware of it and fix it 🙏 
Thank you

ThatGingerbread:
still isnt fixed in beta 6 🙂‍↔️

mario-donnarumma:
[...]

PR #6029, opened 2026-09-12
fix: clip glass panel corners on macOS 27

josdyr:
On macOS 27 the switcher's glass panel can show a rectangular outline outside its rounded corners, and it wiggles while switching (#5757).

### Changes
- Clip the glass backing layer to the same continuous corner radius as `NSGlassEffectView`, and update it when a cached effect view's appearance changes.
- macOS 27 and later only. The material and window shadow are unchanged; macOS 26 and earlier are untouched.
- Manual regression scenarios added to `AppearanceSpecs.md`.

### Testing
- Reproduced on macOS 27.0 (26A428) with the Titles style. Repeated summon/dismiss captures show the rectangular outline before and rounded corners after, and the outline no longer wiggles during normal switching.
- Disabling the window shadow, allowing glass layout, or clipping to a rectangle didn't fix it in experiments; only the rounded clip is included. This works around the rendering and doesn't claim to identify Apple's underlying bug.
- Builds with Xcode 27. Not checked: dark mode, other styles and sizes, macOS 26, later macOS 27 builds, CI.

Related: #5757

---


lwouis:
This was cherry picked in the latest release. Thank you very much!

PR #6030, opened 2026-09-13
feat: show website icons for browser windows

josdyr:
## Summary

Adds an opt-in **Show website icons** setting (Settings > Appearance). Browser windows show the displayed site's icon instead of the browser's app icon, so several windows of the same browser are easy to tell apart.

- No browser extension and no browser-specific code. The page URL comes from the standard macOS accessibility attributes (`AXDocument`, or `AXURL` on the tab's `AXWebArea`) that WebKit, Chromium and Firefox browsers expose.
- The icon is fetched from the site's public homepage only (no page path, query, cookies or credentials) and normalized into a rounded tile like app icons.
- Off by default. Turning it off restores app icons immediately.

The branch now holds one tidied commit, [`2e89ba01`](https://github.com/josdyr/alt-tab-macos/commit/2e89ba01). It follows the Specs/Tests/implementation triad: `WebsiteIcons.swift`, `WebsiteIconResolver.swift`, `WebsiteIconRenderer.swift`, `WebsiteIconsSpecs.md` and 13 unit tests in `WebsiteIconsTests.swift`. Details, measurements and browser coverage are in the comment below.

Related: #527, #4651


lwouis:
Hi @josdyr,

I have drawn the line for AltTab with web browsers. None of them expose their tabs contents in a standard way. This means that integrating with a browser is unique to each browser. So if we would support them, we would need to implement N custom integrations. Furthermore, the browser vendors constantly update their browsers. Sometimes multiple times a day. These updates would break our integrations regularly. This is an impossible pace for me to maintain N custom integrations over time.

For these reasons, I've decided not to attempt supporting custom browser integrations.

Thank you

josdyr:
Hi @lwouis, thanks for taking the time to explain, and for AltTab. I understand the concern about maintaining N browser integrations that break with every browser release. I'd like to share what we found, because this implementation doesn't integrate with any browser, and I think that changes the maintenance picture.

## What it looks like

AltTab switcher with website icons on Safari, Chrome and Firefox windows

*The switcher with **Show website icons** on (macOS 27). Safari, Chrome and Firefox windows each show the icon of the site they display, so several windows of the same browser are easy to tell apart. The selected row shows GitHub's transparent logo on a white backing, which keeps it visible on the accent color. Everything else keeps its normal app icon. Note: the separate app-name column and the Titles layout tweaks come from my personal build; this PR only changes the icon.*

## No browser integration at all

- **No extension, no AppleScript or Apple Events, no JavaScript, no per-browser code or bundle IDs.** There is one generic code path.
- The page URL comes from the **standard macOS accessibility attributes**: `AXDocument` on the window, or `AXURL` on the tab's `AXWebArea`. These are the attributes VoiceOver relies on, so every engine keeps them working: WebKit, Chromium, and Firefox since Mozilla's 2020 macOS accessibility rewrite.
- Tested **unchanged** in Safari, Safari Technology Preview, Chrome, Orion and Firefox 155 on macOS 27. I never wrote a line for Orion or Firefox; they just worked.
- The contract is the public accessibility API, not browser internals. A browser update doesn't need an AltTab change. If a browser ever stopped exposing the URL, its windows would simply show the normal app icon again. Nothing breaks and nothing needs patching.

## How it works

1. **Find the page:** a bounded background accessibility lookup (max 128 elements, depth 8, 100 ms per call). It stops at the first web area, never reads page content, and never inspects AltTab's own windows.
2. **Fetch the icon:** only the site's public homepage `https://host/` is requested. The page path, query, fragment, cookies and credentials are never sent. It reads `<link rel="icon">` / `apple-touch-icon`, then falls back to `/favicon.ico`. Local and private network addresses are refused. Work is shared and capped: 8 transfers, 1 MiB responses, HTML stops at `</head>`, results cached for 30 s.
3. **Draw it:** a 64 px rounded tile like app icons. Transparent logos get a white backing, and logos that would vanish on white get a dark one, so icons stay visible on the selected row in light and dark mode. Web pages without an icon show a globe; every other window keeps its app icon.

It's **off by default**. The setting's description says the requests happen outside the browser, including for private windows.

## Size and maintainability

- **10 files, +881/-3.** The feature lives in 3 new files: `WebsiteIcons.swift` (141 lines), `WebsiteIconResolver.swift` (304 lines, mostly request limits and privacy guards) and `WebsiteIconRenderer.swift` (198 lines). **154 lines are unit tests** (13 tests in the existing `unit-tests` target) and 44 lines are `WebsiteIconsSpecs.md`, following the Specs/Tests/implementation triad from `AGENTS.md`. The hooks into existing code are about 23 lines: the preference, the Appearance toggle and one `TileView` call.
- All 1,280 unit tests pass locally with Xcode 27. I haven't run CI.

## Performance, measured honestly

When an accessibility client reaches a Chromium page, Chromium enables its *basic* web accessibility mode for that process (see `render_widget_host_view_cocoa.mm`). That's lighter than VoiceOver's complete mode, which this doesn't trigger. To measure the cost, I used a throwaway Chrome profile with 4 windows of a page mutating 50 DOM nodes every 100 ms (deliberately heavy), and summoned AltTab 20 times a minute with the setting off and on, twice each:

| | Setting off | Setting on |
|---|---|---|
| Chrome CPU, 60 s of summoning | 24.6 s / 24.4 s | 24.8 s / 25.9 s |
| Chrome CPU, 60 s idle afterwards | 26.4 s / 26.1 s | 26.6 s / 28.5 s |
| Chrome memory | ~1,845 MB | ~1,845 MB |

That's roughly **3 to 5% more Chrome CPU on a worst-case page, with no memory change**. AltTab itself used 0.25 s more CPU per 20 summons. Google notes that 5 to 10% of Chrome users already have accessibility enabled by password managers and similar tools, and this setting is opt-in.

## Known limits

- It shows the site's homepage icon, which can differ from a page-specific favicon.
- Private windows can't be detected, which is why the setting discloses it.
- A browser's first-run dialog (Firefox's welcome screen) hides the tab until it's dismissed.

## Why I think it fits AltTab

It addresses the long-standing requests in #527 and #4651 without the N-integrations problem, it's self-contained, and it's off unless someone wants it. I'm an AltTab Pro customer and run many browser windows as separate web apps, so this is the thing I miss every day.

Update: the tidied implementation, including live icon updates while the switcher is open, is now a fresh draft PR, #6039, with this explanation, the measurements and a screenshot. If it's still a no, completely understood. Thanks either way!

PR #6032, opened 2026-09-13
fix: keep selection and hover stable while windows change

josdyr:
Three bugs in AltTab 11.6.1 that show up while the switcher is open and window titles or the window list change underneath it. This happens constantly with browser tabs loading, terminals with spinners, or closing windows from the switcher.

**1. Arrow selection resets on title changes.** Selecting a row with up/down and then letting any window title change snaps the selection back to the default window. Vertical navigation never set `userPickedSelection`, unlike Tab, Shift-Tab and left/right, so `SelectionResolver` re-derived the default on the next refresh. One line in `TilesView`.

**2. Closing the selected window selects the wrong neighbor.** Pressing the close, minimize, hide or quit shortcut on a row often lands on the wrong window afterwards. The app focuses another of its windows when one closes, which reorders the list before the removal arrives, so the stored index skips the original successor. The fix captures the visible neighbors by identity just before the action (`SelectionResolver.removalFallback`). If the target leaves, the first surviving original successor is selected, then the preceding neighbors. Explicit navigation clears the fallback; title refreshes keep it. No timers.

**3. Hover highlight flickers on title updates.** A hover dismissed by keyboard navigation came back on the next title update, and a hover over a stationary pointer disappeared. `TileOverView` now keeps a dismissed hover dismissed until the pointer moves, and refresh-driven reselection (`.selectAt`) no longer resets the hover.

### Testing
- 7 new `SelectionResolverTests` (removal fallback, reorder before removal, hidden neighbors, end of list, identity preserved). All 1,273 unit tests pass locally (Xcode 27, macOS 27.0).
- Live-tested on macOS 27 by closing, hiding, minimizing and quitting windows from the switcher in Chrome (single and multi-tab), Safari, Finder and Calculator, first, middle and last rows.
- Not checked: CI, macOS 26 and earlier.

Related: #6035 changes how held-key cycling wraps, in the same navigation code paths.

---


lwouis:
I've cherry-picked some and modified here. I've credited you as co-author of course. I'll close this PR now
Thank you very much!

PR #6033, opened 2026-09-13
fix: follow native preview corners on macOS 27

josdyr:
On macOS 27 the selected-window preview outline is drawn as a rectangular outer path with a fixed inner radius, which visibly disagrees with the preview panel's rounded corners (the Tahoe variant is #4985).

This follows the native container-concentric corner radius with a continuous layer border, and updates it when the effective corners or the appearance change. macOS 26 and earlier keep the current rendering. It's independent of #6029, which fixes the switcher's glass panel corners.

### Testing
- Checked visually on macOS 27.0 (26A428) with ordinary windows and Device Hub. Builds with Xcode 27.
- Manual regression scenarios added to `AppearanceSpecs.md`.
- Not checked: fullscreen and custom-shaped windows, multiple displays, macOS 26, CI.

---
Part of a set of independent changes from daily use on macOS 27. Each one stands alone and can be merged in any order: #6029 glass panel corners · #6033 preview corners · #6034 app identity · #6036 Device Hub focus · #6032 selection and hover · #6040 snappy close and quit · #6035 keep cycling · #6037 Titles style · #6039 website icons. **Try them all together:** [build guide and demo preset](https://gist.github.com/josdyr/4160589b52b513d72ebcb540c613d1bc)

PR #6035, opened 2026-09-13
feat: add an option to keep cycling while holding the shortcut

josdyr:
Adds **Keep cycling while holding the shortcut** under Controls > Additional controls, off by default.

Today, holding the shortcut stops at the first and last window. That's a sensible default because a held key can't overshoot. But some people (me included) like to hold the key and let the selection loop continuously through their windows, the way a list scroller wraps. With the option on, held repeats wrap in both directions and across rows. Individual presses behave as before, and trackpad swipes keep their edge stops either way.

Implementation: the existing repeat guard in `Windows.cycleSelectedWindowIndex` and the row guard in `TilesView.nextRow` skip the edge stop when the preference is on. No new timer: the repeat rate still follows macOS `InitialKeyRepeat` and `KeyRepeat`, with the existing visibility gating and late-tick rejection. `KeyRepeatTimerSpecs.md` documents both behaviors and the manual checks.

### Testing
- Built with Xcode 27. All unit tests pass locally.
- Unconditional wrapping (the behavior with the option on) has been in daily use in my build on macOS 27: several laps forward and backward, and releasing stops immediately. The preference toggle itself, and the option-off path in this branch, still need a manual check.
- Not checked: CI.

Related: #6032 fixes selection bugs in the same navigation code paths.

---


lwouis:
Hi,
Thank you for sharing this work. Since #351, I'm very careful not to add any setting that would see niche usage. I'm afraid this setting would. For this reason, I can't merge the MR.
I appreciate your sharing this work.
Thank you

PR #6037, opened 2026-09-13
feat: improve the Titles style with an app name column and fitted width

josdyr:
Improvements to the Titles style that make it read like a native macOS list. They're opt-in where they change content, and on by default where they only change presentation.

Titles style with the app name column, fitted width and solid selection

*From my build on macOS 27 with the app name column and right alignment enabled. The website icons in this screenshot come from #6039, not this PR.*

### App name column (Customize style, Titles, off by default)
- **Show app names in a separate column** puts each window's app name in its own column before the icon, so icons and window titles line up across rows. The title then shows only the window title.
- **Right-align app names** aligns names toward their icons. Both mirror in right-to-left layouts.
- The column fits the widest visible app name, capped at 240pt and a quarter of the row. It never shrinks while searching, so filtering doesn't shift titles.

### Fitted width
- The panel now fits the widest visible row instead of a fixed readability width, so it shrinks when long-titled windows close and grows for long titles.
- **Minimum width** (240 to 600pt, default 300) and **Maximum width** (50 to 95% of the visible screen, default 90%) sliders in Customize style.
- Spare room of max(24pt, twice the font size) absorbs spinners and counters in titles, so the panel doesn't jitter. Selection changes never resize it.
- Titles are measured with the current font, so switching between displays of different sizes (Auto size) doesn't truncate titles.

### Selection and hover
- The selected row uses the solid system selection color with the matching text color for the title, app name and status icons, like native macOS lists.
- Hover is a light accent tint (stronger with Increase Contrast) without an outline.
- An icon whose silhouette has under 1.5:1 contrast with the selection color (a blue icon on blue) gets a tight neutral halo, stronger with Increase Contrast. It's sampled once per icon at 16x16.

### No hover tooltips in the switcher
- Titles, app names and status icons no longer show mouse-over tooltips. While hovering rows to pick a window they popped up over neighboring rows and repeated what the row already shows. This applies to every style, since those tooltips are shared code.
- The full title and status descriptions remain available to VoiceOver as accessibility labels and help. Settings tooltips are unchanged.

Apart from the tooltips, Thumbnails and App Icons styles are unchanged.

### Testing
- New tests in `AppearanceTests` for fitted and stable width, app name column width, luminance contrast and icon edge sampling. All 1,273 unit tests pass locally (Xcode 27, macOS 27.0).
- Specs and manual checks in `AppearanceSpecs.md`.
- The same layout code has been in daily use in my build on macOS 27 in light mode, across a built-in display and an external monitor. In this branch the app name column is a proper preference (my build used a local defaults key), which I haven't run live yet. Not checked: right-to-left layouts, dark mode, CI.

---


lwouis:
This is an interesting change. For now, I'd like to keep what we have in the app regarding the Titles appearance
Thank you

PR #6039, opened 2026-09-13
feat: show website icons on browser windows

josdyr:
Browser windows all show the same browser icon, so several windows of Safari or Chrome are hard to tell apart. This adds an opt-in **Show website icons** setting (Appearance, off by default): each browser window shows the icon of the site it displays. It picks up where #6030 left off, with the concern raised there answered below: **there is no browser integration**.

AltTab switcher with website icons on Safari, Chrome and Firefox windows

*Safari, Chrome and Firefox windows on macOS 27, each with its site's icon. The selected row shows GitHub's transparent logo on a white backing, which keeps it visible on the accent color. The app name column comes from #6037.*

### No browser integration
- No extension, no AppleScript or Apple Events, no JavaScript, no bundle IDs or per-browser code. One generic path.
- The page URL comes from the standard macOS accessibility attributes: `AXDocument` on the window, or `AXURL` on the tab's `AXWebArea`. VoiceOver relies on these, so WebKit, Chromium and Firefox keep them working.
- Tested unchanged in Safari, Safari Technology Preview, Chrome, Orion and Firefox. Orion and Firefox needed no code at all.
- Browser updates don't need AltTab changes. If a browser ever stopped exposing the URL, its windows would simply show the normal app icon again.

### Changes
- **Find the page:** a bounded background accessibility lookup (max 128 elements, depth 8, 100 ms per call). It stops at the first web area, never reads page content, and skips AltTab's own windows.
- **Fetch the icon:** only the site's public homepage `https://host/` is requested. The page path, query, fragment, cookies and credentials are never sent. `<link rel="icon">` / `apple-touch-icon`, then `/favicon.ico`. Local and private network addresses are refused, redirects included. Work is shared and capped: 8 transfers, 1 MiB responses, HTML stops at `</head>`, results cached 30 s.
- **Draw it:** a 64 px rounded tile like app icons. Transparent logos get a white backing; logos that would vanish on white get a dark one. Web pages without an icon show a globe; other windows keep their app icon.
- **Live updates:** icons fill in while the switcher is open, including windows opened and pages loaded meanwhile. A result stands until the window title changes (navigation). A missing web area is retried at 0.25, 0.5, 1 and 2 s. Windows without a web area ignore title changes, so a terminal spinner costs nothing.
- **Privacy:** the setting's description says requests happen outside the browser, including for private windows (they can't be detected). Turning it off cancels transfers and restores app icons immediately.

[...]

PR #6040, opened 2026-09-13
perf: remove closed and quit windows from the switcher immediately

josdyr:
Closing a window or quitting an app from the open switcher (the W and Q shortcuts) feels sluggish, most noticeably with Chrome. The row stays for about a third of a second after closing, and for as long as the app takes to quit.

I instrumented the close path on macOS 27 with Chrome. Keypress to AX destroy notification: 13 ms. AX destroy to removal: about 80 to 160 ms, of which 150 ms is the fixed settle delay in `reconcileAxElementEnd`. Removal to repaint: up to 218 ms, because `refreshOpenUiAfterExternalEvent` goes through the 200 ms UI throttle, which stays busy while any window title has a spinner. Total: about 320 ms.

### Changes
- **Removals repaint at once:** `refreshOpenUiAfterExternalEvent(windowRemoved: true)` bypasses the UI throttle. Removals are rare and already batched by `Windows.removeWindows`; the throttle keeps coalescing everything else.
- **Expected closes skip the settle delay:** a window the user closed from the switcher within the last 2 s queries its AX-end immediately instead of after 150 ms. The AX and WindowServer queries still decide, so the "reflect OS state, never a predicted one" rule holds, and a window that survives its close (a Chromium node rebuild, a cancelled close) stays listed.
- **Quit hides rows immediately:** quitting an app from the switcher hides its rows right away instead of waiting for termination, which can take seconds for a browser. If the app is still running 5 s later (a save dialog, a cancelled quit), a scheduled refresh brings its rows back. A second quit still force-quits.

Documented in `WindowLifecycleSpecs.md`.

### Testing
- Instrumented runs on macOS 27.0: Chrome close went from about 320 ms to about 50 ms keypress-to-repaint. Quitting Calculator hid its row 4 ms after the key, before termination completed (which took another 25 ms).
- All unit tests pass locally.
- Not checked: the 5-second restore after a cancelled quit (e.g. an unsaved-changes dialog), CI.

Related: #6032 keeps the selection on the right neighbor after these removals.

---


lwouis:
Hi @josdyr,

Thank you very much for sharing all these pull requests!

I'm a bit overwhelmed currently with the macOS Golden Gate release. I'm trying really hard to ship a fix for the visual glitch it introduced, as well as other things I was working on before that happened.

As soon as I'm done with that batch of fixes and this release, I'll look into the multiple pull requests you have shared and be able to be more relaxed and review them. Thank you for your patience 🙇 

Thank you

lwouis:
Hi @josdyr,

Thank you for measuring this so precisely. The breakdown made it easy to see where the time goes.

**The 200ms UI throttle.** Since you opened this PR, I've reworked how the open switcher schedules its repaints. It now repaints about 16ms after a window is removed, instead of waiting up to 200ms behind the throttle. Title updates (spinners included) are throttled on their own and no longer hold repaints back. This will ship in the next release, and closing windows from the switcher should feel much snappier there. The first change of this PR won't be needed anymore.

**Skipping the 150ms settle delay.** I'd prefer not to take this one as is. Here is why the delay exists.

When an app says the accessibility handle of a window died, it can mean two things: the window was closed, or the app threw the handle away and is building a new one for the same, still-open window. Chromium and Electron do the latter routinely. That notice tends to arrive a little before the app has finished updating itself, so we wait a moment, then ask both the app and the WindowServer whether the window still exists. We ask only once (with a 2s timeout as a fallback), so whatever answer we get at that moment is final.

Asking immediately can go wrong in both directions:

1. **A live window disappears.** The 2s "the user just closed this" flag is set whenever close is pressed, including when the close doesn't actually happen: Chrome's "Leave site?" or "Close all tabs?" prompts, or an unsaved-changes prompt in a native app. If Chromium rebuilds the handle during those 2s, we'd ask in the middle of the rebuild, the app would say the window is gone while it's still on screen, and we'd mark it closed. Once marked closed, the background sweep doesn't re-add it, so it stays missing until the user focuses it.
2. **A closed window lingers.** Right at teardown, the app may not answer (it's busy), may still list the window, or, for native tabs like Finder or Terminal, may not have updated its tab group yet. All of these read as "unsure" or "still alive", and since there's no retry, the row stays until the WindowServer's own destroy event arrives. We've measured that event lagging by seconds to minutes, and sometimes it never comes. This is the ghost-row bug fixed in 7fbba111 and 134fbb0a.

So "a window that survives its close stays listed" holds only if the app answers accurately right away, which is what the delay protects against. A safer version could be a shorter delay for closes started from the switcher, plus one retry when the answer is "unsure" or "still alive", tested on Chrome, Electron, native tabs and a cancelled close. Before going there, I'd like to see how closing feels on the next release with the faster repaint, since the 150ms may matter much less once the 200ms is gone.

Thank you again for digging into this!

